PR #5, opened 2016-04-03
Client api call

schroerbrian:
ugh, the entire app.


jfhamlin:
Should we start with just the starter kit? Then it'll be easier to identify what we're adding and removing here.

schroerbrian:
yes that is very wise idea

PR #7, opened 2016-04-05
Import the latest react-starter-kit with a workaround for a test failure.

jfhamlin:
Uses a recent HEAD of master of https://github.com/kriasoft/react-starter-kit.

See https://github.com/kriasoft/react-starter-kit/issues/522 for the test issue.


jfhamlin:
@achadha says this looks good! I'll merge.

PR #32, opened 2016-05-17
Restructure

achadha:


jfhamlin:
A few high-level questions and comments:

1. Let's squash these into a couple of commits before merging!
2. Is there a test harness?
3. What do we need to do to get the CI build working?

.gitignore
```diff
@@ -0,0 +1,9 @@
+# Include your project-specific ignores in this file
+# Read about how to use .gitignore: https://help.github.com/articles/ignoring-files
+
+build
+node_modules
+ncp-debug.log
+npm-debug.log
+restructure/dist
+.DS_Store
```
jfhamlin: This isn't really a project-specific file. We should all have this in a global .gitignore: https://help.github.com/articles/ignoring-files/#create-a-global-gitignore

achadha:
2. At the moment, no. Need to do that.
3. I think we need to add 'sudo: required' to the travis.yml. I feel like it shouldn't be necessary, but something is up with the permissions, and that was my way around it.

components/Resource/Resource.js
```diff
@@ -0,0 +1,65 @@
...
+class Resource extends React.Component {
+	constructor() {
+		super();
+		this.state = {resource: {}};
+	}
+
+	loadResourceFromServer() {
+		let { query } = this.props.location;
+		let resourceID = query.id;
+		let url = 'http://localhost:3000/resources/' + resourceID;
```
jfhamlin: Can we make the API URL configurable, and initialize it in one place?

PR #33, opened 2016-05-25
Updated front page / categories selection

lelandg559:


components/LargeHeader.js
```diff
@@ -0,0 +1,43 @@
...
+                    <img src={'assets/images/logo-small-white@3x.png'} />
+                </a>
+            </div>
+            <nav id="navbar" className="collapse navbar-collapse">
+            </nav>
+        </div>
+        <div id="largeheader_content" className="container">
+            <div className="row">
+                <div className="largeheader_centerText">
+                    Find the best community<br />
```
achadha: Try not to keep text directly in a div tag

components/LargeHeader.js
```diff
@@ -0,0 +1,43 @@
+import React from 'react';
+
+function LargeHeader() {
+  return (
+    <nav className="navbar navbar-default largeheader_background" style={{backgroundImage: 'url(' + 'assets/images/bg.png' +')'}}>
```
lelandg559: need to change this so that its loaded properly in CSS with urlloader or fileloader (havent worked that out)

achadha: Agree. Will have to use a loader in webpack.

http://survivejs.com/webpack/loading-assets/loading-images/

components/CategoryPage.js
```diff
@@ -2,6 +2,8 @@
 import React, { Component, PropTypes } from 'react';
 import { Link } from 'react-router';
 import Gmap from './resources/Map.js';
+import LargeHeader from './LargeHeader.js'
+import Footer from './Footer.js'
 
 var categories = [];
```
achadha: I'd prefer for all of the components to be created using the same standards. CategoryBox, CategoryList, and Category should probably be using ES6 classes like ContentPage is. It might be a bit nitpicky, and probably doesn't need to block a merge, but I feel like we should do it at some point.

components/CategoryPage.js
```diff
@@ -65,13 +67,26 @@ var CategoryList = React.createClass({
 var Category = React.createClass({
   render: function() {
     return  (
-      <li className="category_list_item">
-        <div className="category-name">
+    <div>
+      <div className="col-lg-2 col-md-2 col-sm-2 hidden-xs category_option">
           <Link className="category_brand" to={{ pathname: "resources", query: { categoryid: this.props.categoryid } }} >
-            <span>{this.props.name}</span> 
+              <img src={'assets/images/' + this.props.image_path} alt="boohoo" className="img-responsive"/><br />
```
achadha: Lol, alt should probably be helpful. What do? I mean, the category's name will be right there too.

components/CategoryPage.js
```diff
@@ -44,18 +47,17 @@ var CategoryList = React.createClass({
   
     var categoryNodes = this.props.categories.map(function(category) {
       return (
-        <Category name={category.name} key={category.id} categoryid={category.id} />
+        <Category name={category.name} key={category.id} categoryid={category.id} image_path={category.image_path}/>
```
achadha: You could probably also remove the categoryid prop and just use key below.

components/CategoryPage.js
```diff
@@ -65,13 +67,26 @@ var CategoryList = React.createClass({
...
           </Link>
-        </div>
-      </li>
+          {this.props.name}
+      </div>
+      <div className="hidden-lg hidden-md hidden-sm col-xs-12">
+        <Link className="category_brand" to={{ pathname: "resources", query: { categoryid: this.props.categoryid } }} >
+          <button className="btn btn-default category_button" type="submit">
+            <div className="col-xs-4 category_xs_button_image">
+                <img src={'assets/images/' + this.props.image_path} alt="boohoo" className="img-responsive"/><br />
```
achadha: Same alt thing.

components/CategoryPage.js
```diff
@@ -65,13 +67,26 @@ var CategoryList = React.createClass({
 var Category = React.createClass({
   render: function() {
     return  (
-      <li className="category_list_item">
-        <div className="category-name">
+    <div>
```
achadha: Let's look at how we can simplify this container in the future.

components/LargeHeader.js
```diff
@@ -0,0 +1,43 @@
...
+        </div>
+        <div id="largeheader_content" className="container">
+            <div className="row">
+                <div className="largeheader_centerText">
+                    Find the best community<br />
+                    resources for your needs
+                </div>
+            </div>
+            <div className="row">
+                <div className="col-lg-2 col-md-2 col-sm-1 col-xs-1">
```
achadha: The bootstrap classes scale up. So if you had only used the class col-xs-12, it would use 12 columns for all screen views. As such, your classnames can be simplified here into "col-xs-1 col-md-2".

components/LargeHeader.js
```diff
@@ -0,0 +1,43 @@
...
+                </div>
+                <div className="input-group col-lg-8 col-md-8 col-sm-10 col-xs-10">
```
achadha: Can simplify bootstrap columns (see comment above).

components/CategoryPage.js
```diff
@@ -65,13 +67,26 @@ var CategoryList = React.createClass({
...
+      <div className="col-lg-2 col-md-2 col-sm-2 hidden-xs category_option">
```
achadha: The bootstrap classes scale up. So if you had only used the class col-xs-8, it would use 8 columns for all screen sizes. As such, your classnames can be simplified here into "col-xs-1 col-md-2".

PR #36, opened 2016-05-26
Don't require superuser access for development.

jfhamlin:
Updates the webpack config to put the compiled bundle in `./build/public` rather than the developer's `/dist` directory.

We still need to do more to get CI working:

1. ~~Fix `npm run build`.~~
2. ~~Fix `npm run test`.~~ In an ideal world it would also run tests.


lelandg559:
LGTM

PR #39, opened 2016-06-01
Silverhawk/alt user location

lelandg559:


index.html
```diff
@@ -8,7 +8,7 @@
 </head>
 <body>
   <div id="root"></div>
-  <script></script>
   <script src="dist/bundle.js"></script>
```
lelandg559: we cant have /dist anymore right?

jfhamlin: `/dist` in the URL is fine. The server maps it to the local `public` directory. What we used to be doing was actually installing the files to the `/dist` directory.

lelandg559:
LGTM

PR #49, opened 2016-06-05
Remove old react-starter-kit code and centralize application code.

jfhamlin:
- Move app code and assets to the app/ directory from the root directory.
- Leverage webpack for cleaner references to static assets.

Sign-off from both @ZanderSparrow and @lelandg559 would be great.


app/components/LargeHeader.js
```diff
@@ -1,8 +1,11 @@
 import React from 'react';
 
+let bgImage = require('../assets/images/bg.png');
```
jfhamlin: Using `require` tells webpack that the image is a dependency, so it will a) copy it to the output directory for us and b) return the public path to the image, so we don't have to hard-code it as we used to.

lelandg559: got it

lelandg559:
when I ran this with master for the backend, I get a 502 from http://askdarcel.lvh.me:8080/api/categories when i load the front page, any thoughts?

app/components/Category.js
```diff
@@ -0,0 +1,38 @@
+import React, { Component, PropTypes } from 'react';
+import { Link } from 'react-router';
+
+var icons = require.context("../assets/images", true, /ic-.*\.png$/i);
```
jfhamlin: This looks hairy, but it lets us leverage webpack when using these images even though we've got the indirection that keeps us from referring to them using constants in the code (that is, the server returns the image name).

ZanderSparrow:
This looks good to me. Thanks for cleaning up those old files!

PR #50, opened 2016-06-07
Fixes #41, Fixes #42

### LINKED ISSUE #41: Duplicate numbers show up in map on right hand side of search results page
Only one of each number should show up. Also, we should consider limiting the amount of numbers displayed.

### LINKED ISSUE #42: Only ten results show on search results page; others display as "loading" but do not load

ZanderSparrow:
* The API key for google maps is git ignored, but now loads
* Only 9 resource results show at a time, can navigate with Next and Previous buttons

### DIFF
--- a/app/components/Resources/Map.js
+++ b/app/components/Resources/Map.js
@@ -5,7 +5,9 @@ class Gmap extends Component {
     super(props);
 
     this.state = {
-      userLocation: {}
+      userMarker: null,
+      map: null,
+      markers: []
     };
   }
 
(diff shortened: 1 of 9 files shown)


app/utils/config.example.js
```diff
@@ -0,0 +1,4 @@
+
+module.exports = {
+  GOOGLE_API_KEY : 'REPLACE_ME'
```
jfhamlin: If we're within the free usage tier's limits, no key is required: http://stackoverflow.com/questions/2769148/whats-the-api-key-for-in-google-maps-api-v3. Should we leave it out for now?

ZanderSparrow: We could do that. I could just delete the '&key=...' from the script and leave the rest in place so that adding a key will be easy if we want to track usage or eventually exceed the quota.

jfhamlin: We still need a way to configure the key at deploy time, which means serving an additional script with the config or find-replace on `bundle.js`. That's not a high priority for me yet, I think, unless staging won't work without it.

app/components/Resources/ResourcesTable.js
```diff
@@ -13,7 +13,13 @@ const cats = {
 class ResourcesTable extends Component {
 	constructor() {
 		super();
-		this.state = {categoryName: 'Category Name', resources: [], location: null};
+		this.state = {
+			categoryName: 'Category Name',
```
jfhamlin: Should this be in `props`, since (I think) it's immutable?

jfhamlin: Ah, we fetch it in `loadResourcesFromServer()` below. I briefly talked to @achadha about react-redux a few weeks ago. We should look into whether that's the direction we want to go for the project; I think it leverages React's virtues well.

lelandg559:
LGTM

app/components/Resources/Map.js
```diff
@@ -17,17 +19,30 @@ class Gmap extends Component {
...
+      markers[i].setMap(map);
+    }
+  }
+
+  generateMarkers(locs, map) {
+    let markers = [];
+    let offset = 1;
+
+    locs.forEach((loc, index) => {
+      var latLang = new google.maps.LatLng(loc[0], loc[1]);
```
lelandg559: Talked about code style when it comes to using potentially unclear object indices to get/store variables.  I infer that loc[0] is probably latitude and loc[1] is probably longitude but it can become difficult to sort that out when those situations get more complex.

I'm a fan of being very explicit about what variables are and defining them, but I'm not familiar with standards of javascript as far as that goes.

PR #51, opened 2016-06-09
Updated headers

lelandg559:


lelandg559:
I plan to merge common code from header and largeheader together but I am not ready to do that yet.

jfhamlin:
> I plan to merge common code from header and largeheader together but I am not ready to do that yet.

👍

app/components/Header.js
```diff
@@ -1,16 +1,20 @@
 
 import React from 'react';
 import { Link } from 'react-router';
+let bgImage = require('../assets/images/bg.png');
+let smallLogo = require('../assets/images/logo-small-white@3x.png');
```
jfhamlin: Might be cool to throw a js module somewhere, maybe under `/assets`, that provides an index of all images so we don't have this boilerplate.

lelandg559: yeap, ill review that when I make a second pass for internal cleanup.

lelandg559:
Sondra gave LGTM

PR #52, opened 2016-06-11
Add an endpoint, /_version, that describes the deployed application version.

jfhamlin:
Also removes some lingering react-starter-kit files.


tools/docker-build.sh
```diff
@@ -0,0 +1,31 @@
+COMMIT=${TRAVIS_COMMIT::8}
+SANITIZED_BRANCH=$(echo $TRAVIS_BRANCH|sed 's|/|-|g')
+REPO=sheltertechsf/askdarcel-web
+
+docker login -e $DOCKER_EMAIL -u $DOCKER_USERNAME -p $DOCKER_PASSWORD
+
+if [[ -n "$TRAVIS_TAG" ]]; then
+    TAG="$TRAVIS_TAG"
+else
+    if [ "$TRAVIS_PULL_REQUEST" != "false" ]; then
```
jfhamlin: Looks like I've got tabs in here, but just want spaces. Will fix.

lelandg559:
http://askdarcel.lvh.me:8080/_version redirects back to the home page for me, is that the correct path?

jfhamlin:
It only affects CI and the docker image, so it won't affect the local server. You should know what version you're running in local dev :-P.

lelandg559:
LGTM

PR #54, opened 2016-06-12
Updated UI to handle new backend API JSON format

lelandg559:


lelandg559:
Merging this to match the change to API response format

PR #57, opened 2016-06-14
Derekfidler/css styling

derekfidler:
CSS Refactoring of the whole front end. Removed bootstrap, added custom mixin library, reset, and variables, and other custom framework modules. Style changes to home page and result pages.


app/components/Category.js
```diff
@@ -10,26 +10,16 @@ icons.keys().forEach(function (key) {
...
-              <img src={iconPathMap[this.props.image_path]} alt={this.props.name} className="img-responsive"/><br />
-            </div>
-            <div className="col-xs-8 category_xs_text">
-              {this.props.name}
-            </div>
-          </button>
-        </Link>
-      </div>
-    </div>
+    <li className="category-item">
```
jfhamlin: :+1:

.idea/watcherTasks.xml
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
jfhamlin: Looks like an IDE config file. Usually want to add these to your global git ignore settings so they don't end up in commits. I can get a reference on this to you if you haven't done it before.

derekfidler: Yes, it is. Send me the ref, I didn't want to mess with gitignore until i knew what i was doing. I thought I left all the ide files untracked but this one must have been added accidentally.

jfhamlin: https://help.github.com/articles/ignoring-files/#create-a-global-gitignore

package.json
```diff
@@ -24,6 +24,8 @@
     "webpack": "^1.13.0"
   },
   "devDependencies": {
+    "babel-core": "^6.9.1",
+    "babel-loader": "^6.2.4",
```
jfhamlin: We already have `babel-loader` above in `dependencies`. Do we need it here?

jfhamlin: TBH though I'm not yet familiar with the finer distinctions between `dependencies` and `devDependencies`. Since we're just serving static files in production and not running node, maybe everything should end up in `devDependencies`? Not suggesting this get tackled here. Just thinking 'aloud.'

jfhamlin:
I can't say much about the CSS. Once `.idea/watcherTasks.xml` is gone, I'd be happy to merge. We can see if anyone else has comments.

.gitignore
```diff
@@ -6,3 +6,4 @@ node_modules
 ncp-debug.log
 npm-debug.log
 app/utils/config.js
+.idea
```
jfhamlin: Ah, this doesn't really belong here, but in your _global_ gitignore file. Project .gitignore files should refer to project-specific files (e.g. output files that we'll never want to check in).

PR #58, opened 2016-06-14
Fixes #26 - Implement 'Open Now' filter

### LINKED ISSUE #26: Add open now functionality and remove walking distance and just for me filters
Filter buttons:

- [ ] Open Now
- [ ] Walking Distance
- [ ] Just For Me

ZanderSparrow:


app/components/Resources/ResourcesTable.js
```diff
@@ -315,6 +334,31 @@ function buildAddressCell(addresses) {
 	return <span>{addressString}</span>
 }
 
+// Returns the open hours today or null if closed
+function openHours(schedule_days) {
```
jfhamlin: I think camel-case is conventional in JS.

ZanderSparrow: You're right. I was just using what was there when I started working on the project--I think it was done with an underscore to mimic the style of the property on the object that comes back from the database. I can change it, though. I think I might need to make a bunch of changes anyway once the other pull requests are merged, to fix conflicts.

lelandg559:
LGTM

Also dang, UI looks way nicer now. @derekfidler

PR #60, opened 2016-06-17
Add street view images

ZanderSparrow:
Plus some additional cleanup:
* Fix numbering on resources
* Header branding links to front page


app/components/Resources/ResourcesTable.js
```diff
@@ -360,6 +360,16 @@ function buildAddressCell(addresses) {
 	return <span>{addressString}</span>
 }
 
+function buildImgURL(addresses) {
+	if(addresses.length) {
+		return "https://maps.googleapis.com/maps/api/streetview?size=400x400&location=" +
+		  addresses[0].latitude + "," + addresses[0].longitude +
+		  "&fov=90&heading=235&pitch=10";
+	} else {
+		return "http://lorempixel.com/200/200/city/";
```
jfhamlin: I wonder if @derekfidler has an icon we could use here.

app/components/Resources/ResourcesTable.js
```diff
@@ -159,7 +159,7 @@ class ResourcesTable extends Component {
 						<div className="results-table">
 							<header>
                 <h1 className="results-title">{this.state.categoryName}</h1>
-                <span className="results-count">{this.state.allResources.length} Results</span>
+                <span className="results-count">{this.state.allResources.length - 1} Results</span>
```
jfhamlin: Interesting. Why the `- 1`?

app/components/Resources/ResourcesTable.js
```diff
@@ -168,10 +168,10 @@ class ResourcesTable extends Component {
 								</ul>
 							</div>
               <div className="results-table-body">
-                <ResourcesList resources={this.state.currentPage} location={this.state.location} />
+                <ResourcesList resources={this.state.currentPage} location={this.state.location} page={this.state.page} />
                 <div className="pagination">
                   <div className="pagination-count">
-                    <p>1 — {this.state.currentPage.length} of {this.state.allResources.length} Results</p>
+                    <p>{this.state.page * 9 + 1} — {(this.state.page + 1) * 9 < this.state.resources.length ? (this.state.page + 1) * 9 : this.state.resources.length - 1} of {this.state.resources.length - 1} Results</p>
```
jfhamlin: Should we pull `9` out into a constant somewhere?

lelandg559:
Works for me, but I'd prefer the "9" thing be configurable like @jfhamlin noticed.

app/components/Resources/ResourcesTable.js
```diff
@@ -56,7 +58,7 @@ class ResourcesTable extends Component {
 		let page = this.state.page + 1;
 		this.setState({
 			page: page,
-			currentPage: this.state.resources.slice(page * 9 + 1, page * 9 + 10)
+			currentPage: this.state.resources.slice(page * resultsPerPage, page * resultsPerPage + (resultsPerPage + 1))
```
jfhamlin: `slice`'s `end` parameter is exclusive, right? For page zero, the call will be `this.state.resources.slice(0, 10)`, which returns resources indexed 0-9. That's ten resources. Do we actually want nine?

app/components/Resources/ResourcesTable.js
```diff
@@ -65,7 +67,7 @@ class ResourcesTable extends Component {
 		let offset = page ? 1 : 0;
```
jfhamlin: What's the offset for?

app/components/Resources/ResourcesTable.js
```diff
@@ -168,13 +170,13 @@ class ResourcesTable extends Component {
 								</ul>
 							</div>
               <div className="results-table-body">
-                <ResourcesList resources={this.state.currentPage} location={this.state.location} />
+                <ResourcesList resources={this.state.currentPage} location={this.state.location} page={this.state.page} />
                 <div className="pagination">
                   <div className="pagination-count">
-                    <p>1 — {this.state.currentPage.length} of {this.state.allResources.length} Results</p>
+                    <p>{this.state.page * resultsPerPage + 1} — {(this.state.page + 1) * resultsPerPage < this.state.resources.length ? (this.state.page + 1) * resultsPerPage : this.state.resources.length} of {this.state.resources.length} Results</p>
```
jfhamlin: For the second number in the range, I think we can simplify to:

```javascript
Math.min(this.state.resources.length, (this.state.page + 1) * resultsPerPage)
```

app/components/Resources/ResourcesTable.js
```diff
@@ -168,13 +170,13 @@ class ResourcesTable extends Component {
...
                   </div>
                   {this.state.page ? <button className="btn btn-link" onClick={this.getPreviousResources.bind(this)}> Previous </button> : null}
-                  {this.state.page <= Math.floor(this.state.resources.length / 9) - 1 ? <button className="btn btn-link" onClick={this.getNextResources.bind(this)}> Next </button> : null}
+                  {this.state.page <= Math.floor(this.state.resources.length / resultsPerPage) - 1 ? <button className="btn btn-link" onClick={this.getNextResources.bind(this)}> Next </button> : null}
```
jfhamlin: What if `this.state.resources.length` is a multiple of `resultsPerPage`?

app/components/Resources/ResourcesTable.js
```diff
@@ -98,16 +100,16 @@ class ResourcesTable extends Component {
 		let page = 0;
 		if(this.state.openFilter) {
 			this.setState({
-				page: page,
+				page,
 				resources: this.state.allResources,
-				currentPage: this.state.allResources.slice(page, page + 9),
+				currentPage: this.state.allResources.slice(page, resultsPerPage),
```
jfhamlin: Should this logic be the same as what we use for setting `currentPage` above?

app/components/Resources/ResourcesTable.js
ZanderSparrow: yes! good catch.

app/components/Resources/ResourcesTable.js
ZanderSparrow: I think it was part of an attempt to quickly fix something that I since fixed for real. removed.

app/components/Resources/ResourcesTable.js
ZanderSparrow: that works

app/components/Resources/ResourcesTable.js
ZanderSparrow: I tried to make the logic hear more readable in the newest version

PR #61, opened 2016-06-24
Add user location to resources query

ZanderSparrow:


app/components/Resources/ResourcesTable.js
```diff
@@ -98,16 +101,16 @@ class ResourcesTable extends Component {
 		let page = 0;
 		if(this.state.openFilter) {
 			this.setState({
-				page: page,
+				page,
 				resources: this.state.allResources,
-				currentPage: this.state.allResources.slice(page, page + 9),
+				currentPage: this.state.allResources.slice(page, resultsPerPage),
```
jfhamlin: If `page` were anything but zero, this wouldn't be correct.

app/components/Resources/ResourcesTable.js
```diff
@@ -116,26 +119,20 @@ class ResourcesTable extends Component {
...
+			this.getLocationGoogle()
+			.then(loc => {
+				self.loadResourcesFromServer(loc);
+			});
+			
 		}
-
-		this.loadResourcesFromServer();
-
+		this.loadResourcesFromServer({lat: 1, lng: 1});
```
jfhamlin: Do we want to make this call? Both branches of the `if` above call `loadResourcesFromServer`.

ZanderSparrow: yeah, good call, that was left over from testing

app/components/Resources/ResourcesTable.js
ZanderSparrow: page is explicitly reset to 0 at the beginning of the function. I changed it, but it seems like a lateral change to me.

jfhamlin: Yeah, it's a minor thing. My only qualm is that the expression looks like it happily takes a variable, `page`, but if that variable actually varied, the expression wouldn't work. We can leave it.

jfhamlin:
LGTM

PR #67, opened 2016-06-30
Fixed bug preventing street view images displaying

lelandg559:


lelandg559:
Comfortable with these changes, I'll circle with UI people on it, but merging now so staging isnt left with this bug.

PR #72, opened 2016-07-17
Add descriptions to resource page

ZanderSparrow:
Replace the placeholder with actual info

### DIFF
--- a/app/components/Resource/Resource.js
+++ b/app/components/Resource/Resource.js
@@ -51,7 +51,7 @@ class Resource extends Component {
                 </div>
               </div>
               <div className="org-desc">
-                <p>This is an organisation's long description{this.props.description}</p>
+                <p>{this.state.resource.long_description || this.state.resource.short_description || "Description of the resource"}</p>
               </div>
               <ul className="org-details">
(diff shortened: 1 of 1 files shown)


app/components/Resource/Resource.js
```diff
@@ -51,7 +51,7 @@ class Resource extends Component {
                 </div>
               </div>
               <div className="org-desc">
-                <p>This is an organisation's long description{this.props.description}</p>
+                <p>{this.state.resource.long_description || this.state.resource.short_description || "Description of the resource"}</p>
```
jfhamlin: Do we really want "Description of the resource" as a fallback?

ZanderSparrow: "No description available"? or just blank?

jfhamlin: I'm curious what @derekfidler thinks.

achadha: Pinged Derek on Slack, but if he doesn't respond by tomorrow, let's just change it to "No description available." Other than that, this looks good to merge.

derekfidler: no description available is preferable.

PR #73, opened 2016-07-17
Fix loading bug when there are no results

ZanderSparrow:


app/components/Resources/ResourcesMap.js
```diff
@@ -71,11 +71,12 @@ class Gmap extends Component {
       userMarker: userMarker
     })
 
-    let latLang = new google.maps.LatLng(this.props.markers.results[0][0], this.props.markers.results[0][1]);
-    map.setCenter(latLang);
-
-    this.generateMarkers(this.props.markers.results, map);
+    if(this.props.markers.results && this.props.markers.results.length) {
+      // let latLang = new google.maps.LatLng(this.props.markers.results[0][0], this.props.markers.results[0][1]);
```
jfhamlin: If we don't want this code to be executed, should we just remove it?

ZanderSparrow: wasn't sure what we'd want. okay to always center on the user rather than a resource?

achadha: Going to just remove the commented code for now. We can discuss tomorrow whether we'd want to bring it back.

PR #77, opened 2016-07-20
Initial link of clicking search button to redirecting to search results

lelandg559:


app/components/LargeHeader.js
```diff
@@ -19,7 +26,7 @@ function LargeHeader() {
         <div className="search-container form-row" role="search">
           <input type="text" className="search-field" placeholder="I'm looking for..." name="srch-term" id="srch-term" />
           <input type="text" className="location-field" placeholder="Near me" name="srch-location" id="srch-location" />
-          <button id="largeheader_searchbutton" className="button search" type="submit"><i className="material-icons">search</i> <span>Search</span></button>
+          <button onClick={submitSearch} id="largeheader_searchbutton" className="button search" type="submit"><i className="material-icons">search</i> <span>Search</span></button>
```
lelandg559: I suspect we need to more than onClick here, is there a best practice on handling any sort of input field submit? (enter key or click on button?)

app/components/LargeHeader.js
```diff
@@ -1,8 +1,15 @@
 import React from 'react';
 import { images } from '../assets';
+import { browserHistory } from 'react-router'
 
 
 function LargeHeader() {
+  var submitSearch = function() {
+    browserHistory.push({
+        pathname: '/resources',
+        query: { query: document.getElementById("srch-term").value }
```
lelandg559: unsure if theres a more 'react' way to get the value from the input text field

app/components/LargeHeader.js
jfhamlin: We're in a form, yeah? Is it `onSubmit` On the form tag? Then the button should have a "submit" type attribute. Not 100%, that's just what a quick Google search turned up.

app/components/LargeHeader.js
jfhamlin: I think you can restrict the selector to the component rather than the full document. Also, should we validate the text? We may not want an empty search bar or a whitespace-only entry to submit

ZanderSparrow: you can use refs to get the value of inputs and things:
https://facebook.github.io/react/docs/more-about-refs.html

app/components/LargeHeader.js
ZanderSparrow: I don't have time to test this, but to get functions working I usually need something like onClick={this.myFunction.bind(this)}. Also if you use submit you might need to capture the event and call preventDefault for example submitSearch(e){ e.preventDefault(); and then more functionality...}

jfhamlin:
LGTM

PR #95, opened 2016-08-25
Search fixes

lelandg559:
window.location.reload(); is janky, looking for better ways to refactor that behavior

### DIFF
--- a/app/components/Header.js
+++ b/app/components/Header.js
@@ -1,24 +1,37 @@
-
 import React from 'react';
 import { Link } from 'react-router';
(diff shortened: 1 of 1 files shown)


app/components/Header.js
```diff
@@ -1,24 +1,37 @@
-
 import React from 'react';
 import { Link } from 'react-router';
+import { browserHistory } from 'react-router'
```
achadha: These can be in the same import statement. We can address later.

achadha:
LGTM

PR #97, opened 2016-08-31
Fix href for website.

achadha:
If someone could verify, that would be great. Having trouble hitting the backend.


lelandg559:
LGTM and verified

PR #100, opened 2016-09-07
Remove image_path from categories call

lelandg559:
Depends on remove_image_path PR in api


lelandg559:
back end change was merged, need to do this to keep app stable.

PR #103, opened 2016-10-06
Temporary google API fix tweak

lelandg559:
Not ideal but we need this for now to fix behavior.


lelandg559:
need to test this in staging, I've verified behavior in dev, this can be reverted if need be.

PR #104, opened 2016-10-06
Bugfix with temporary API key fix

lelandg559:


lelandg559:
need to test this in staging, I've verified behavior in dev, this can be reverted if need be.

PR #116, opened 2016-12-18
walking direction link added

twolfe2:


lelandg559:
Works for me, I'll get Avneet to take a look.

achadha:
Let's get the Redux stuff in first, then this. Looks like this is dependent on that branch.

twolfe2:
Should be good to merge now.

achadha:
Yup looks good.

PR #121, opened 2016-12-29
Thomas/auth

twolfe2:


app/store/configureStore.dev.js
```diff
@@ -3,8 +3,19 @@ import reduxImmutableStateInvariant from 'redux-immutable-state-invariant';
 import thunkMiddleware from 'redux-thunk';
 import rootReducer from '../reducers';
 
+
 // Download redux dev tools here: https://github.com/zalmoxisus/redux-devtools-extension
 
+// const store = createStore(
```
achadha: I want to merge this into master, so if we could either uncomment this, or remove it (and then make an additional issue to track it), that would be great.

app/store/configureStore.prod.js
```diff
@@ -2,6 +2,11 @@ import { createStore, applyMiddleware } from 'redux';
 import thunkMiddleware from 'redux-thunk';
 import rootReducer from '../reducers';
 
+// const store = createStore(
```
achadha: Same as above.

achadha:
Also, the styles for the login page need to be looked at.

PR #123, opened 2017-01-12
fixed auth so access token is updated on each request

twolfe2:


app/reducers/authReducer.js
```diff
@@ -1,8 +1,8 @@
 import * as types from '../actions/actionTypes';
 
-const authHeaders = JSON.parse(localStorage.getItem('authHeaders'));
+const authHeaders = localStorage.getItem('access-token');
 
-const isAuthenticated = authHeaders && authHeaders['access-token'] && authHeaders.client ? true : false;
+const isAuthenticated = authHeaders ? true : false;
```
achadha: Will this work if there is an old token in local storage?

Issue #141, opened 2017-02-08
Verify Open Now filter on search results is displaying accurate data


derekfidler:

jjfreund:
It is not -- we need to use the generic function, I think this has custom functionality not expecting minutes.

richardxia:
Looks like this was fixed at some point.

PR #153, opened 2017-02-12
Added open until and closed until to resource list

twolfe2:


twolfe2:
Fixes #141

app/components/Search/ResourcesRow.jsx
```diff
@@ -35,6 +35,55 @@ class ResourcesRow extends Component {
...
+      closingOpeningTimes[item.day.replace(/,/g, '')] = {
+        open: item.opens_at,
+        close: item.closes_at
+      };
+    });
+
+    if(closingOpeningTimes[currDay] && hour < closingOpeningTimes[currDay].close) {
+      return {
+        open:true, 
+        time: timeToString(closingOpeningTimes[currDay].close)
```
ZanderSparrow: this is the closing time, right? maybe change to something more descriptive like 'closingAt'.

app/components/Search/ResourcesRow.jsx
```diff
@@ -35,6 +35,55 @@ class ResourcesRow extends Component {
...
+            open: false,
+            time: `${timeToString(closingOpeningTimes[d].open)} ${d}`
+          };
+          return true;
+        }
+      });
+
+      if(result.time) return result;
+
+      days.some(d => {
```
ZanderSparrow: not sure how this is different from the block above (other than splicing out today). If the resource is closed (so return not yet reached) why not just jump to here and skip the else above?

app/components/Search/ResourcesRow.jsx
twolfe2: Good idea. I'll apply this change

app/components/Search/ResourcesRow.jsx
twolfe2: The first loop checks to see if the resource is open any days that are remaining in the week. 

If no open days are found for the remainder of the week, I start at the beginning of the week and check to see the soonest day the resource is open.

ZanderSparrow: I see--You want the next open day, so you want to start with the days after the current day. You could do it in one loop by splitting the array at the current day and then putting them back together with current day at the end. This is also fine, though.

Issue #127, opened 2017-02-07
Wrong localisation on Chrome


Kphi:
My laptop is positioned near Denver airport. Don't remember if I already granted permission for Chrome to give my localisation to askdarcel.

derekfidler:
Should display an error banner if localisation is not available.

PR #158, opened 2017-02-23
Pull user location checks to App component and add bounds check.

richardxia:
This is my first attempt at trying to do something in React that's more than just editing minor UI things, so please let me know how I should do things correctly in React.

Previously the user location was being checked in multiple places separately:
- the top-level App component
- the Search/ResourcesTable component
- the Resource/ResourceMap component

I moved it all to the the App component and passed it all the way down to both the other components, but I had to mess around with the state vs. props setting a bit in order to get it to work correctly without running into an infinite loop. I'm not too familiar with how asynchronous state updates should get passed down, but what I did was update the state at the top level and pass the location to subcomponents as props and it seems to work now.

The other thing I did was to check if the user's location is within the bounding box that represents San Francisco, rejecting the location if it's not. Since we have two services for getting the user's location (the browser geolocation API and Google's Geolocation service API), I think we have a high chance of getting it right. If they both fail, then I just set it the location to a random coordinate in the middle of SF, around Market and Van Ness. This fixes the issue as described in #127, but it is kind of questionable because I've hardcoded coordinates specific to SF.

### DIFF
--- a/app/components/App.js
+++ b/app/components/App.js
@@ -3,97 +3,97 @@ import Navigation from './Navigation';
 import CategoryPage from './Find/FindPage';
 import ResourcesTable from './Search/ResourcesTable';
 
-/**
- * Get user location.
- *
- * Makes use of both the HTML5 Geolocation API and the Google Maps Geolocation
- * API. Currently restricts the location to within San Francisco to avoid
- * inaccurate geolocation results, but this should be removed if more locations
- * are added.
- *
- * @returns A Promise of a location, which is either an object with `lat` and
- * `lng` properties or an error if location is unavaible or out of bounds.
- */
-function getLocation() {
-  return new Promise((resolve, reject) => {
-    getLocationBrowser()
-      .then(resolve)
-      .catch(e => {
-        getLocationGoogle()
-          .then(resolve)
-          .catch(reject);
-      });
-  });
-}
-
-function coordsInSanFrancisco(coords) {
-  // These are conservative bounds, extending into the ocean, the Bay, and Daly
-  // City.
-  let bb = {
-    top: 37.820633,
-    left: -122.562447,
-    bottom: 37.688167,
-    right: -122.326927
-  };
-  return coords.lat > bb.bottom &&
-    coords.lat < bb.top &&
-    coords.lng > bb.left &&
-    coords.lng < bb.right;
-}
-
-/**
- * Get location via HTML5 Geolocation API.
- */
-function getLocationBrowser() {
-  return new Promise((resolve, reject) => {
-    if (navigator.geolocation) {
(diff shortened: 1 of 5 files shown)


twolfe2:
@richardxia  Great work!

The way you are passing the updates down to the child components is correct. 

Once we convert the app fully to Redux it will become much easier to pass around props between components.

app/components/App.js
```diff
@@ -3,6 +3,95 @@ import Navigation from './Navigation';
...
+ *
+ * Makes use of both the HTML5 Geolocation API and the Google Maps Geolocation
+ * API. Currently restricts the location to within San Francisco to avoid
+ * inaccurate geolocation results, but this should be removed if more locations
+ * are added.
+ *
+ * @returns A Promise of a location, which is either an object with `lat` and
+ * `lng` properties or an error if location is unavaible or out of bounds.
+ */
+function getLocation() {
```
twolfe2: Typically in React, functions are placed within the component. We don't follow this practice throughout the app but this is a small change that could be made.

richardxia: Got it. Fixed in dac192e.

app/components/Search/ResourcesTable.js
```diff
@@ -138,19 +124,14 @@ class ResourcesTable extends Component {
...
-        location: userLocation
-      });
-      this.loadResourcesFromServer(userLocation);
-    } else {
-      let self = this;
-      this.getLocationGoogle()
-        .then(loc => {
-          self.loadResourcesFromServer(loc);
-        });
+    if (this.props.userLocation) {
```
ZanderSparrow: Probably fine to take this out of the if - there's a hard-coded location passed down and I think the api can handle requests without coords, anyway.

app/components/App.js
```diff
@@ -6,25 +6,113 @@ import ResourcesTable from './Search/ResourcesTable';
...
+   *
+   * Makes use of both the HTML5 Geolocation API and the Google Maps Geolocation
+   * API. Currently restricts the location to within San Francisco to avoid
+   * inaccurate geolocation results, but this should be removed if more locations
+   * are added.
+   *
+   * @returns A Promise of a location, which is either an object with `lat` and
+   * `lng` properties or an error if location is unavaible or out of bounds.
+   */
+  getLocation() {
```
ZanderSparrow: nice organization, looks much better

app/components/Search/ResourcesTable.js
richardxia: I'm not sure if it's safe to take this out. `loadResourcesFromServer` doesn't do a null check before trying to access the `lat` and `lng` of the location, but even if it did, if we decide to expand to more cities, it'll be confusing for the API to return results from everywhere.

PR #159, opened 2017-02-23
Fix: Title will now match category searched for

twolfe2:


achadha:
LGTM other than Travis builds timing out

PR #176, opened 2017-03-19
Achadha/add services

achadha:


app/components/Edit/EditServices.js
```diff
@@ -5,26 +5,47 @@ class EditServices extends Component {
...
 
-		this.state = {};
+		this.state = {
+			services: {},
+			existingServices: props.services.map((service) => {
+				let newService = service;
+				newService.key = service.id;
+				return newService;
+			}),
+			uuid: -1
```
achadha: Keeping track of a unique key for new services. Purely for the key attribute of added components.

PR #177, opened 2017-03-24
Achadha/add notes

achadha:


achadha:
This will have to be cleaned up really badly. Rushing to get stuff in.

PR #208, opened 2017-04-20
Add scrollbars back

richardxia:
This might require testing across multiple browsers to ensure that no regressions have slipped through. There was a bit of a domino effect where changing one thing had many other undesirable effects, but I think I stomped them all out.

## Adding scroll bars back in
Several styles were preventing scrollbars from displaying:
- it looks like we had some webkit-specific styles to explicitly hide the scrollbars by making them have 0 width and be transparent
- the #root element had a height of 100vh set, which meant that it could never be larger than the viewport
- the body and html elements had `position: fixed` set, which prevents scrollbars from appearing

I undid all of those and it now allows scrollbars to appear.

Basically, originally the `<html>` element had a height that was fixed to the viewport height, which prevented normal scrollbars from appearing. After my changes, the `<html>` has a height that reflects the true height of all the page content.

I also got rid of an unnecessary `overflow-y: scroll` on a non-root element, since that was causing a scroll bar to appear in the middle of the page on Edge, even though the element itself was unscrollable because its height was the same as its content. These types of things are difficult for me to detect because OS X will hide scrollbars by default if a mouse isn't attached.

## Fix positioning of map on search results page
On larger screens, the map is displayed to the right of the search results. Previously, when the `<html>` element had a height equal to the viewport, the map would be in a fixed position while only the search results element would internally scroll through its overflowed content.

With my changes, the actual page height is now taller, but this caused the map to get placed in the center of the page, meaning that you'd have to scroll down to see it. To fix this, I told the flexbox container holding both the results and the map to align to the top, not the center, and I added a `position: fixed` to the map.

On smaller screens, the map is displayed above the search results. Previously, you could only scroll the results section of the page, and on a phone, it was kind of annoying to have such little screen real estate available. It looks like my changes actually changed this behavior so that when you scroll, the results "slide over" the map, due the fixed positioning I added. I could have added a media query to remove the fixed positioning on mobile, but I think it actually looks kind of neat, so I left it in.

## Scroll position weird when clicking links
Because the page's height now reflects its true content height and is not restricted to the viewport height, this caused weird behavior when you clicked a link. If you were scrolled down when you clicked a link, then when the next page loaded, it would also be in the same scroll position instead of being at the top. This is a [known issue](https://github.com/ReactTraining/react-router/issues/2019) with react-router, and the reason it hasn't been fixed is because the behavior you'd want depends on the context (if you had a tabbed navigation that wasn't fixed positioned, then you don't actually want to scroll back to the top).

Our version of react-router is kind of old, so I couldn't apply the latest [recommended fix](https://github.com/ReactTraining/react-router/blob/v4.1.1/packages/react-router-dom/docs/guides/scroll-restoration.md), which only works on 4.x. I found an [older workaround](https://github.com/ReactTraining/react-router/issues/2019#issuecomment-256591800) that works on 2.8 and applied it. Basically, it's a bit of JavaScript to scroll to the top on any routing action except when going back to a previous page.

## How this was tested

Here's what I was able to test this on:

- OS X Chrome, latest
- OS X Firefox, latest
- OS X Safari, latest. Actually, the app performance is terrible on Safari. Going back a page using the trackpad gesture causes the page to lockup for several seconds. I just checked out askdarcel.org on Safari, and the problem is not as bad, but going back a page with the trackpad gesture causes the page to completely reload and scroll back to the top of the page. Interestingly enough, this seems to be less of a problem if you click the back button in the browser chrome.
- Android Chrome, latest
- Xcode iOS simulator
- Windows 10 Edge, via VM

I don't have a great way of testing other versions of Internet Explorer, but I could download more VMs from Microsoft. What's the oldest version of IE we support?

### DIFF
--- a/app/components/Search/ResourcesMap.js
+++ b/app/components/Search/ResourcesMap.js
@@ -76,7 +76,9 @@ class Gmap extends Component {
 
   render() {
     return (
-      <div ref="map_canvas" className="map_canvas"></div>
+      <div className="results-map-inner">
+        <div ref="map_canvas" className="map_canvas"></div>
+      </div>
     )
   }
 }
--- a/app/components/Search/ResourcesTable.js
+++ b/app/components/Search/ResourcesTable.js
@@ -1,6 +1,8 @@
 import React, { Component } from 'react';
 import { Link } from 'react-router';
 import Gmap from './ResourcesMap.js';
+import Loader from '../Loader';
+
 import queryString from 'query-string';
 import ResourcesList from './ResourcesList'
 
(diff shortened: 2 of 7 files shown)


richardxia:
Just fixed up the merge conflicts and am going to merge this in now.

PR #216, opened 2017-05-04
Change requests edit ability

twolfe2:
Change requests on askdarcel.org/admin are now editable. To test with the BE pull down `jjfreund/make-changerequests-editable`


richardxia:
It looks like you committed a dummy file you created named "Find Results" containing the output of a grep.

richardxia:
Did `form.html` accidentally sneak in here? Doesn't seem related and it isn't referenced by the code anywhere.

richardxia:
I checked out this branch and tried to run it locally, but I think it broke editing a resource. I tried to make an edit to the resource, but it created a change request that doesn't display properly. Any edits that I had previously made, before checking out this branch, do display properly, though.

For example:

screen shot 2017-05-10 at 7 38 33 pm

The top four were edits I made after checking out this branch, and the bottom one was one that I made before checking out this branch.


Also, when I edited a change request and hit save, it seemed to persist the original value of the change request instead of the new value that I entered in. Do I need to run a specific branch on askdarcel-api for this to work?

richardxia:
> Also, when I edited a change request and hit save, it seemed to persist the original value of the change request instead of the new value that I entered in. Do I need to run a specific branch on askdarcel-api for this to work?

Nevermind about this, I found Joe's branch on api and that works. I'm still seeing the other problem I mentioned, though.

twolfe2:
@richardxia Nice catch. I broke the data service we were using and that caused all of this. I've since fixed it and when I merge in the ability to edit service change requests it will start working again

PR #223, opened 2017-05-17
Make new service requests editable

twolfe2:
This fixes the data service and makes it so that new service requests can be edited on the admin page.


richardxia:
I think you might want to rebase this branch onto master; the diff appears to contain changes that I think have already been merged into master, like the reindentation of all the files.

PR #235, opened 2017-06-04
229 Enforce linting (on some files)

richardxia:
cc @lexholden, who I couldn't add as a reviewer (probably needs to be added as a member of ShelterTechSF?)

I set up ESLint with Airbnb's style guide as our linter, configured Travis to run it, and converted a few of our files to follow the linting rules. I didn't convert all the files because we had almost 2700 linter errors, but I'm hoping the several files I chose to convert are a representative sample.

After running `npm install`, you'll be be able to run eslint with the following command: `./node_modules/.bin/eslint app/`

I tried to minimize the number of custom rules we have, but there was one that I felt was appropriate to override: the one that expected all JSX files to have a `.jsx` file extension. Check out the changes to the files that I updated, and let me know if you think we should override any other rules. A couple rules I'm on the fence about but erred on the side of not overriding are:
- Object keys should never be quoted if not necessary (e.g. prefer `{foo: 'bar'}` over `{'foo': 'bar'}`, which looks a little strange if some keys *need* to be quoted)
- `++` is not allowed, even in a `for` loop (though there is a way to only allow it in `for` loops)

Currently, I've added almost every existing file to `.eslintignore` except for the ones that I updated. I think our development policy moving forward should be to gradually fix up the files that we touch and remove them from `.eslintignore`. This should minimize the amount of work for any one person to do in one PR, and spreads the work across the whole team.


richardxia:
If no one has any objections, I'm going to merge this tonight so that I can start working on adding static type checking.

twolfe2:
@richardxia I'm going to merge this now. About to start work on Add a resource and I want to make sure I'm following the new style guide

PR #241, opened 2017-06-12
Added function to specifically format time for form inputs.

achadha:


jjfreund:
@achadha this needs another patch set...

achadha:
Fixed in different PR

PR #244, opened 2017-06-15
Disable the add button for phone numbers

richardxia:
This PR disables the "Add" button for phone numbers, since we actually haven't even implemented the code path for handling adding new phone numbers: https://github.com/ShelterTechSF/askdarcel-web/blob/c090b215dfc7e8ccf74b73dc7d49b6f7d732b864/app/components/Edit/EditSections.js#L152

See https://github.com/ShelterTechSF/askdarcel-api/issues/172 for what I think we need in general in order to create change requests that represent new database model instances that should get created, as opposed to change requests that only edit existing model instances.


achadha:
LGTM

Issue #236, opened 2017-06-04
Resource hours don't transfer to edit page


twolfe2:
When you click 'Edit' on a resource, the hours from the resource are not pre-populated on the edit page. I think this is being cause by the `value` of the input being incorrectly set

PR #245, opened 2017-06-15
Use the correct format for time inputs.

richardxia:
As @twolfe2 suspected in #236, we're not passing correctly formatted time values into `<input type="time">` fields. We were passing in things that looked like `7:00 PM`, but what it actually wants is `19:00`.

I added a new helper function, `timeToTimeInputValue()`, which takes our API's "concatenated hour-minute" format (e.g. `1900`) and formats it correctly.

An orthogonal problem that I noticed is that the edit page isn't correctly displaying schedules. The edit pages currently has one field per day, but we can have multiple time ranges per day. What we really need is to have multiple fields per day, and to also allow a user to add more time ranges, but I think that's outside the scope of this PR.

### DIFF
--- a/app/components/Edit/EditSchedule.js
+++ b/app/components/Edit/EditSchedule.js
@@ -1,5 +1,5 @@
 import React, { Component } from 'react';
-import { timeToString, stringToTime, daysOfTheWeek } from '../../utils/index';
+import { timeToTimeInputValue, stringToTime, daysOfTheWeek } from '../../utils/index';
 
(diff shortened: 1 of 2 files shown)


achadha:
I had a PR up for this that was failing the Travis build, but yours is clean and everything passes. Logic is the same as what I did, so LGTM.

jjfreund:
Yeah, thanks Avneet. I couldn't merge yours in because Travis was failing.

PR #246, opened 2017-06-15
Removed static note content

jjfreund:


achadha:
While, I'm ok with notes permanently displaying as meowmewomeow, I understand taking this out. LGTM.

PR #248, opened 2017-06-20
Admin Pages tweaks

lexholden:
- [x] Restructured to conform to new wireframe
- [x] Implemented bulkActionHandler and refactored actionHandler to be a bit cleaner/reuse code
- [x] If there was no value in a field, that has been added, show a `{ NEW }` as the old value below the text input for consistent page spacing
- [x] Refactor edited components to pass (most) linting checks, including moving functions to static methods, const not let and others
  - [x] I overrode the stupid rule `no-plusplus` unless anyone has any arguments against.
- [x] Added some sass variables for padding, colors and buttons for future reuse
  - [x] Added some consistent button CSS that can be reused across the app (`.btn class`)
- [x] Look over code with fresh eyes to clean up
- [x] Disabled broken login redirect that made it impossible to access the login pages by redirecting to `/testAuth` (unless i've missed something?)

Outstanding questions
- [ ] The wireframe suggests removal of the accordion styles on the page so the design can be refactored into two columns (the right containing the acceptAll button, the left containing fields and content). I think collapsible headers makes it easier to find specific CR's and the page less overwhelming, though I guess most admins will just scroll through in order. Definitely remove them?
- [ ] Do we need separate edit/approve icon buttons. They do the same thing, ultimately. When you edit, it will commit any changes you have made to the text box, editing the original changeRequest
- [ ] It would make sense that fields like `long_description` should accept some form of markdown/richtext so we aren't stuck with long singleline blocks, at which point we need to not just use on `TextareaAutosize` component

Will look at it again later to clean up, but any suggestions welcome. Closing the pull open from my personal fork.


app/components/Admin/ChangeRequest.js
```diff
@@ -16,23 +14,23 @@ class ChangeRequest extends React.Component {
...
   retrieveModifiedObject() {
-    let changeRequest = this.props.changeRequest;
-    let resource = changeRequest.resource;
-    //"ChangeRequest" is 13 characters, so this will give us the first part of the string
-    let objectType = changeRequest.type;
+    const changeRequest = this.props.changeRequest;
+    const resource = changeRequest.resource;
+    // "ChangeRequest" is 13 characters, so this will give us the first part of the string
+    const objectType = changeRequest.type;
+    let resourceNotes;
```
lexholden: The eslint rule `no-case-declarations` demands that variables are instantiated in the parent block. Agreed it feels messy when it is just assigned once.

Shall I make the rule a warning?

app/components/Admin/Admin.js
```diff
@@ -39,63 +39,66 @@ class Admin extends React.Component {
...
+        changeRequestsLoaded: true,
       });
     });
   }
 
+  bulkActionHandler(action, changeRequests) {
+    return Promise.all(
+      changeRequests.map((changeRequest) => { // Create an action handler for each CR
+        const changeRequestFields = changeRequest.field_changes.reduce((a, c) => { // reduce fields into one object
+          a[c.field_name] = c.field_value;
```
lexholden: Good idea. Done, will log a warning if there is a duplicate, we should probably eventually support some popup/push notification architecture so non devs can debug issues like that can be a followup issue.

.eslintrc.js
```diff
@@ -9,6 +9,7 @@ module.exports = {
     "browser": true
   },
   "rules": {
-    "react/jsx-filename-extension": 0
+    "react/jsx-filename-extension": 0,
+    "no-plusplus": 0,
```
lexholden: Hmm ok, useful to know, did not know the reason for this rule.

Have removed it again, but not removed usage of `++`/`--` in the codebase, can remove them as we progress.

app/components/Admin/ChangeRequest.js
richardxia: I don't have strong opinions about it. I'm guessing it doesn't like the variable being declared inside of the case statement because case statements don't provide scope, so the scope of the variable is actually greater than what it looks.

It looks like you ended up fixing it by making it a const anyway, so let's not change the linter.

PR #263, opened 2017-07-02
Add a resource

twolfe2:
Rather than creating a whole new component for adding a resource, I refactored our current edit resource page to work with adding new resources. 

- [x] Refactor Edit resource to work with adding a new resource
- [x] Confirm the payload data model works for the Back end code


twolfe2:
This is ready to merge except for the ~200 linting errors. @richardxia are we enforcing those for all new PRs now?

richardxia:
> This is ready to merge except for the ~200 linting errors. @richardxia are we enforcing those for all new PRs now?

I think it's your judgment call. I'd like for us to make progress on cleaning up the style of existing files, but if there are too many files to touch in a single PR, then it may be too onerous to update them all. You can add files to ignore to `.eslintignore`.

PR #264, opened 2017-07-03
Services inherit schedule

achadha:


achadha:
Going to need a backend change before we can confirm this can be confirmed on the Admin page. At the moment, it seems like the backend isn't properly serializing the boolean, 'true'. When on the Admin page, I see the updated field with the value, "t", instead of true.

achadha:
This is going to need a backend fix before it's marked as done. On the admin page I'm just seeing the field value "t" instead of true. I expect it's only taking the first bit of the string "true" when converting the incoming payload value to the appropriate field_value.

app/components/Edit/EditSections.js
```diff
@@ -271,7 +278,11 @@ class EditSections extends React.Component {
 
     postSchedule(scheduleObj, promises, uriObj) {
         if(scheduleObj) {
-            this.postObject(scheduleObj, 'schedule_days', promises);
+            if(scheduleObj.inheritSchedule) {
+                //Post schedule with flag set
```
alexanderturinske: Should this do something?

app/components/Edit/EditSections.js
```diff
@@ -60,6 +60,7 @@ class EditSections extends React.Component {
 	    let url = '/api/resources/' + resourceID;
 	    fetch(url).then(r => r.json())
 	    .then(data => {
+            data.resource.services[0].inherit_schedule = true;
```
alexanderturinske: Maybe break this out into two lines so it is more readable? Create a variable that is descriptive so that it isn't just you assigning some property to a random object. Something like the following:
let scheduleService = data.resource.services[0];
scheduleService.inherit_schedule = true;

app/components/Edit/EditServices.js
```diff
@@ -107,8 +114,8 @@ class EditService extends Component {
 		return (
 			<li className="edit--service edit--section">
 				<header className="edit--section--header">
-      		<h4>Service {this.props.index+1}: {this.props.service.name}</h4>
-      	</header>
+      				<h4>Service {this.props.index+1}: {this.props.service.name}</h4>
+      			</header>
```
alexanderturinske: These shouldn't be additionally tabbed.

app/styles/components/_edit.scss
```diff
@@ -232,3 +232,21 @@
...
+	}
+
+	label {
+		color: black;
+		font-size: calc-em(16px);
+	}
+}
+
+.hidden {
+	display: none;
```
alexanderturinske: tab needed

app/styles/components/_edit.scss
```diff
@@ -232,3 +232,21 @@
 	flex-direction: column;
 	margin-top: calc-em(60px);
 }
+
+.checkbox-label {
+	display: flex;
+	flex-direction: row;
```
alexanderturinske: tabs needed

lexholden:
We had some discussions on the implementation for this a couple of weeks back. Given the new heirarchy `org => program => service` with addresses and schedules inheriting downwards, I think this is probably obsolete and should be moved redux anyway.

Closing this, since it hasn't been touched in almost a year.

PR #265, opened 2017-07-07
Enable adding multiple phones; also display and make editable phone service_type.

richardxia:
Depends on https://github.com/ShelterTechSF/askdarcel-api/pull/180

I added a new API endpoint for adding new phone numbers on the edit page. I also noticed that the `service_type` of the phone number wasn't hooked up, so I added it in both the display and edit views. One thing of note is that the linksf data appears to have put phone number extensions in the `service_type` field for many numbers. This will at least have the side effect of displaying them and making them editable, but we should probably be putting the extensions into the number field.

screen shot 2017-07-06 at 7 55 28 pm

screen shot 2017-07-06 at 7 56 16 pm

I attempted to lint the files that I edited here, but that's proving to be quite the endeavor because of how long the EditSections.js and ResourceInfos.js files are. I'm going to try to add linting on top of this, but if I don't get to that in time, I think this can merge in first. We're definitely going to want to do some extensive manual QA if I do get this linted because it's forcing me to make pretty large changes to EditSections.js.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -12,9 +12,7 @@ app/components/Auth/RequireAuth.js
 app/components/Edit/EditAddress.js
 app/components/Edit/EditCollection.js
 app/components/Edit/EditNotes.js
-app/components/Edit/EditPhones.js
 app/components/Edit/EditSchedule.js
-app/components/Edit/EditSections.js
 app/components/Edit/EditServices.js
 app/components/Find/CategoryItem.js
 app/components/Find/FindHeader.js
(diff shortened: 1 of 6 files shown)


richardxia:
Just finished linting the rest of the files I touched. For an easier diff of the actual features added, see https://github.com/ShelterTechSF/askdarcel-web/compare/eda62bbc3c1d3d21608df0069370b60c12668cc8...72de616c870048ee13337854f287a92deec72b3c

PR #266, opened 2017-07-12
Achadha/associate services with category

achadha:


app/components/App.js
```diff
@@ -2,6 +2,7 @@ import React, { Component } from 'react';
 import Navigation from './Navigation';
 import CategoryPage from './Find/FindPage';
 import ResourcesTable from './Search/ResourcesTable';
+import 'react-select/dist/react-select.css';
```
achadha: I assume I can just add something to the webpack config to import this CSS?

richardxia: It looks like this works fine. I found this thread that shows how you can do it with the CSS modules-loader, but we're not using it: https://github.com/JedWatson/react-select/issues/1324#issuecomment-256591621

app/components/Admin/ProposedService.js
```diff
@@ -86,6 +95,16 @@ class ProposedService extends React.Component {
     }
     return scheduleOutput;
   }
+
+  renderCategoryFields() {
+    return (
+      <div key={"cat"} className="request-entry">
+        <p className="request-cell name">Categories</p>
+        <textarea readOnly className="request-cell value" value={this.state.categories} />
```
richardxia: So this is readonly? That's probably fine for now, but we should let the data team know that they won't be able to edit the proposed categories.

app/components/Edit/CategoriesDropdown.js
```diff
@@ -0,0 +1,48 @@
+import React, { Component } from 'react';
+import Select from 'react-select';
+import * as dataService from '../../utils/DataService';
+
+class CategoriesDropdown extends Component {
+    constructor(props) {
+        super(props);
+
+        this.state = {
+            selectedValue: '',
```
richardxia: Should this be renamed to `selectedValues` and initialized to `[]`? Looks like when you update it, it gets set to an array of values.

app/components/Admin/ProposedService.js
```diff
@@ -86,6 +95,16 @@ class ProposedService extends React.Component {
     }
     return scheduleOutput;
   }
+
+  renderCategoryFields() {
+    return (
+      <div key={"cat"} className="request-entry">
```
richardxia: I don't really know React that well, but do you need a `key` if this isn't really a list of stuff?

achadha: Oops you don't. Bad job copy-pasting and then editing lol

app/components/Admin/ProposedService.js
achadha: Yeah, after talking with joe we realized it was complicated. All of the other fields are string values, but here, you want to edit the object. As an enhancement, I can write the stuff to map the edited strings to the category objects.

PR #273, opened 2017-07-15
Change services render for email to be type email instead of URL

trucnguyen:


richardxia:
Sorry for letting this go unreviewed for so long! Changes LGTM

PR #279, opened 2017-08-07
[WIP] Subnavigation for individual service listings

lexholden:


lexholden:
From discussing this on slack with @trucnguyen, the debate is whether to:

1. Stick with anchor links and directly reference the ID of the service
2. Implement an onclick event for the links, which would enable `document.querySelector().scrollIntoView()` for smooth scrolling

I think the convention is usually both, since 2 requires polyfills for older browsers. The click event can `e.preventDefault` to still support old crappy browsers.

(also @trucnguyen I'm pretty sure we can directly use anchor links rather than ref the exisiting static links already do, just make sure it's done using `id`)

trucnguyen:
It's functioning now at least the anchor links in page are working.

I did get it to execute the onclick callback which triggers scrollToElement function. I struggled with preventDefault since it's a Synthetic Event? I still have an incomplete understanding of bindings, context `this`, and event handling.

What seemed to be a simple UI interaction introduced me to a lot of cool React concepts and it was fun to work through. Excited to learn more 😄 

PTAL @lexholden

PR #280, opened 2017-08-09
Refactor of the design/structure of the admin pages to fix various bugs and optimize

lexholden:
- Split changeRequests page into 2 panes and moved to `/admin/changes`
- Style improvements to ChangeRequest
  - Added `NEW` helper next to field changes that don't have a previous value
  - Colored buttons and previous text in alignment with the TextArea
  - Added title of the resource and link to the resource page
  - Split up changes into sections for the resource and services, with a header for services.
  - Added functions for changing the layout of different types of changeRequest
    -  as yet unused, full implementation is a followup task
- Disabled current glitchy auth implementation that refreshes tokens given by the API. It seems to given invalid (identical) tokens on some API requests. Will open a seperate issue to track.
  - Had to remove the bulk approve feature, since simulataneous requests cause auth to invalidate
- Replaced old `/admin` page with a basic landing page, to be replaced with a central dashboard
- A bunch of linting fixes
- Still a lot of improvments to make. Tracking followup tasks with #325


lexholden:
This has a bunch of stuff I'd like to improve, but should work and is reviewable @jjfreund @lgarofalo @richardxia

Issue #288, opened 2017-08-31
Enable category editing for services


jjfreund:

trucnguyen:
I'll take a shot at this

PR #299, opened 2017-09-09
Allow category editing for existing services

trucnguyen:
Fixes Issue #288 
Show dropdown component in edit mode instead of rendering category list.
Not sure if this resulted in a correct change request since my admin UI isn't loading


lgarofalo:
Change LGTM

PR #302, opened 2017-09-10
Only display top-level categories on home page.

richardxia:
This is the front-end change that goes with https://github.com/ShelterTechSF/askdarcel-api/pull/195

This filters categories for only the top-level categories on the home page. This is actually safe to merge and deploy before the API change because the API currently ignores any GET parameters that you pass to it, so until the backend is deployed, and the migration runs, we'll still see the old categories list.

The other change I made was to change the CSS styles so that we allow `flex-wrap` on the category items because we now have enough categories that it looks squished when we force them to fit on one row.


webpack.config.js
```diff
@@ -49,7 +49,7 @@ module.exports = {
         use: ['style-loader', 'css-loader', 'sass-loader']
       },
       {
-        test: /\.(ttf|otf|eot|svg|woff(2)?)(\?[a-z0-9]+)?$/,
+        test: /\.(ttf|otf|eot|woff(2)?)(\?[a-z0-9]+)?$/,
```
richardxia: This was picking up .svg files and putting them in a folder named `fonts`, which was confusing, so I just removed this because we don't have any custom fonts anyway.

PR #310, opened 2017-09-15
Handle null categories on CategoriesDropdown.

richardxia:
My last PR broke the ability to add a new service to a resource because I wasn't properly handling the case where you might pass in `null` to the CategoriesDropdown component. Now I use the `defaultProps` member of the React component to set the default value to `[]`. I manually tested that this fixes the issue for me locally.


lgarofalo:
LGTM

richardxia:
Thanks for the quick review!

PR #315, opened 2017-10-02
24 hour toggle for edit pages

twolfe2:
## Summary
This adds a 24 hour toggle to each day on the edit resource page and will help to keep our convention for 24 hours consistent (`0` to `2359`)


app/components/Edit/EditScheduleDay.js
```diff
@@ -12,31 +13,52 @@ class EditScheduleDay extends Component {
...
+        }
       </li>
     );
   }
 
   render() {
+    let { dayHours } = this.props;
     let day = this.props.day;
+    let is24Hours = false;
+    let is24Toggle = null;
```
richardxia: This variable seems to be unused. (Maybe also run the linter on this file?)

app/components/Edit/EditScheduleDay.js
```diff
@@ -12,31 +13,52 @@ class EditScheduleDay extends Component {
...
+
+    if (dayHours[0].opens_at === 0 && dayHours[0].closes_at === 2359) {
```
richardxia: This pattern seems to reoccur a lot. Not blocking this PR, but it'd be a good idea to create a class that represents a `ScheduleDay` and give it methods like `is24Hours()` and `toggle24Hours()` so that this logic isn't repeated everywhere.

app/components/Edit/EditSchedule.js
```diff
@@ -98,6 +121,31 @@ class EditSchedule extends Component {
     return timeToTimeInputValue(time, true);
   }
 
+  toggle24Hours(day, is24Hours) {
```
richardxia: It'd be good to add a comment that briefly describes what `is24Hours` should be. Otherwise it's not clear without reading the code whether it represents the current value *before toggling* or the desired value *after toggling*.

app/components/Edit/EditSchedule.js
```diff
@@ -16,11 +27,23 @@ function buildSchedule(schedule) {
...
-      curr.closeChanged=false;
-      currSchedule[curr.day] ? currSchedule[curr.day].unshift(curr) : currSchedule[curr.day] = [curr];
+      currDay = curr.day;
+      if (!is24Hours[currDay]) {
+        // Check to see if any of the hour pairs for the day
+        // indicate the resource/service is open 24 hours
+        // if there is a pair only have that in the schedule obj
+        if (curr.opens_at === 0 || curr.closes_at === 2359) {
+          is24Hours[currDay] = true;
+          currSchedule[currDay] = [{ opens_at: 0, closes_at: 2359, id: curr.id }];
```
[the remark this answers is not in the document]
twolfe2: We don't need `scheduleId` here since there is already an `Id` present. We only need `scheduleId` if we are creating a schedule that doesn't already exist. Added a comment to clarify this

app/components/Edit/EditSchedule.js
```diff
@@ -16,11 +27,23 @@ function buildSchedule(schedule) {
...
     schedule.schedule_days.forEach((curr) => {
-      curr.openChanged=false;
-      curr.closeChanged=false;
-      currSchedule[curr.day] ? currSchedule[curr.day].unshift(curr) : currSchedule[curr.day] = [curr];
+      currDay = curr.day;
+      if (!is24Hours[currDay]) {
+        // Check to see if any of the hour pairs for the day
+        // indicate the resource/service is open 24 hours
+        // if there is a pair only have that in the schedule obj
+        if (curr.opens_at === 0 || curr.closes_at === 2359) {
```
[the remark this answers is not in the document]
twolfe2: Nice catch

PR #320, opened 2017-10-19
chore(nginx): updated docker build to use latest stable nginx and removed https rewrite

lgarofalo:


docker/tiller/templates/askdarcel-web.conf.erb
```diff
@@ -4,10 +4,6 @@ server {
 
     root /app/askdarcel;
 
-    if ($http_x_forwarded_proto != 'https') {
```
lgarofalo: I removed this because it was causing an issue with doing a HTTP or HTTPS healthcheck from ELB, and this can be better handled at the edge network anyway.

PR #322, opened 2017-10-26
fix(cookies): send cookies with API calls to work with auth

lgarofalo:
Verifying that this solves authentication issue, if so then I will update rest of application to do this.


lgarofalo:
I've verified this locally and an moving to staging to confirm.

PR #326, opened 2017-11-03
Add end to end testing

richardxia:
Don't merge yet, still working out the kinks with getting it working in Travis.

In case if you want to try using these for the QA party tonight, here's my branch where I've added [TestCafe](https://devexpress.github.io/testcafe/) end-to-end tests.

### Quick summary of what TestCafe is and how it works
It's a framework for running end-to-end tests (read: real browser tests) that injects your tests onto an existing web page. Architecturally, they spin up a lightweight proxy server that wraps your web page, and when you connect a browser to the proxy server, it serves the requested page with the test driver injected into it.

It's essentially an alternative to writing Selenium tests, and I've found it nice to use because it mimics many of the common HTML5 DOM APIs and because they've added a lot of reasonable default behavior that Selenium lacks, such as properly waiting for events to finish running and for elements to appear before running your assertions.

### How to run
If you are *not* using Docker and all the services are bound to localhost, then you should just be able to run:

```sh
$ npm run testcafe -- --skip-js-errors chrome testcafe/*.js
```

If you are using Docker, then you'll need to run it somewhat like this:

```sh
$ docker run --rm -v $PWD:/usr/src/app -p 1337:1337 -e BASE_URL=>URL_TO_ASKDARCEL_WEB> -w /usr/src/app node npm run testcafe -- --skip-js-errors remote --skip-js-errors --hostname localhost --ports 1337,1338
```

This will spin up a web server at http://localhost:1337/ and print out a URL to use. You should manually enter it into your browser to start the tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,6 +25,7 @@ before_install:
 
 script:
   - npm run lint
+  - ./travis-testcafe.sh
 
 after_success:
   - tools/docker-build.sh
--- a/app/components/Edit/EditPhones.js
+++ b/app/components/Edit/EditPhones.js
@@ -67,4 +67,5 @@ EditPhone.propTypes = {
 };
 
 const EditPhones = editCollectionHOC(EditPhone, 'Phones', {});
+EditPhones.displayName = 'EditPhones';
 export default EditPhones;
--- a/app/styles/components/_org-page.scss
+++ b/app/styles/components/_org-page.scss
@@ -41,6 +41,8 @@
 		margin: 0 auto;
 		padding: 0 calc-em(20px);
 		transform: translateY(-96px);
+		// Prevent empty space to the right of image from covering other elements on page
+		pointer-events: none;
 		&--img {
 			@include aspect-ratio(1, 1);
 		  width: calc-em(130px);
--- a/package-lock.json
+++ b/package-lock.json
@@ -4,6 +4,24 @@
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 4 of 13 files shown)


richardxia:
Okay, I've finally figured out why the tests are flaky *only* on Travis/SauceLabs. The page looks like this when viewed on one of those services:

screen shot 2017-11-26 at 11 27 51 am

The map and the street view picture don't show up because the Google APIs fail, due to the fact our API key hasn't whitelisted the URL that they use to connect to the Webpack dev server. More importantly, as you can see from my screenshot, the actual width of the div containing the placeholder street view image spans the entire width of the page. This actually prevents the "Make Edits" button along with the "Get Directions" button from being clickable, since the street view div is above them in the Z-index stack.

This is what it looks like on the real site:

screen shot 2017-11-26 at 11 32 41 am

Notice that a real street view image is smaller than the placeholder image, which shrinks the parent div and allows both buttons to be clickable. What's interesting is that on the real site, that blue region on the map in my screenshot is actually unclickable; you can't drag the map or anything within that region.

I'll see if I can figure out how to make the parent div of the street view image not span the whole width of the page in order to not cover up the buttons or even the map.

richardxia:
I ended up setting the CSS property `pointer-events: none`, which will allow clicks to pass through the element: https://github.com/ShelterTechSF/askdarcel-web/pull/326/commits/26b2f6892d30c56cc5d44176ea6e568c8b7d9fd1

This branch should be good to merge now, since it looks like the TestCafe tests are finally passing.

Issue #321, opened 2017-10-22
Trim latitude and longitude values sent from browser to API


lgarofalo:
Example seen after clicking into a category 

Request URL:https://staging.askdarcel.org/api/resources?category_id=2&lat=37.7813454&long=-122.497668
Request Method:GET

alexanderturinske:
Trim the values down to a particular decimal place? If so, how many?

lgarofalo:
Unsure, looks like 4 decimal places is appropriate.

https://www.thethingsnetwork.org/forum/t/best-practices-when-sending-gps-location-data/1242

There's a side issue here as well if you are looking into the GPS data here, 
https://github.com/ShelterTechSF/askdarcel-web/issues/128

Also in general there is some quirky behavior in general with the accuracy of the location, there might be some better things we can do with the google geolocation API

alexanderturinske:
I'm not sure if this is the right place to round these coordinates; I couldn't figure out where they were being sent in as props.

alexanderturinske:
Ah, found the right place to do this. It is in App.js

lexholden:
Merged into master. Closing.

PR #328, opened 2017-11-12
Issue #321: Trim coordinates being sent api on user's location

alexanderturinske:
- Created rounding function based on a blog post about rounding in JS (http://www.jacklmoore.com/notes/rounding-in-javascript/)
- Rounded functions to the fourth decimal place when retrieved from geolocation


lgarofalo:
LGTM

PR #332, opened 2017-12-03
Open now filter fix

twolfe2:
In the past, the `Open Now` filter on the resource list page did not account for when a resource was open past midnight on a particular day. 

For example, imagine it's Sunday at 2am and the current resource is open from 8pm-4am on Saturday. In this case, we would still want to display the resource as open even if the schedule is for the day before (Saturday). 

Now, I've added special logic to handle this case and confirmed that this is now working as intended.

### DIFF
--- a/app/components/Search/ResourcesRow.jsx
+++ b/app/components/Search/ResourcesRow.jsx
@@ -44,6 +44,14 @@ class ResourcesRow extends Component {
     let closingOpeningTimes = {};
     let currDay = days[day];
     let result = { open: false };
+    let resource = this.props.resource;
+
+    if (resource.isOpen) {
+      return {
+        open: true,
+        time: timeToString(resource.openUntil),
+      };
+    }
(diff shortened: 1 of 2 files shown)


app/components/Search/ResourcesRow.jsx
```diff
@@ -147,7 +148,7 @@ class ResourcesRow extends Component {
             <div className="entry-meta">
               <p className="entry-organization">{serviceName}</p>
               <p className="entry-description">{resourceDescription}</p>
-              <p className="entry-hours">{open ? `Open until ${time}` : time ? `Closed until ${time}` : 'No hours found for this location'}</p>
+              <p className="entry-hours">{open ? `Open until ${time}` : time ? `Closed` : 'No hours found for this location'}</p>
```
jjfreund: I forget, did we agree to remove this functionality?

PR #333, opened 2017-12-03
Added hashing to bundle.js for invalidation

twolfe2:
Added hashing for `bundle.js` in order to prevent caching issues when we push changes


lexholden:
@twolfe2 is this still needed? Can ship in this release if so.

alexanderturinske:
I'm pretty sure these tests are failing because there is a problem with the version of chrome they were run against (v63). I was having trouble myself (v66) but the latest commit that was pushed up ran against v67 and all the tests pass. I updated my chrome to v67 and everything is good.

PR #334, opened 2017-12-04
Deactivation functionality

twolfe2:
Added the ability to deactivate both resources and services via the edit pages

### DIFF
--- a/app/components/Edit/EditSections.js
+++ b/app/components/Edit/EditSections.js
@@ -332,11 +332,19 @@ class EditSections extends React.Component {
   handleDeactivation(type, id) {
     let path = null;
     if (type === 'resource') {
-      path = `api/resources/${id}`;
+      path = `/api/resources/${id}`;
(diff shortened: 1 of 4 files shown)


app/components/Edit/EditSections.js
```diff
@@ -328,6 +329,24 @@ class EditSections extends React.Component {
...
+  handleDeactivation(type, id) {
+    let path = null;
+    if (type === 'resource') {
+      path = `/api/resources/${id}`;
+    } else if (type === 'service') {
+      path = `/api/services/${id}`;
+    }
+    dataService.APIDelete(path, { change_request: { status: "2" } })
+    .then(() => {
+      alert('Successfully deactivated! \n \nIf this was a mistake, please let someone from the ShelterTech team know.')
```
alexanderturinske: Was there a reason that you chose alert instead of confirm here? It would prevent accidental clicks.

PR #339, opened 2017-12-07
Fixed adding phone numbers

twolfe2:
Bug fix for adding phone numbers


richardxia:
Can you remove the `.skip` from [this TestCafe test](https://github.com/ShelterTechSF/askdarcel-web/blob/c69c1d2ed660194612203d8eb5ac173ab811749e/testcafe/editResource.js#L80)? That test should be fixed with this change.

Issue #345, opened 2017-12-10
Add confirmation modal for deactivating a resource/service


twolfe2:
To prevent people from accidentally deactivating a resource/service, we should add a confirmation modal.

Issue #330, opened 2017-11-16
Add "change received" feedback


jjfreund:
When a change request is submitted, feedback indicating that the change was accepted should be returned to the user.

alexanderturinske:
Similarly, there should a few more things changed on the Edit page:
- a "Cancel Changes" button
- a warning/confirmation modal if a user tries to navigate away from the page Edit page with changes that haven't been saved
- #283 relates to this ticket

trucnguyen:
Should this match the specs @Garance91540 put together here for user story "When I hit submit, the page changes but there's no feedback that it worked" https://sheltertech.quip.com/8duuARDxODjT ? 

cc @derekfidler

derekfidler:
No, I’d prefer to use a toast style notification, as opposed to a modal. A user should not need to take action to remove the notification as garance had it, it would be better if it automatically dismissed.

Sent from my mobile

On Jan 7, 2018, at 14:03, Truc <notifications@github.com> wrote:

Should this match the specs @Garance91540 put together here for user story "When I hit submit, the page changes but there's no feedback that it worked" https://sheltertech.quip.com/8duuARDxODjT ?

cc @derekfidler

Garance91540:
@derekfidler the 2 last wireframes shared on Dec 12 and Nov 29 are toast notifications as suggested. 

https://sheltertech.quip.com/8duuARDxODjT

derekfidler:
My mistake, @garance I didn't see the final mockups when i was on mobile.
I've moved the mockups out of chat and into the document. You can see the
mockups here https://sheltertech.quip.com/8duuARDxODjT


Derek Fidler
Product Lead, ShelterTech
Product Design, Bebo

Follow me on Twitter @derekfidler <http://twitter.com/derekfidler>
www.derekfidler.com
df@derekfidler.com <derek@derekfidler.com>
415.857.0879

PR #346, opened 2017-12-14
Issue #330, #345: Add discard changes button and confirmation modal

alexanderturinske:
#330 : Add "Discard Changes" button on resources
#345 : Add confirmation modal for deactivation and discard changes button


alexanderturinske:
@richardxia It looks like saucelabs tests are breaking on things I didn't touch; could you take a look?

richardxia:
It looks like the breakages are related to changes you made. I figured out that you can watch the video recording on SauceLabs from the "Watch" tab (https://saucelabs.com/beta/tests/48b0aa23cea44c5fb7bc39701fcc311e/watch#2), and if you watch the video, it looks like the tests get stuck when trying to find the "Save changes" button. This is because the [TestCafe selector](https://github.com/ShelterTechSF/askdarcel-web/blob/1161831149279ac75fc13d5c01f31ebd6bab2d2c/testcafe/pages/EditPage.js#L32) I wrote depended on the CSS class, and it looks like you changed the CSS class on the button. Can you update the selector so that it can find the button again?

app/components/Edit/EditSections.js
```diff
@@ -330,21 +337,23 @@ class EditSections extends React.Component {
...
+      let path = null;
+      if (type === 'resource') {
+        path = `/api/resources/${id}`;
+      } else if (type === 'service') {
+        path = `/api/services/${id}`;
       }
-    });
+      dataService.APIDelete(path, { change_request: { status: "2" } })
+      .then(() => {
+        alert('successfully deactivated! \n \nif this was a mistake, please let someone from the sheltertech team know.')
```
richardxia: Can you use proper capitalization here? Since this is user-facing, we should make it look professional.

app/components/Edit/EditSections.js
```diff
@@ -587,10 +596,16 @@ class EditSections extends React.Component {
 
   render() {
     let resource = this.state.resource;
-    let actionButtons = [<button className="edit--aside--content--submit" disabled={this.state.submitting} onClick={this.handleSubmit}>Save changes</button>,
-    <button className="edit--aside--content--deactivate" disabled={this.state.submitting} onClick={() => this.handleDeactivation('resource', resource.id)}>Deactivate</button>];
+    let actionButtons = [
+      <button className="edit--aside--content--button" disabled={this.state.submitting} onClick={this.handleSubmit}>Save changes</button>,
```
richardxia: Can we use consistent capitalization on button text? From looking on other pages, it looks like we generally capitalize every word besides articles like "as".

app/components/Edit/EditSections.js
alexanderturinske: Definitely! Good catch!

app/components/Edit/EditSections.js
alexanderturinske: Definitely! Good catch!

alexanderturinske:
Ahh, you are totally right; it was definitely my fault. I will update the test; thank you!

PR #347, opened 2017-12-19
Added certified by hap icon and button to certify

lexholden:
- Added basic 'HAP Certify' button to the Edit Page to call the certify API
  - We have no consistent button styles, only hyper specific classes, so margin of buttons is messy and needs a refactor
  - Currently calling POST `/resource/:id/certify` in staging returns a 404, so this button doesn't work properly.
- Added basic "HAP Certified" icon to the resource page
  - just added the icon to all pages for now, with no condition on the API state, since it can't be changed.


app/components/Resource/Resource.js
```diff
@@ -87,6 +96,7 @@ class Resource extends Component {
               <div className="org--main--header--description">
                 <header>About this resource</header>
                 <p>{resource.long_description || resource.short_description || 'No Description available'}</p>
+                <pre>{JSON.stringify(resource, null, 2)}</pre>
```
richardxia: We probably should have deleted this line.

PR #362, opened 2018-01-07
Warning for leaving edit pages if changes have been made

twolfe2:
A confirmation modal will now show when a user tries to leave an edit page they have made changes too.


richardxia:
The TestCafe tests are now failing because of the modal. Should the modal appear even if someone clicks the save button? If we do want a confirmation modal for saving changes, then we at least need to change the text.

derekfidler:
I think a confirmation for any change, failed change, or confirmation of leaving the page while incomplete is appropriate. 

On Jan 7, 2018, at 15:11, Richard Xia <notifications@github.com> wrote:

The TestCafe tests are now failing because of the modal. Should the modal appear even if someone clicks the save button? If we do want a confirmation modal for saving changes, then we at least need to change the text.

twolfe2:
Ok, I’ll need to update it so that modal doesn’t show on save because the
message it shows would be confusing for that case.

PR #363, opened 2018-01-07
Fix mobile Safari search

richardxia:
I don't own an iPhone, so I tested this using the iOS simulator on my Macbook. I was able to reproduce the issue with the search page not working on iOS Safari. It seems like there's actually an invisible field representing the search location which is never shown. This is never even shown on a desktop browser because both the default style and the media queried style use `display: none`, so it's never displayed.

My hypothesis as to why causes a problem on iOS is because normally the "return" key on the iOS keyboard will take you to the next field on a form, but in this instance the next field is invisible, so it is unable to select it. By removing the HTML for the location field altogether, I can confirm that hitting the "return" key *will* submit the form, which I assume is because there are no more fields left to fill out.

I think it's safe to remove that search location field because the React function that's bound to the form doesn't even read from it, so it just looks like it was never fully implemented.

I *tried* to write a TestCafe test to exercise this bug, but unfortunately, the TestCafe API for typing on the keyboard seems to bypass this problem somehow (I assume it's using some browser API for sending keys, and not the OS on-screen keyboard), so it's not of any use with respect to preventing this problem in the future. That said, having more tests is better in general, so I'll leave the test around in the test suite.


twolfe2:
LGTM

richardxia:
I decided to be a good citizen and lint and fix the style errors in the one JS file I touched. It should still be functionally equivalent, so I'll merge as soon as Travis passes.

Issue #175, opened 2017-03-16
Clicking Ask Darcel logo to go home causes browser page load


richardxia:
If you click the Ask Darcel logo on the top left, it causes the browser to make a full page load instead of just rendering the home page entirely within React. In addition to being a lot slower, it also has the undesirable effect of causing the user's location to be recomputed, which affects the cacheability of API requests. See https://github.com/ShelterTechSF/askdarcel-api/pull/147#issuecomment-286958850

alexanderturinske:
I'll work on this.

PR #365, opened 2018-01-08
Issue #175: Replace a-tags with Link-tags

alexanderturinske:
- increases performance
- allows for caching of API requests
- prevents user's location from being recomputed


lexholden:
(My bad, on reevaluation I realised your comments are actually about the tag change, had lazily not occured to me that page hard refreshed.)

PR #366, opened 2018-01-08
Cleaning up HAP Certified functionality

lexholden:
For HAP Certified to hold some weight, needs to be restricted to authenticated users after implementation of a more general auth/login system #367 

- Removed old junk console log code
- Adds new `HAP Approve` button
- Unverified resource showing `HAP Approve` button:
screenshot 2018-01-08 16 14 22 2
- After button is clicked, it disappears and `HAP CERTIFIED` icon appears:
screenshot 2018-01-08 16 15 24 2

### DIFF
--- a/app/components/Resource/Resource.js
+++ b/app/components/Resource/Resource.js
@@ -22,6 +22,7 @@ class Resource extends Component {
     super(props);
     this.state = { resource: null };
     this.verifyResource = this.verifyResource.bind(this);
+    this.hapCertify = this.hapCertify.bind(this);
   }
 
   componentDidMount() {
@@ -45,6 +46,21 @@ class Resource extends Component {
         // TODO: Do not use alert() for user notifications.
         if (response.ok) {
           alert('Resource verified. Thanks!');  // eslint-disable-line no-alert
+        } else {
+          alert('Issue verifying resource. Please try again.');  // eslint-disable-line no-alert
+        }
+      });
+  }
+
+  hapCertify() {
+    dataService.post(`/api/resources/${this.state.resource.id}/certify`)
+      .then((response) => {
+        // TODO: Do not use alert() for user notifications.
+        if (response.ok) {
+          alert('HAP Certified. Thanks!');  // eslint-disable-line no-alert
+          const resource = this.state.resource;
+          resource.certified = response.ok;
+          this.setState({ resource });
         } else {
           alert('Issue verifying resource. Please try again.');  // eslint-disable-line no-alert
         }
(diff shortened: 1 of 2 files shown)


richardxia:
This doesn't seem to touch the old HAP approve button that's on the Edit page. Do we still want that there? Otherwise, this LGTM. I clicked rebuild on Travis because I'm not sure why the Sauce authentication failed.

lexholden:
I added it as a separate button because technically "verified by HAP" and "Mark as Correct" make different api calls so I figured it was worth keeping both. Particularly since long term hap verification will only be visible to logged in users.

If the tests fail again will investigate.

richardxia:
Tests failed again, but I'm not sure if it's related to this PR since it's still complaining about the Sauce credentials. I can take a look when I'm home tonight. One thing you might want to try is rebasing this onto master, in case if there are changes related to Sauce that you need to grab from master.

derekfidler:
I haven't looked at the branch, but I think it's okay to have separate
buttons. But I think these buttons should only appear on the resource page,
not the edit page. I'm not sure that was clear in the previous comments.
That's because the UX is weird if you go to Edit, then make changes, then
have to decide which button to push to complete the changes.

Derek Fidler
Product Lead, ShelterTech
Product Design, Bebo

Follow me on Twitter @derekfidler <http://twitter.com/derekfidler>
www.derekfidler.com
df@derekfidler.com <derek@derekfidler.com>
415.857.0879

lexholden:
@derekfidler I have added screenshots to the pull above, let me know if you can think of design improvements.

derekfidler:
@lexholden those look good. thanks.

richardxia:
This branch passes for me when I push a copy of it to sheltertechsf/askdarcel-web, so after thinking about it for a bit, I think the problem is that this branch lives in @lexholden's repo and for security reasons, Travis doesn't allow builds triggered from forks to use any secret environment variables. We can probably let this merge without a green build because I manually ran it on a different branch, but in the future we should only push branches to this repo.

app/components/Resource/Resource.js
```diff
@@ -151,6 +166,15 @@ class Resource extends Component {
...
               </button>
+              {
+                !resource.certified &&
+                <button
+                  className="org--aside--content--button"
+                  onClick={this.hapCertify}
+                >
+                  HAP Approve
+                </button>
+              }
```
alexanderturinske: Nice solution.

lexholden:
Closing this and opening a new pull not from my fork then. Agreed using lots of forks is often messier than it is helpful.

PR #368, opened 2018-01-09
Cleaning up HAP Certified functionality

lexholden:
Replaces #366 so that tests pass. See that pull for description/screenshots.


app/components/Resource/Resource.js
```diff
@@ -51,6 +52,21 @@ class Resource extends Component {
       });
   }
 
+  hapCertify() {
+    dataService.post(`/api/resources/${this.state.resource.id}/certify`)
+      .then((response) => {
+        // TODO: Do not use alert() for user notifications.
```
alexanderturinske: Nice

alexanderturinske: Nice

PR #384, opened 2018-02-05
Delete Note 

twolfe2:
Adds the ability to delete notes from the edit page. 

TODO: Still needs some design love


twolfe2:
Ready to go!

Issue #386, opened 2018-02-06
Refactor EditServices to align with coding standards


alexanderturinske:
Few things wrong with the file:
- multiple components in one file
- file is long
- unused methods
- inefficient coding

alexanderturinske:
I don't understand why this wasn't automatically closed with the merge of https://github.com/ShelterTechSF/askdarcel-web/pull/387, but I am closing.

PR #387, opened 2018-02-08
Issue #386: Refactor EditServices to conform with coding standards

alexanderturinske:
- break out into several reusable components
- fix tabing/spacing
- use updated ES6
- remove unused methods

### DIFF
--- a/app/components/Edit/EditServices.js
+++ b/app/components/Edit/EditServices.js
@@ -1,42 +1,53 @@
 import React, { Component } from 'react';
-import Loader from '../Loader';
-import EditNotes from './EditNotes';
-import EditSchedule from './EditSchedule';
+import PropTypes from 'prop-types';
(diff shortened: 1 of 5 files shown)


app/components/Edit/EditServices.js
```diff
@@ -1,206 +1,97 @@
...
   }
 
+  /* @method addService
+   * @description Creates a brand new service
+   */
   addService() {
-    let existingServices = this.state.existingServices;
-    let newUUID = this.state.uuid - 1;
+    const { existingServices } = this.state,
+      newUUID = this.state.uuid - 1;
```
richardxia: I know you didn't write this line originally, but this is definitely not a safe way of generating a new UUID.

alexanderturinske: Hahaha, yeah; for sure.

PR #400, opened 2018-02-20
Fix for phone numbers on resource creation

twolfe2:
Made it so front end passes back phone numbers upon resource creation. 

Still need to ensure back end can support the formatting of the phone numbers in the payload:
image


richardxia:
Looks like this also contains the changes in https://github.com/ShelterTechSF/askdarcel-web/pull/399? Either way, I think this LGTM. I haven't run the code locally, but I believe that the backend doesn't do any validation of the phone number.

Issue #405, opened 2018-03-01
Cannot build with Docker due to error


alexanderturinske:
Run docker-compose run --rm web npm install

Error
gyp verb download contents checksum {"node-v8.4.0-headers.tar.gz":"724d091c6610255ece8c310e7e3924f0e675d1486d63bca34e7ac0c8c8e07497"}
gyp verb validating download checksum for node-v8.4.0-headers.tar.gz (724d091c6610255ece8c310e7e3924f0e675d1486d63bca34e7ac0c8c8e07497 == 724d091c6610255ece8c310e7e3924f0e675d1486d63bca34e7ac0c8c8e07497)
gyp verb get node dir target node version installed: 8.4.0
gyp verb build dir attempting to create "build" dir: /usr/src/app/node_modules/node-sass/build
gyp verb build dir "build" dir needed to be created? /usr/src/app/node_modules/node-sass/build
gyp verb build/config.gypi creating config file
gyp verb build/config.gypi writing out config file: /usr/src/app/node_modules/node-sass/build/config.gypi
gyp verb config.gypi checking for gypi file: /usr/src/app/node_modules/node-sass/config.gypi
gyp verb common.gypi checking for gypi file: /usr/src/app/node_modules/node-sass/common.gypi
gyp verb gyp gyp format was not specified; forcing "make"
gyp info spawn /usr/bin/python2
gyp info spawn args [ '/usr/src/app/node_modules/node-gyp/gyp/gyp_main.py',
gyp info spawn args   'binding.gyp',
gyp info spawn args   '-f',
gyp info spawn args   'make',
gyp info spawn args   '-I',
gyp info spawn args   '/usr/src/app/node_modules/node-sass/build/config.gypi',
gyp info spawn args   '-I',
gyp info spawn args   '/usr/src/app/node_modules/node-gyp/addon.gypi',
gyp info spawn args   '-I',
gyp info spawn args   '/root/.node-gyp/8.4.0/include/node/common.gypi',
gyp info spawn args   '-Dlibrary=shared_library',
gyp info spawn args   '-Dvisibility=default',
gyp info spawn args   '-Dnode_root_dir=/root/.node-gyp/8.4.0',
gyp info spawn args   '-Dnode_gyp_dir=/usr/src/app/node_modules/node-gyp',
gyp info spawn args   '-Dnode_lib_file=/root/.node-gyp/8.4.0/<(target_arch)/node.lib',
[...]

PR #406, opened 2018-03-01
Issue #405: Fix error when building with Docker

alexanderturinske:
- https://github.com/sass/node-sass/issues/1945 says
  that our current version of node-sass is not compatible
  with node v8, so I updated our version of node-sass


richardxia:
Thanks, @alexanderturinske! Sorry for holding off from merging this in, we were dealing with some prod issues so I didn't want to merge in anything to master until we got that resolved, which finally happened this morning.

PR #452, opened 2018-03-25
Added Google Analytics to the page, only in prod mode

lexholden:
It's crazy that we have no idea who is using AskDarcel, GA account created under my Sheltertech email and can share it with anyone who wants access.

screenshot 2018-03-25 00 18 14 4

Should probably add the same for Sheltertech.org


alexanderturinske:
LGTM!

PR #451, opened 2018-03-25
Issue #450: Add new sidebar to edit page

alexanderturinske:
- Creates sidebar component
 - Closes #450

PR #455, opened 2018-04-02
Fixed bug when resource doesn't exist (e.g. new resource page)

lexholden:
Fixes a bug missed in #451

### DIFF
--- a/app/components/Edit/EditSections.js
+++ b/app/components/Edit/EditSections.js
@@ -685,6 +685,7 @@ class EditSections extends React.Component {
         submitting={this.state.submitting}
         certifyHAP={this.certifyHAP}
         newServices={this.state.services.services}
+        newResource={this.state.newResource}
       />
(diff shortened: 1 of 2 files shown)


lexholden:
So @richardxia do you think we should roll back the original and remerge once both issues are fixed, or merge this and create an urgent followup issue for the additional bug you've found?

I'd suggest the latter personally, since we're about to refactor the whole page anyway.

alexanderturinske:
Next datathon is this Saturday; will we have time to fix it by then? I can work on this tonight.

richardxia:
If we think we can fix this with the refactoring in time for the datathon, then we don't need to rollback. Or additionally, we could just remove the "Add new resource" link entirely, if we don't think people will need to create new resources during the datathon.

alexanderturinske:
Pushed up a modified patch which fixes the issue fully.

PR #457, opened 2018-04-03
Added testcafe tests for listings page.

zjipsen:
to pass in travis, requires merging changes to populate.rake in askdarcel-api

### DIFF
--- a/testcafe/listings.js
+++ b/testcafe/listings.js
@@ -1,0 +1,35 @@
+import config from './config';
+import SearchPage from './pages/SearchPage';
+
+const searchPage = new SearchPage();
+
+fixture `Listings Page`
+  .page `${config.baseUrl}/resources?categoryid=234`;
+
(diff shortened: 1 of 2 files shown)


lexholden:
Rerunning the build, will merge if it passes.

lexholden:
@zjipsen it seems it still fails even with the above merged API change.

richardxia:
I am planning to look at this this week to see why it seems to be failing, since it looks like an unrelated test is failing.

richardxia:
I figured out the bug. @zjipsen accidentally removed the `SearchPage.resultEntry` selector, which the failing test relied on. I added it back in, but in order to share the `baseSelector` helper variable we created, I had to bump the testcafe and testcafe-react-selector version to support the new `findReact()` method that allows you to chain React selectors under other React selectors.

LMK if this looks good to y'all and we can merge it in.

PR #459, opened 2018-04-12
434: Add testcafe test for deleting phone number

zjipsen:


richardxia:
Yes, sorry about being really late to the review. Thanks for reviewing and merging, @lexholden!

PR #464, opened 2018-05-03
Issue 458 Create stateless tab view component for search

akatrent:
This is the tab component for search. The 'how to apply section' is static; I believe the data should be returned in application_process but every value in the object for application_process is null.


twolfe2:
LGTM

Issue #427, opened 2018-03-08
Confirm adding resource works


jjfreund:
New resource added should be immediately viewable, with all new fields correctly added including:

--Name
--Address and included fields
--Phone number (number and service type)
--Website
--Email
--Description
--Hours (including multiple open and close times per day)
--Notes

alexanderturinske:
This ticket can be completed now that submitting a resource now takes the user to the new resource page.

- [ ] Name
- [ ] Address and included fields
- [ ] Phone number (number and service type)
- [ ] Website
- [ ] Email
- [ ] Description
- [ ] Hours (including multiple open and close times per day)
- [ ] Notes

alexanderturinske:
Relates to https://github.com/ShelterTechSF/askdarcel-api/issues/449

PR #470, opened 2018-05-20
Add basic test for New Resource Page.

richardxia:
This sort of but not fully implements #427. It adds a basic test that visits the new resource page, fills out the minimum amount of information that won't trigger a server validation error, and hits the submit button.

What isn't easy for me to test is that the filled out information is actually present on the show page. This is because after you hit the submit button, it takes you to the home page, and it's not easy to find the show page for the resource you just created.

While I would prefer to ship what I have now and defer actual changes to the app to a separate PR, what you all think about changing the behavior of the submit button to redirect you to the show page of the resource that was just created? That would make testing a lot easier.

### DIFF
--- a/testcafe/editResource.js
+++ b/testcafe/editResource.js
@@ -1,19 +1,18 @@
-import config from './config';
 import ResourcePage from './pages/ResourcePage';
-import EditPage from './pages/EditPage';
+import EditResourcePage from './pages/EditResourcePage';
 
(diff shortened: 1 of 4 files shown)


lexholden:
(Since it works, makes more sense to add the enhancement to resource redirect when we get there, an imperfect feature is better than no feature at our current gait.

Issue #438, opened 2018-03-08
Confirm editing existing service works


jjfreund:
All fields should be editable including:

--Name
--Email
--Description
--Application Process
--Required Documents
--Eligibilities (from list)
--Fee
--Schedule (multiple open/close times per day supported)
--Note(s)
--Categories

PR #493, opened 2018-06-03
TestCafe: Add ServicePage and basic Service test

alexanderturinske:
- Relates to #438

### DIFF
--- a/app/components/listing/ActionSidebar.jsx
+++ b/app/components/listing/ActionSidebar.jsx
@@ -22,7 +22,7 @@ class ListPageSidebar extends React.Component {
               action.to || action.handler
                 ? <Link to={action.to} onClick={action.handler} className={`listing--aside--${action.name.toLowerCase()}`}>
                   { this.renderButtonContent(action) }
-                </Link >
+                </Link>
                 : <a href={action.link} target="_blank" className={`listing--aside--${action.name.toLowerCase()}`}>
                   { this.renderButtonContent(action) }
                 </a>
--- a/app/components/listing/ActionSidebar.jsx
+++ b/app/components/listing/ActionSidebar.jsx
@@ -20,10 +20,10 @@ class ListPageSidebar extends React.Component {
           <li key={action.name}>
             {
               action.to || action.handler
-                ? <Link to={action.to} onClick={action.handler} className={`listting--aside--${action.name.toLowerCase()}`}>
+                ? <Link to={action.to} onClick={action.handler} className={`listing--aside--${action.name.toLowerCase()}`}>
                   { this.renderButtonContent(action) }
-                </Link>
-                : <a href={action.link} target="_blank">
+                </Link >
+                : <a href={action.link} target="_blank" className={`listing--aside--${action.name.toLowerCase()}`}>
                   { this.renderButtonContent(action) }
                 </a>
             }
--- a/app/components/listing/ActionSidebar.jsx
+++ b/app/components/listing/ActionSidebar.jsx
@@ -10,7 +10,7 @@ class ListPageSidebar extends React.Component {
         <i className="material-icons">{ action.icon }</i>
         { action.name }
       </span>
-    )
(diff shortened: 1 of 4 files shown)


app/components/listing/ActionSidebar.jsx
```diff
@@ -20,7 +20,7 @@ class ListPageSidebar extends React.Component {
           <li key={action.name}>
             {
               action.to || action.handler
-                ? <Link to={action.to} onClick={action.handler}>
+                ? <Link to={action.to} onClick={action.handler} className={`listting--aside--${action.name.toLowerCase()}`}>
```
alexanderturinske: Two "t"s

testcafe/pages/ServicePage.js
```diff
@@ -0,0 +1,21 @@
...
+    const baseSelector = ReactSelector(baseSelectorName);
+    this.serviceName = baseSelector.find('.listing--main--left header');
+    this.description = baseSelector.find('.listing--main--left--about');
+    this.details = baseSelector.find('.listing--main--left--details');
+    // TODO: Can't use nested React component name PhoneNumber because it is
+    // instantiated in both the header and the body of the page and because the
+    // testcafe-react-selectors plugin is currently unable to mix CSS selectors
+    // in between React component names.
+    // https://github.com/DevExpress/testcafe-react-selectors/issues/51
+    this.phones = baseSelector.find('..listing--main--left--info');
```
alexanderturinske: Two "."

testcafe/pages/ServicePage.js
```diff
@@ -0,0 +1,21 @@
...
+    this.editButton = baseSelector.find('.listing--aside-edit');
```
alexanderturinske: Single "-"

testcafe/service.js
```diff
@@ -0,0 +1,22 @@
+import ServicePage from './pages/ServicePage';
+
+const servicePage = new ServicePage();
+
+fixture `Service Page`
+  .page(servicePage.url(1));
+
+test('Confirm Page Loads with Information', async (t) => {
+  const serviceName = 'Women\'s clinicA service offered by New Resource Name';
```
alexanderturinske: This works on local, but not the testing suite. Going to switch to check for existence.

testcafe/pages/ServicePage.js
```diff
@@ -0,0 +1,22 @@
...
+    const baseSelectorName = 'ServicePage';
+    const baseSelector = ReactSelector(baseSelectorName);
+    this.serviceName = baseSelector.find('.listing--main--left header');
+    this.description = baseSelector.find('.listing--main--left--about');
+    this.details = baseSelector.find('.listing--main--left--details');
+    // TODO: Can't use nested React component name PhoneNumber because it is
+    // instantiated in both the header and the body of the page and because the
+    // testcafe-react-selectors plugin is currently unable to mix CSS selectors
+    // in between React component names.
+    // https://github.com/DevExpress/testcafe-react-selectors/issues/51
```
richardxia: That ticket is out of date. You can now nest both CSS and React selectors under other ReactSelectors with the `.findReact()` method: https://github.com/DevExpress/testcafe-react-selectors#searching-for-nested-components

This also means that you don't need to add all the CSS classes that you added just for the sake of selecting components for tests.

alexanderturinske: The problem I have with that is then the tests are dependent on the layout of the HTML; by relying on classes, those elements can move anywhere on the page and the tests won't break. For example, the line below (editButton) would turn into `this.editButton = baseSelector.findReact('ListPageSidebar').findReact('ul').nth(0)` if I didn't have the class.

richardxia: Okay, yeah, that's a fair comment, though I feel like our CSS names are a bit too generic and there's a minor risk of accidentally selecting something else. But we can deal with that later.

PR #496, opened 2018-06-09
Migrated developer specific config to a yaml file in the root, with config.example.yml with instructions

lexholden:


richardxia:
Since it passed Travis, I'm going to merge this in so that I can submit a followup PR to help deal with the config issue.

PR #498, opened 2018-06-12
Adding react-markdown module, use for desc and notes

trucnguyen:
Added react-markdown module and used the component to render notes and resource description

### DIFF
--- a/app/components/App.jsx
+++ b/app/components/App.jsx
@@ -6,6 +6,7 @@ import { round } from '../utils/index';
 import 'react-select/dist/react-select.css';
 import { connect } from 'react-redux';
 import userActions from '../actions/userActions';
+import ReactMarkdown from 'react-markdown';
 
 
 class App extends Component {
--- a/app/components/Resource/Notes.jsx
+++ b/app/components/Resource/Notes.jsx
@@ -1,4 +1,5 @@
 import React from 'react';
+import ReactMarkdown from 'react-markdown';
 
 export default function Notes(props) {
     return (
@@ -17,7 +18,7 @@ function renderNotes(notes) {
     return notes.map((noteObj) => {
         return (
             <li className="service">
-                <p className="service--description">{noteObj.note}</p>
+                <div className="service--description"><ReactMarkdown source={noteObj.note} /></div>
             </li>    
         );
     });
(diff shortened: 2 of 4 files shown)


trucnguyen:
Seems to work.

screen shot 2018-06-11 at 10 05 43 pm

screen shot 2018-06-11 at 10 05 15 pm

Also formats tables as expected in the examples given in the react-markdown component readme doc.
screen shot 2018-06-11 at 10 05 25 pm

I've added this Resource description and the Notes component. 

However, it looks like I missed some of the Service fields. Text editable fields in Services components have its own renderer function. Shall I add to service description and ServiceEligibility?

richardxia:
> Also, it feels weird that a user can add an h1 that is bigger than the page title, but I'm not sure of a solution beyond perhaps some config passed down to remarkable (any ideas, @trucnguyen ?)

We could change the CSS styles such that if you're in a description field that was generated from Markdown, we make the headings smaller.

lexholden:
True. I guess we'd still do weird things for SEO working out what was important on the page, but we can solve that issue down the line if it becomes a problem.

richardxia:
Under HTML5, it is legal to restart the h1-h6 numbering within nested `<section>`s. See https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Using_HTML_sections_and_outlines#The_HTML5_outline_algorithm

I believe search engines have been dealing with this for a while now, so it should not affect SEO. I think a possibly greater negative effect on SEO may be the fact that we need to do React routing and rendering HTML from Markdown in JS in order to produce any of this, so it depends on how sophisticated of a JavaScript execution engine the web crawler uses.

lexholden:
Good point, I guess that's a future enhancement.

Then I say we merge this for this milestone and create a future task to improve sizing of h1-6 in description fields and an explanation that we support markdown in the edit flow?

richardxia:
Yep, agreed with merging this now, since the UI still looks acceptable.

trucnguyen:
Added those changes for note on edit flow that we support MD and overriding the styles so it's never as big and bold as the section title.

app/components/Resource/Services.js
```diff
@@ -1,6 +1,7 @@
 import React, { Component } from 'react';
 import classNames from 'classnames';
 import DetailedHours from './DetailedHours';
+import ReactMarkdown from 'react-markdown';
```
richardxia: Did you mean to use this on this page? I don't see it being used.

app/components/App.jsx
```diff
@@ -6,6 +6,7 @@ import { round } from '../utils/index';
 import 'react-select/dist/react-select.css';
 import { connect } from 'react-redux';
 import userActions from '../actions/userActions';
+import ReactMarkdown from 'react-markdown';
```
richardxia: Same here.

PR #504, opened 2018-06-15
Redirect New Section Creation

### LINKED ISSUE #485: When you create a new resource, you should be redirected to the page itself
Blocking #470

brabo-hi:
After creating new resource, the page redirects to it. Closes #485

### DIFF
--- a/app/components/Edit/EditSections.js
+++ b/app/components/Edit/EditSections.js
@@ -288,7 +288,7 @@ class EditSections extends React.Component {
       .then((response) => {
         if (response.ok) {
           alert('Resource successfuly created. Thanks!');
-          response.json().then(res => browserHistory.push(`/resources?id=${res.resources[0].resource.id}`));
+          response.json().then(res => browserHistory.push(`/resource?id=${res.resources[0].resource.id}`));
         } else {
           Promise.reject(response);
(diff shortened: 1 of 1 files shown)


app/components/Edit/EditSections.js
```diff
@@ -288,7 +288,7 @@ class EditSections extends React.Component {
       .then((response) => {
         if (response.ok) {
           alert('Resource successfuly created. Thanks!');
-          browserHistory.push('/');
+          response.json().then(res => browserHistory.push(`/resources?id=${res.resources[0].resource.id}`));
```
richardxia: I checked out this branch locally, and I noticed that after submitting a new resource, I get redirected to a page that never loads completely. I think you actually want to redirect to `/resource`, not `/resources`. The former is the individual resource page, while the latter is the listing of multiple resources, like the search results page.

lexholden: (I plan to change this URL convention in my current resource page rewrite because it's confusing, will add this change now myself so we can deploy this ASAP.

richardxia: Agreed that the current URLs are confusing because they don't fit the usual RESTful conventions.

Issue #428, opened 2018-03-08
Confirm modifying resource fields works


jjfreund:
Check that all resource fields can be modified, including:

--Name
--Address and related fields
--Phone(s) (number and service type)
--Website
--Email

(Schedule day and note modifications are handled by separate tests)

alexanderturinske:
As of this message, the following are done:
- [x] [Name](https://github.com/ShelterTechSF/askdarcel-web/blob/73c37163e2a04cb76e512b18a9fb4fe6264c5a1b/testcafe/editResource.js#L20)
- [x] [Address and related fields](https://github.com/ShelterTechSF/askdarcel-web/blob/73c37163e2a04cb76e512b18a9fb4fe6264c5a1b/testcafe/editResource.js#L25)
- [x] [Resource phone number](https://github.com/ShelterTechSF/askdarcel-web/blob/73c37163e2a04cb76e512b18a9fb4fe6264c5a1b/testcafe/editResource.js#L65)
- [ ] Resource phone type
- [x] [Website](https://github.com/ShelterTechSF/askdarcel-web/blob/73c37163e2a04cb76e512b18a9fb4fe6264c5a1b/testcafe/editResource.js#L124)
- [x] [Email](https://github.com/ShelterTechSF/askdarcel-web/blob/73c37163e2a04cb76e512b18a9fb4fe6264c5a1b/testcafe/editResource.js#L133)

PR #506, opened 2018-06-18
Test editing and viewing resource fields

### LINKED ISSUE #419: Confirm all resource fields are displayed properly on resource page
The following fields should be displayed:

--Open status (either as "Open until x" or "Closed")
--Each address associated with the resource
--Each phone number associated with the resource
--The description of the resource

richardxia:
Fixes #419 and #428, except I'm going to claim that some of the things mentioned in those tickets are better done with other TestCafe tickets.

## Justification for rescoping these tickets
From #419:

> --Open status (either as "Open until x" or "Closed")

Should be done with one of the tickets related to schedule days.

> --Each address associated with the resource
> --Each phone number associated with the resource

Already done previously.

From #428:

> --Address and related fields

Already done previously.

> --Phone(s) (number and service type)

This should be done with some of the other tickets related to adding/removing phone numbers.

## Actual changes made
The email on a resource wasn't easy to select because the DOM element it is in has the class "website"  for some reason. I did a bit of refactoring so that there's an Email component, similar to the Website component, and I pulled out the common bits between that and the Website component into an ExternalLink component.

Aside from that, I added tests for checking that the website, email, and description fields can be modified and that the modifications show up on the show page. Note that the Alternative Name does not seem to actually appear on the show page, so I left that out.

### DIFF
--- a/app/components/Resource/Resource.jsx
+++ b/app/components/Resource/Resource.jsx
@@ -1,7 +1,7 @@
 import React, { Component } from 'react';
 import { Link } from 'react-router';
 import PropTypes from 'prop-types';
-import { AddressInfo, TodaysHours, PhoneNumber, ResourceCategories, Website, StreetView } from './ResourceInfos';
+import { AddressInfo, TodaysHours, PhoneNumber, ResourceCategories, Website, Email, StreetView } from './ResourceInfos';
 import DetailedHours from './DetailedHours';
 import Services from './Services';
 import Notes from './Notes';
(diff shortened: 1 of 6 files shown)


app/components/Resource/Resource.jsx
```diff
@@ -1,7 +1,7 @@
 import React, { Component } from 'react';
 import { Link } from 'react-router';
 import PropTypes from 'prop-types';
-import { AddressInfo, TodaysHours, PhoneNumber, ResourceCategories, Website, StreetView } from './ResourceInfos';
+import { AddressInfo, TodaysHours, PhoneNumber, ResourceCategories, Website, Email, StreetView } from './ResourceInfos';
```
lexholden: Given our existing component soup problem (difficult to find anything) not a fan of big multifile components, like here in ResourceInfos personally, but we can address this down the line.

richardxia: Agreed. Also if we could create more reusable components, that would help, instead of having all components being organized by "page" or "React route."

PR #515, opened 2018-07-01
Category filtering and MOHCD badge on search results

twolfe2:


twolfe2:
@lexholden I think that's do to your local DB not having the addresses, I've updated this branch so `No address found` will be displayed instead of `undefined`

PR #519, opened 2018-07-15
Added Markdown to the services page

lexholden:


jjfreund:
The search results page also needs to support markdowns.

PR #537, opened 2018-08-05
Bugfixes: Search opening times, service emails, duplicate also at listings

### LINKED ISSUE #499: Open / closed status is no longer displaying correctly on search results page
Some resources/services display as closed though they are open, and some display as open though they are closed.

### LINKED ISSUE #529: On the service page, the service currently displayed is shown in the list of "other services at this location"

### LINKED ISSUE #533: Service emails not visible on the organization page

lexholden:
Closes #529 
Closes #533 
Closes #499


app/components/Resource/Services.js
```diff
@@ -65,8 +67,9 @@ class Service extends Component {
       { infoHidden ? null :
         <div className="service-application-process-container">
           <ul className="service--details">
+            <ServiceContactDetails email={service.email}/>
             <ServiceEligibility subject='How to apply' result={this.props.service.application_process}/>
-            <ServiceEligibilities subject='Eligibilities' eligibilities={this.props.service.eligibilities}/>
+            <ServiceEligibility subject='Eligibilities' result={this.props.service.eligibility}/>
```
richardxia: If `ServiceEligibilities` isn't being used anymore, then perhaps we should delete class?

app/components/Search/ResourceEntry.jsx
```diff
@@ -52,7 +55,11 @@ class ServiceEntry extends Component {
           <div className="entry-details">
             <h4 className="entry-headline">{hit.name}</h4>
             <div className="entry-subhead">
-              <p>{`${hit.address ? hit.address.address_1 : 'No address found'} • ${timeInfo}`}</p>
+              <p>
+                { hit.address && hit.address.address_1 ? hit.address.address_1 : 'No address found' }
+                { hit.schedule ? ' • ' : null }
```
richardxia: Should this be `schedule` instead of `hit.schedule`? This would make this code more parallel to the conditional one line below.

app/components/Resource/Services.js
lexholden: It's odd, we have an eligibilities table (tag style field linking to a service), and a column in services for eligibility (which is just a text field, and a tag style field)

We should really deprecate the text field and rely entirely on tags I'd think, but both fields have data on prod.

I'd keep the component for now, since the ServiceEligibilities component renders the list style of tags

app/components/Search/ResourceEntry.jsx
lexholden: Yeah you're right, I thought this after I'd committed actually but didn't bother to go back and change. Patching now.

app/components/Resource/Services.js
richardxia: Got it, I agree with keeping the component, then. When we made the large DB schema changes earlier this year, we didn't fully migrate over to the new fields, so it'd be a good idea to make that a longer-term objective of the askdarcel-api project. @jjfreund

PR #540, opened 2018-08-12
Mohcd badge on service pages

twolfe2:
Added the MOHCD badge to the service pages


app/components/Search/ResourceEntry.jsx
```diff
@@ -48,7 +48,7 @@ class ServiceEntry extends Component {
     } else {
       timeInfo = 'Closed';
     }
-
+    // debugger;
```
lexholden: We should remove commented out debug code before merging.

Issue #543, opened 2018-08-26
RFC: Fix production deployment process


richardxia:
We currently have a pretty big issue regarding our production deployment process. This needs to be fixed before the next production deployment, or else we will have to continue to fix things by hand.

## tl;dr
Staging credentials (e.g. Algolia) are getting built into the production containers because credentials are injected at webpack compile-time, not container runtime.

## Summary of the problem

Our software development lifecycle (SDLC) looks like the following:

1. New features and bugfixes are developed on feature branches and eventually merged into the `master` branch.
2. An update to the `master` branch triggers an automated build on Travis CI, which will build, tag, and upload a Docker image and trigger a deployment to the staging environment (staging.askdarcel.org).
3. When we decide that we want to release to production, we do QA on the staging environment, and when satisfied, we create a git tag/GitHub release, which triggers another build on Travis CI, creating a new Docker image and tags the image with the version number used in the git tag.
4. We update the AWS CloudFormation stack with the new Docker image tag, which will then deploy the new image to the production environment.

The problem occurs in step 3), where the production image is created with the staging environment's settings for services such as Algolia. Under our current process, without manual intervention, if we deploy images built in this process to production, they end up connecting to the staging environment's Algolia indexes.

A misleading part of our CloudFormation stack is the fact that we specify environment variables for configuring the Algolia index. This in fact does nothing because the environment variables are only used at the time that the React app is compiled, which is at Docker image build time. When the image is deployed, it simply ignores these environment variables.

## Possible solutions

I think there are essentially two different categories of solving this problem. Either we:

1. Allow the environment variables to actually affect the containers at runtime, or
2. Build a separate Docker image for every set of environment variables we'd like to set.

I'm in favor of 1), since I think it's a more flexible solution. There are several benefits to 1):

- We can reuse the same Docker image across all environments. We'd actually be QAing the same code in staging that we deploy to production.
- We can change the configuration values without having to build another image. For example, if we needed to change the Algolia index name, we could do this without having to trigger a new Travis CI build.
- We can more easily support new types of environments, such as if we ever wanted to do more whitelabeling deployments or spin up regional versions of Ask Darcel. Having to define more variants of the build pipelines seems less scalable than just injecting different settings into a single Docker image.

[...]

PR #544, opened 2018-08-26
Upgrade webpack and webpack loader dependencies

richardxia:
I want to play around with webpack code splitting in order to work on #543, but we're using some pretty old versions of webpack and webpack loaders. Here I've bumped the versions of various dependencies to the latest and made sure that the app still runs fine locally. I'd like to decouple this work from work on #543 since it's a good idea to keep our dependencies up-to-date anyway, and merging this change earlier will avoid issues with other people developing on older versions of the dependencies.


richardxia:
Pushed one more commit to upgrade NodeJS due to a dependency requiring it and removing yarn.lock, because it was causing Travis to run `yarn` instead of `npm`. It's important for everyone to use a single package manager because otherwise we'll never be able to keep the lock files up to date. Since we were previously using npm and all of our README instructions say to use npm, I've decided that we're going to continue using npm.

PR #546, opened 2018-09-09
Change so new notes and services go to end of their lists and added new service button to the sidebar on the edit page

twolfe2:
Changes:
- New notes and services go to end of list now
- Button for new notes and services moved to the end of their respective lists
- Add service button added to the edit sidebar to make it easier to create a new service


app/components/edit/EditServices.jsx
```diff
@@ -41,14 +41,18 @@ class EditServices extends Component {
     });
   }
 
+  getAlert() {
+    alert('clicked');
+  }
+
```
richardxia: I'm going to remove this debug code.

Issue #340, opened 2017-12-08
Sub-heading links jump to below the Service name


lgarofalo:
After clicking some sub-heading links, the navigation sends me to the first paragraph, and the service name is hidden, and I must scroll up to see it.

PR #547, opened 2018-09-10
fixes: #340

genovese28:
It looks like the scrolling was messed up because of the Notes section which has the same id as Services. I commented it out, and the scrolling behaves as it should now. Notes looks like it was an older version or duplicate work of Services. If it's something else, it should be refactored.

I also fixed up Services so that the list items are direct children of the unordered list instead of being nested inside divs. I also consolidated renderService and renderServicesSection, because it looked confusing and renderServicesSection was just calling renderService anyways.


richardxia:
> It looks like the scrolling was messed up because of the Notes section which has the same id as Services. I commented it out, and the scrolling behaves as it should now. Notes looks like it was an older version or duplicate work of Services. If it's something else, it should be refactored.

We actually have notes on both Resources and Services, so it's important that we still display the Resource notes. It *is*, however, a problem that the `<Notes>` component on both resources and services have the same ID. Let me clean this up a bit by having the `<Notes>` component take in an ID and passing in a different ID for each call to it.

PR #566, opened 2018-09-30
Refactor of file structure to remove loads of junk code and old components, plus name things more consistently

lexholden:


app/components/search/ResourcesRow.jsx
```diff
@@ -133,6 +133,8 @@ class ResourcesRow extends Component {
     let schedule = this.props.resource.schedule ? this.props.resource.schedule.schedule_days : [];
     let { open, time } = this.getOpenTime(schedule);
 
+    console.log(this.props)
```
richardxia: Remove this `console.log`.

PR #568, opened 2018-10-04
Fix google maps attribution

lgarofalo:


lgarofalo:
This looks like it fixes, once it passes again ill merge

PR #574, opened 2018-10-21
573 Restore TestCafe tests

richardxia:
This PR is basically two main changes:

1. Adding a new Google API for use in development/staging which can access the Google Maps APIs
2. Fixing the TestCafe tests that have been broken while we weren't running them, since the structure of the pages has changed.

Note: I did make one major change to our Travis CI settings (from the web UI, not .travis.yml), which was to remove the `ALGOLIA_INDEX_PREFIX`, which was previously set to `production`. I needed to make this change in order to ensure that the web and api instances that run in Travis are actually using the same Algolia index, which is now required because we got rid of some of the legacy endpoints that allowed us to bypass Algolia search.

This shouldn't be necessary anymore because the web Docker images can now read that value from the environment instead of baking it in, but we should be careful next time we deploy to make sure that we're reading from the correct Algolia index.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,8 @@ addons:
     - ubuntu-toolchain-r-test
     packages:
     - g++-4.8
+env:
+  - ALGOLIA_INDEX_PREFIX=travisci
 before_install:
 - cp config.example.yml config.yml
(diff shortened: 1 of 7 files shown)


alexanderturinske:
lgtm 👍

PR #576, opened 2018-10-29
Fix organizations/resources without addresses, closes #476

### LINKED ISSUE #476: Fix issue with not finding lat/long that is causing resource pages not to load
For whatever reason, some pages have not correctly created lat/long when the resource was added. HAP has gathered a list of these failures: https://quip.com/OSRHAOCOayjl

AnnaKalkanis:
In addition to the organization page, I've also updated the service listing page to protect against missing addresses. I just added some conditionals and pulled some logic out of the jsx and into methods.


lgarofalo:
Going to merge this rather than let it become stale.

PR #587, opened 2018-11-12
Closed/Open hours now behave as they should

lexholden:
One bug still to fix, but at least can get a preliminary review on this.


app/components/listing/RelativeOpeningTime.jsx
```diff
@@ -2,6 +2,13 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import { buildHoursText } from '../../utils';
 
+const parseAsDate = (fourDigitNumber) => {
```
lexholden: Need to adapt this to work with 3 digit numbers for 00-959 but this fix will give us a bunch of different strings including Closed, Closed today, Open, Open 24/7, Open 24 hours today and opens in N minutes and closes in N minutes if N < 30

richardxia: I think this was already implemented in [utils/index.js](https://github.com/ShelterTechSF/askdarcel-web/blob/eb14b43f8ab78def845b417cde714fddcc2822a3/app/utils/index.js#L21). If that implementation is incomplete, then we should replace that one with this, but I think that one handles the 3-character times.

app/components/listing/RelativeOpeningTime.jsx
```diff
@@ -17,18 +24,74 @@ class RelativeOpeningTime extends React.Component {
...
+    const STATUS_OPEN = 'status-green';
+    const STATUS_CAUTION = 'status-amber';
+
+    const currentDate = new Date();
+    const currentDay = currentDate.getDay();
+    const currentDayString = RelativeOpeningTime.daysOfWeek[currentDay];
+
+    const todayHours = schedule_days.filter(hours => hours.day === currentDayString);
+    const tomorrowHours = schedule_days.filter(hours => hours.day === RelativeOpeningTime.daysOfWeek[currentDay + 1 > 6 ? 0 : currentDay + 1]);
+    const daysOpen24Hours = schedule_days.filter(hours => hours.opens_at === 0 && hours.closes_at === 2359)
```
richardxia: A refactoring suggestion: We could pull out a function `is24Hours()` that takes in an `hours` object and returns true iff it's open 24 hours. This same check appears several times in this function.

app/pages/debug/ListingDemoPage.jsx
```diff
@@ -0,0 +1,50 @@
...
+  [[[915, 1200], [1300, 1800]], [900, 1800], [900, 1800], [900, 1800], [900, 1800], [900, 1200], [900, 1200]],
+  [[0, 2359], [900, 1800], [900, 1800], [900, 1800], [900, 1800], [900, 1200], [0, 2359]],
+  [[0, 2359], [900, 1800], [900, 1800], [900, 1800], [900, 1800], [900, 1200], null], // Closed all sunday, 24h Mon
+  [[0, 2359], [0, 2359], [0, 2359], [0, 2359], [0, 2359], [0, 2359], [0, 2359]], // 24/7
+  [[0, 2359], [0, 2359], [0, 2359], [0, 2359], [0, 2359], [0, 2359], [0, 2125]],
+  [[0, 2359], [0, 2359], [0, 2359], [0, 2359], [0, 2359], [0, 2359], [2155, 2300]],
+  [[900, 1800], [0, 2359], [0, 2359], [0, 2359], [0, 2359], [0, 2359], null],
+  [[900, 1800], null, [0, 2359], [0, 2359], [0, 2359], [0, 2359], [900, 2200]],
+  [[0, 2359], [0, 2359], [0, 2359], [0, 2359], [0, 2359], [0, 2359], [[0, 900], [1100, 1800], [1900, 2300]]],
+];
```
richardxia: 👍 

This would also be great to add to a unit test, eventually.

app/components/listing/RelativeOpeningTime.jsx
lexholden: Yep I decided not to use the other one because this does more and is a bit more maintainable. Even with the utils function, different places had different implementations.

app/components/listing/RelativeOpeningTime.jsx
lexholden: Huh? This is the only time this is checked, after that we  check if there are 7 days or if today or tomorrow is 24h

PR #595, opened 2018-11-26
Improved RelativeTimeStamp to support overnight dates and timezones

lexholden:


app/components/listing/RelativeOpeningTime.jsx
```diff
@@ -1,62 +1,94 @@
...
+    const opens_at = currentDate
+      .clone()
+      .add(daysFromToday, 'days')
+      .startOf('day')
+      .add(openingTime.getHours(), 'hours')
+      .add(openingTime.getMinutes(), 'minutes');
+
+    const closes_at = currentDate
+      .clone()
+      .add(daysFromToday, 'days') // If the close is before the open, add a day
```
richardxia: I understand why we need to do the logic described by the comment, but I don't actually see where we are adding the extra day. To me, it looks like `opens_at` and `closes_at` will both have the same day, even if the `closes_at` hour is before the `opens_at` hour.

app/components/listing/__tests__/RelativeOpeningTimes.spec.jsx
```diff
@@ -33,32 +34,34 @@ describe('<RelativeOpeningTimes />', () => {
...
+    const date = moment().startOf('day').add(17, 'hours').add(50, 'minutes');
     const schedule = getNamedSchedule('nine_to_six');
     const comp = shallow(<RelativeOpeningTime schedule={schedule} currentDate={date} />);
     expect(comp.find('.relative-opening-time').text()).to.equal('Closes in 10 mins');
     expect(comp.find('.relative-opening-time').hasClass('status-amber')).to.be.true;
   });
+
+  it('should display "Closed until tomorrow" for a schedule that was open today, but has closed until tomorrow', () => {
+    const date = moment().startOf('day').add(18, 'hours').add(30, 'minutes');
+    const schedule = getNamedSchedule('nine_to_six');
```
richardxia: Is this the right schedule? It looks like this is a normal 9am-6pm schedule.

app/components/listing/RelativeOpeningTime.jsx
lexholden: Hmm you're right. I think I accidentally removed this in cleaning up my own code... Adding it back.

app/components/listing/__tests__/RelativeOpeningTimes.spec.jsx
lexholden: Yes, because the test feeds in a fake current time as 6:30pm, so it was open today, closed 30m ago and isn't open until 9am tomorrow

richardxia: Ah, I realized I completely misread the description of the test. I mistakenly thought that this was going to be about the case where you have a close time that is before the open time, indicating that the close time should actually be one day in the future.

richardxia:
I'm going to try to fix up this branch and get it fully working, since we need this for fixing the Open/Close times on prod.

app/components/listing/RelativeOpeningTime.jsx
```diff
@@ -2,8 +2,8 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import moment from 'moment';
 
-// TODO Until we use actual timestamps, this one constant converts everything to PST below
-const timezoneOffset = 8;
+// TODO Until we use actual timestamps, this one constant converts everything to the correct timezone
+const timezoneOffset = new Date().getTimezoneOffset() / 60;
```
richardxia: @lexholden I think I fixed the Travis issues with this one change. The problem with Travis is that their servers set the timezone to UTC, so the hardcoded constant wasn't working correctly on their servers, even if they work on a browser located in California. This is probably a slightly better approach anyway, since San Francisco shifts to -7 hours during Daylight Savings.

PR #596, opened 2018-11-28
Added css for markdown comments so basic properties render correctly

lexholden:


app/components/listing/Services.jsx
```diff
@@ -48,7 +48,7 @@ class Service extends Component {
           <p>updated {service.updated_date}</p>
         </div>
         <h2 className="service--header">{service.name}</h2>
-        <ReactMarkdown className="service--description" source={service.long_description} />
+        <ReactMarkdown className="rendered-markdown" className="service--description" source={service.long_description} />
```
richardxia: Is this legal JSX? I've never tried passing in the same attribute twice before. Do you know if the component just concatenates the `classNames`, or does one replace the other?

lexholden: Good point, I literally just find-replaced all markdown tags, didn't occur to me that any already had classes. Will fix today.

PR #609, opened 2018-12-30
ISSUE #606: Refactor footer to include more information

### LINKED ISSUE #606: Replace landing page footer
Should conform to design referenced by the epic.

alexanderturinske:
Closes #606


app/styles/components/_footer.scss
```diff
@@ -13,46 +13,61 @@
     color: $color-white;
   }
 
+  .footer-service-guide {
+    text-align: center;
+    font-weight: 1000;
```
richardxia: Might be safer to use 900. I didn't actually realize that 1000 is a supported value now, but older browsers might only support 100-900. This chart from MDN doesn't even know if Edge or Safari support 1000: https://developer.mozilla.org/en-US/docs/Web/CSS/font-weight#Browser_compatibility

app/components/ui/Footer.jsx
```diff
@@ -5,14 +5,40 @@ function Footer() {
   return (
     <footer className="site-footer" role="contentinfo">
       <div className="footer-content">
+        <div className="footer-top">
+          <section className="footer-service-guide">
+            SF SERVICE GUIDE
+          </section>
+          <section className="footer-links">
+            <ul>
+              <li>ABOUT US</li>
```
richardxia: Nit: My personal approach for this kind of stuff is to type out the text in the source code using normal casing and adding a CSS rule to set `text-transform: uppercase`. Also, I would probably make this first thing a heading tag like `<h1>/<h2>/...`, since it's more of a heading than a list item like the rest of the things.

app/styles/components/_footer.scss
alexanderturinske: Done

alexanderturinske:
- Update class names to be BEM style
- Added footer icon
- updated for comments

Desktop
screen shot 2019-01-06 at 12 36 30 am
Mobile
screen shot 2019-01-06 at 12 36 42 am

app/components/ui/Footer.jsx
```diff
@@ -5,14 +5,40 @@ function Footer() {
...
+              <li><a href="info@sheltertech.org">Email</a></li>
+              <li><a href="https://www.facebook.com/ShelterTechOrg/">Facebook</a></li>
+              <li><a href="https://twitter.com/sheltertechorg">Twitter</a></li>
+              <li><a href="https://www.instagram.com/shelter_tech/">Instagram</a></li>
+            </ul>
+            <ul>
+              <li>LEGAL</li>
+              <li><Link to="/terms-of-service">Terms of Service</Link></li>
+              <li><Link to="/privacy-policy">Privacy Policy</Link></li>
+              <li><a href="#">API Policy</a></li>
```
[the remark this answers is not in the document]
lexholden: I don't think it exists yet, I added the content for the two above a while back, but those pages didn't exist either at the time. I guess you need to add a new page in `app/components/pages/legal`

app/components/ui/Footer.jsx
```diff
@@ -5,14 +5,40 @@ function Footer() {
...
+            <ul>
+              <li>ABOUT US</li>
+              <li><a href="sheltertech.org">Our Story</a></li>
+              <li><a href="sheltertech.org/volunteer">Volunteer</a></li>
+              <li><a href="sheltertech.org/donate">Donate</a></li>
+            </ul>
+            <ul>
+              <li>HELP</li>
+              <li><a href="#">FAQ</a></li>
+              <li><a href="#">Digital Literacy</a></li>
```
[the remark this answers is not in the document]
lexholden: Same for these, though @derekfidler I guess the specs for this links is a direct request from the city? We need content for them otherwise we should comment the links out until we can add the pages.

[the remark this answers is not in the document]
alexanderturinske: @richardxia @lexholden I will comment out these links until we know where they go. That way the commit can be unblocked.

alexanderturinske:
Commented out links with no well defined redirects
Desktop
image
Mobile
image

app/components/ui/Footer.jsx
```diff
@@ -4,15 +4,44 @@ import { Link } from 'react-router';
...
+            <h1 className="service-guide__text">
+              SF Service Guide
+            </h1>
+          </section>
+          <section className="site-footer__links">
+            <ul>
+              <h1>About Us</h1>
+              <li><a href="sheltertech.org">Our Story</a></li>
+              <li><a href="sheltertech.org/volunteer">Volunteer</a></li>
+              <li><a href="sheltertech.org/donate">Donate</a></li>
```
richardxia: `/donate` doesn't appear to be a real URL.

app/utils/config.example.js
```diff
@@ -0,0 +1,7 @@
+
```
richardxia: I don't think you meant to commit this file.

app/components/ui/Footer.jsx
```diff
@@ -4,15 +4,44 @@ import { Link } from 'react-router';
...
     <footer className="site-footer" role="contentinfo">
-      <div className="footer-content">
-        <div className="footer-text">
-            ©{ new Date().getFullYear() } Shelter Tech
+      <div className="site-footer__content">
+        <div className="site-footer__top">
+          <section className="service-guide">
+            <div className="service-guide__icon" />
+            <h1 className="service-guide__text">
+              SF Service Guide
```
richardxia: I think we want this to conditionally say "Ask Darcel" or "SF Service Guide" depending on what domain name the site is being served from. Here's an example of where we're currently doing a check based on the subdomain: https://github.com/ShelterTechSF/askdarcel-web/blob/861898cbc87b5e6459791b0ecbb3b6cf1cbea791/app/pages/HomePage.jsx#L43

app/components/ui/Footer.jsx
alexanderturinske: Ah, I will comment it out.

app/utils/config.example.js
alexanderturinske: Curses; that is the second time that has gotten pushed up. WIll remove.

app/components/ui/Footer.jsx
alexanderturinske: Will do

app/components/ui/Footer.jsx
alexanderturinske: Done

app/styles/components/_footer.scss
```diff
@@ -13,49 +13,77 @@
...
   }
 }
+
+.site-footer__disclosure {
+  text-align: center;
+  @include r_max($break-tablet-s) {
+    display: block;
+    padding-left: 0;
+    text-align: center;
+    padding-left: 0;
```
alexanderturinske: There are two `padding-left`; why?

alexanderturinske: Deleted

Issue #607, opened 2018-12-13
Replace landing page hero


jjfreund:
Should conform to the design referenced by the epic. The total number of resources should be retrieved via an API call.

PR #610, opened 2018-12-30
#607 - Update landing page hero

AnnaKalkanis:
Still need to update `Search 1,768 resources in San Francisco` so it dynamically pulls in the number of resources. But this hopefully covers all the other changes.

screen shot 2018-12-29 at 9 46 20 pm

screen shot 2018-12-29 at 9 35 25 pm


app/components/layout/FindHeader.jsx
```diff
@@ -33,7 +42,7 @@ class FindHeader extends React.Component {
             ref={(c) => { this.searchComponent = c; }}
             type="text"
             className="search-field"
-            placeholder='try "rental assistance" or "Compass Family Shelter"'
+            placeholder="Search 1,768 resources in San Francisco"
```
alexanderturinske: NIT: It feels bad to hardcode this number. Would "over 1,700" work or something more general, @derekfidler?

app/components/layout/FindHeader.scss
```diff
@@ -15,25 +15,67 @@
...
+      .location-pin {
+        background-repeat: no-repeat;
+        background-size: contain;
+        background-position: center;
+        width: calc-em(42px);
+        height: calc-em(56px);
+        position: absolute;
+        &.pin-1 {
+          background-image: url(../../assets/img/location-pin-1.png);
+          top: calc-em(35px);
```
alexanderturinske: What are these `calc-em`s for? I have never seen them and do not know their benefit over just using `em` or what not. Especially since you are calculating them off of a fixed `px` value, wouldn't it just be better to set them to `35px`?

app/components/layout/FindHeader.jsx
```diff
@@ -22,8 +22,17 @@ class FindHeader extends React.Component {
   render() {
     return (
       <header className="hero header-large" role="banner">
-        <h1>Hello, what can we help you find?</h1>
-        <h3>Search 1,138 housing and homelessness related services in San Francisco</h3>
+        <div className="header-container">
+          <div className="header-text">
+            <p>Welcome to the SF Service Guide</p>
+            <h1>Find food, housing, health<br />resources and more<br />in San Francisco</h1>
```
richardxia: These explicit `<br>`s might cause weird breaking on certain screen sizes, such as in the event where a screen size is small enough to wrap right before the word "health". Can we achieve the desired appearance by instead setting a maximum width on the `<h1>` or possibly the `<div className="header-text">`?

app/components/layout/FindHeader.scss
richardxia: I don't know exactly why we have `calc-em()`, but I think it's probably not necessary anymore given that [`rem`](https://developer.mozilla.org/en-US/docs/Web/CSS/length) is supported by every browser including IE 9 and up. `em` is better than `px` in cases where you want measurements to be proportional to the font size (for accessibility reasons, some users may configure their browser to use a larger base font size), but `rem` is better than `em` in most cases where you want a measurement to be proportional to the root font size and not the font size of a specific element. It looks like [`calc-em()`](https://github.com/ShelterTechSF/askdarcel-web/blob/71c648d9c2be897b38ee3246d7c4fba0659b2e98/app/styles/utils/_mixins.scss#L4-L6) does some weird stuff to reverse compute what the `em` size should be based on the font-size of the current element, but it normalizes based on `rem` anyway.

app/components/layout/FindHeader.jsx
```diff
@@ -22,8 +22,17 @@ class FindHeader extends React.Component {
   render() {
     return (
       <header className="hero header-large" role="banner">
-        <h1>Hello, what can we help you find?</h1>
-        <h3>Search 1,138 housing and homelessness related services in San Francisco</h3>
+        <div className="header-container">
+          <div className="header-text">
+            <p>Welcome to the SF Service Guide</p>
```
richardxia: I have the same question on https://github.com/ShelterTechSF/askdarcel-web/pull/609#pullrequestreview-188890498, but I think we may want to make "SF Service Guide" parameterizable based on the URL, since I *think* we only whitelabel the app for the city when it's at a certain domain name, and we otherwise still want the app to be called Ask Darcel.

app/components/layout/FindHeader.scss
```diff
@@ -15,25 +15,67 @@
...
-  h1 {
-    @extend %font-size-hero;
-    @extend %base-max-width;
-    font-weight: 600;
-    color: $color-grey8;
-    text-align: center;
-    margin-bottom: calc-em(12px);
+  .header-container {
+    @extend %search-max-width;
+    display: flex;
```
richardxia: Is this needed? I don't see you setting any other flexbox properties.

app/components/layout/FindHeader.scss
```diff
@@ -15,25 +15,67 @@
...
+      p {
+        color: $color-grey5;
+      }
+
+      h1 {
+        @extend %font-size-hero;
+        font-weight: 600;
+        color: $color-grey8;
+        @include r_max($break-tablet-s) {
+          width: 100%;
```
richardxia: Is this necessary to set? Block elements by default span the width of their parent.

app/components/layout/FindHeader.scss
```diff
@@ -15,25 +15,67 @@
...
+          margin-top: calc-em(13px);
+          max-width: 20rem;
```
richardxia: Is it necessary to set this max-width?

app/components/layout/FindHeader.jsx
alexanderturinske: Oh, you left that in the description above. 🤦‍♂️ I apologize for not reading.

PR #614, opened 2019-01-07
WIP: Fix lint errors

richardxia:
I noticed that our JavaScript linting wasn't actually working properly. It turns out that we were only linting files that end in `.js` and not `.jsx`, meaning that almost none of our JavaScript was getting linted.

I updated the `npm run lint` script to check both extensions, but now we're failing a lot of lint checks. This PR includes changes to files that could be fixed automatically, but there are still 158 lint errors.


richardxia:
I'm going to try to bring this PR back to life and fix all the remaining issues, since this is just going to accumulate merge conflicts if we don't get it merged quickly.

richardxia:
Doing another attempt at restoring the linter. This time I've only committed the changes that were fixed automatically with `eslint --fix`, and for the remaining failures I've added the files to `.eslintignore`. This will at least allow us to commit these changes and start enforcing lint checks on new files while gradually fixing up the remaining files.

Since I more or less trust `eslint --fix` to only make safe changes, I'm going to just merge this in if it passes Travis.

PR #615, opened 2019-01-07
Fixed inheritance of timestamps for services, and the way we display timestamps

lexholden:


app/pages/ServiceListingPage.jsx
```diff
@@ -86,19 +86,19 @@ class ServicePage extends React.Component {
...
       ? [resource.address].map(address => ({
-          id: address.id,
-          address,
-          name: service.name,
-          schedule
-        }))
+        id: address.id,
+        address,
+        name: service.name,
+        schedule: schedule || resource.schedule,
```
richardxia: I don't know much about the actual data values that come back from the API, but is there a chance that the schedule is an empty list (e.g. `[]`), which would evaluate as truthy in JavaScript? In Rails we tend to use the `.blank?` method on objects which returns true if it is null or if it is an empty collection, and I don't know if we need to do a similar thing here in JavaScript.

lexholden: If no user has added a schedule to the service, pretty sure it will always be null. Someone could add a schedule with no schedule days, but that would have to be a manual effort.

PR #616, opened 2019-01-07
New header mobile

richardxia:
This adds the new navigation and hamburger menu for the mobile view. I'm running to the same issue @alexanderturinske is in #609, where not all the links in @derekfidler's mockups have actual pages that correspond to them. I've only included the links that actually have pages they point to.

The one, unrelated major thing that I did here was start moving over to [CSS Modules](https://github.com/css-modules/css-modules), which is a way of ensuring that our CSS is locally-scoped by automatically munging the CSS class names. So far only the `Navigation` and `HamburgerMenu` components use CSS modules, and we'll have to individually migrate components over to CSS modules by editing webpack.config.js. This is going to cause merge conflicts with @schroerbrian if he's already started on the desktop version of the nav bar, but if he hasn't, then it might be better to fork from this branch.

screen shot 2019-01-06 at 11 50 25 pm
screen shot 2019-01-06 at 11 50 33 pm
screen shot 2019-01-06 at 11 50 39 pm


app/components/App.jsx
```diff
@@ -130,21 +134,41 @@ class App extends Component {
     });
   }
 
+  toggleHamburgerMenu() {
+    this.setState((state, props) => ({hamburgerMenuIsOpen: !state.hamburgerMenuIsOpen}));
+  }
+
+  onHamburgerMenuStateChange(state) {
+    this.setState({hamburgerMenuIsOpen: state.isOpen});
```
alexanderturinske: is there a reason you set the state differently here than above in the toggleHamburgerMenu function? If not, I think it would be better to choose one way and be consistent.

app/components/ui/HamburgerMenu.jsx
```diff
@@ -0,0 +1,94 @@
...
+  toggleHamburgerMenu,
+}) => (
+  <Menu
+    isOpen={isOpen}
+    onStateChange={onStateChange}
+    outerContainerId={outerContainerId}
+    pageWrapId={pageWrapId}
+    right
+    styles={burgerStyles}
+    width="275px"
```
alexanderturinske: Why is this random style here and not in the burgerStyles above?

app/components/App.jsx
richardxia: This is the API that the [`react-burger-menu`](https://github.com/negomi/react-burger-menu) exposes. It expects you to pass in a callback where the first argument is the new state. The API is honestly a bit confusing since it looks like you can use the component in two ways:

1. react-burger-menu maintains its own state and you don't pass in the `isOpen` prop into the component
2. react-burger-menu maintains its own state *and* you can pass in the `isOpen` prop into the component, meaning that you must synchronize state both inside and outside of the component

I opted for 2) because based on Derek's designs, we needed a lot more control over the actual hamburger button than the library provides by itself. If you can think of a better way of handling this, I'm more than happy to change it.

app/components/ui/HamburgerMenu.jsx
richardxia: Once again, it's part of the react-burger-menu API, and I think they did it this way because this one width affects multiple of the subcomponents, since you want the width of the side menu, the distance to translate the side menu when opening it, and the distance to translate the rest of the page to all be the same.

richardxia:
I need to fix the TestCafe tests, which all seem to be failing now.

schroerbrian:
Thanks for the heads up @richardxia. I've completed the desktop version, but I haven't submitted a PR yet. I can try merging my commit into this branch locally to see if there will be conflicts and fix them. Fortunately, the desktop version did not introduce much new code. P.S. In the desktop specs, Derek said that the About link should link to https://www.sheltertech.org, so I suppose that you could add the About link to the mobile nav

schroerbrian:
One other thing to keep in mind is that Derek asked that the footer links be hidden in the mobile version since they will now be in the mobile nav in his spec on Quip: https://sheltertech.quip.com/K0ZMA1vgbxjd/SFServiceGuideorg-Landing-Page-Requirements

app/components/App.jsx
alexanderturinske: Sounds good to me

app/components/ui/HamburgerMenu.jsx
alexanderturinske: Ah. I see. Makes sense.

app/components/ui/Navigation.jsx
```diff
@@ -19,40 +24,65 @@ class Navigation extends React.Component {
...
-              className="nav-search search-container form-row"
               role="search"
             >
               <input
-                ref={(c) => { this.searchComponent = c; }}
+                className={styles.secondarySearchField}
+                ref={c => { this.searchComponent = c; }}
                 type="text"
-                className="search-field"
+                /* TODO: update placeholder text to include dynamic number of resources */
```
alexanderturinske: Were you going to address this TODO in this commit or is this for another one? If it is for another, could you create a ticket for it and put the ticket number in here?

app/components/ui/HamburgerMenu.jsx
```diff
@@ -0,0 +1,94 @@
...
+    pathname: PropTypes.string.isRequired,
+  }).isRequired,
+  onStateChange: PropTypes.func.isRequired,
+  outerContainerId: PropTypes.string.isRequired,
+  pageWrapId: PropTypes.string.isRequired,
+  toggleHamburgerMenu: PropTypes.func.isRequired,
+};
+
+const MenuItem = ({ children, isActive, onClick, to }) => (
+  (to.startsWith('http') || to.startsWith('mailto:')) ?
```
lexholden: Nice and clean :)

PR #623, opened 2019-02-10
Added eligibilities block on home page

lexholden:


app/pages/SearchPage.jsx
```diff
@@ -19,6 +19,7 @@ class SearchPage extends Component {
 
     this.state = { searchState: { ...qs.parse(props.router.location.query) } };
     this.onSearchStateChange = this.onSearchStateChange.bind(this);
+    console.log(this)
```
richardxia: We should probably remove these debugging `console.log()` calls.

app/pages/SearchPage.jsx
```diff
@@ -29,6 +30,7 @@ class SearchPage extends Component {
   }
 
   onSearchStateChange(nextSearchState) {
+    console.log('updating searchState', nextSearchState)
```
richardxia: This one too.

Issue #599, opened 2018-12-06
Replace current category listing block


jjfreund:
All featured categories should be retrieved and displayed in a single row. The current icons will need to be replaced with icons conforming to the new design.

PR #626, opened 2019-02-10
Update Featured Categories Block (#599)

AnnaKalkanis:
screen shot 2019-02-10 at 3 57 14 pm
screen shot 2019-02-10 at 3 57 06 pm


richardxia:
I'm not sure if I screwed something up while trying to resolve the merge conflicts that https://github.com/ShelterTechSF/askdarcel-web/pull/623 caused, but something looks odd about the page in my local setup:

screen shot 2019-02-10 at 8 47 32 pm

There's a big gap between the featured categories and the eligibilities, and looking at the code, that's where the un-featured categories are, but in the Figma mockups, I think the un-featured categories have been removed from the landing page. Assuming that we're not supposed to have un-featured categories anymore, I can try removing that section altogether on the page.

richardxia:
I pushed out a few commits to clean up the weird display things I was seeing and to also fix up some of the mobile views.

AnnaKalkanis:
@richardxia Awesome, thanks for that clean up. Your changes look good to me - should we merge this?

PR #629, opened 2019-02-21
Escape query parameter before using in URL.

richardxia:
This fixes the bug where we weren't escaping the category name, and where having an & in the category name would mess up the search URL. Tested locally to make sure that clicking on a category with an ampersand in the name yields a page with actual search results.


richardxia:
Thanks for the review!

PR #630, opened 2019-02-21
Feature eligibilities: use service_count value

cliffcrosland:
Each item returned from the `eligibilities/featured` API endpoint now
contains a `service_count` value. This more closely matches the number
of search results that the user will see when they click on a featured
eligibility.

See: https://github.com/ShelterTechSF/askdarcel-api/commit/fc636313dafc336d3a6bf1baf56b63910a9ec3b2


app/components/ui/LandingPageEligibilityBlock.jsx
```diff
@@ -33,6 +33,8 @@ LandingPageEligibilityBlock.props = {
     name: PropTypes.string,
     id: PropTypes.number,
     feature_rank: PropTypes.number,
+    service_count: PropTypes.number,
+    // TODO(cliff): Deprecate resource_count
     resource_count: PropTypes.number,
```
richardxia: You should be able to just delete this from the PropTypes if it's unused. You only need to declare the fields that are actually being used in the PropTypes, and it won't complain about extra things being passed in.

PR #639, opened 2019-02-25
633 show sfserviceguide logo if url is sfserviceguide.org. Swap Legal…

schroerbrian:
… and Basic needs block position per Derek's request.

This uses existing logic for displaying different logos depending on site domain. I had to tweak a couple config properties to make it work. I removed the whitelabel section, because it's not in the new design. Lastly, switched the Legal and Basic Needs blocks per Derek's request


config.docker.yml
```diff
@@ -6,4 +6,5 @@ ALGOLIA_INDEX_PREFIX: 'ALGOLIA_INDEX_PREFIX_REPLACE_ME'
 ALGOLIA_APPLICATION_ID: 'ALGOLIA_APPLICATION_ID_REPLACE_ME'
 ALGOLIA_READ_ONLY_API_KEY: 'ALGOLIA_READ_ONLY_API_KEY_REPLACE_ME'
 
-MOHCD_SUBDOMAIN: 'MOHCD_SUBDOMAIN_REPLACE_ME'
+# this is not sensitive data, so we just pass it in as is
+MOHCD_SUBDOMAIN: 'sfserviceguide'
```
richardxia: Actually, this shouldn't change. The magic string needs to be there so that the Docker container is able to substitute the value out at container runtime. This allows us to change the value without having to do a rebuild and push of the Docker image. I'm not sure if the MOHCD subdomain has been finalized yet, so I think there's still utility in keeping it a parameter.

schroerbrian: @richardxia In that case, what is the right way to go about this?

richardxia: I'd say that we should revert the changes back to what they were and make sure that the Docker containers in staging/prod have the right value set. I think we're keeping our container settings in https://github.com/ShelterTechSF/gcloud, but @lgarofalo can correct me if I'm wrong.

schroerbrian: Ok – I have reverted the change `config.docker.yml`. I have created a PR into the gcloud repo, and requested that you and/or Leland review it: https://github.com/ShelterTechSF/gcloud/pull/6

schroerbrian: @richardxia – I chatted with @lgarofalo and he advised that we add a new config value, MOHCD_DOMAIN, to the container settings. He's applied my PR here: https://github.com/ShelterTechSF/gcloud/pull/6

I've added the MOHCD_DOMAIN config to the relevant files in this repo. Please let me know if I'm missing something. Thank you!

app/components/ui/Footer.jsx
```diff
@@ -2,7 +2,7 @@ import React from 'react';
 import { Link } from 'react-router';
 import config from '../../config';
 
-const subDomain = window.location.host.split('.')[0];
+const domain = window.location.host.split('.')[0];
```
lgarofalo: This logic won't work for grabbing the root domain right?

schroerbrian: it does since we are passing in `sfserviceguide` rather than `sfserviceguide.org` as our MOCHD_DOMAIN value. i think it's arguably preferable over say just checking `window.location.host` since it will also work for `staging.sfserviceguide.org`

lgarofalo: Does this work for `staging.sfserviceguide.org` though? Won't that code find `staging` ?

lgarofalo: I think we update it to 
```
window.location.host.split('.')[1]
```
We would be ok

schroerbrian: ack you are right about that; however if we run `window.location.host.split('.')[1]` against `https://sfserviceguide.org` (without the prefixed www), it returns `.org`

i think the most robust/flexible solution, short of using a parsing library (i don't think we'd be able to get that in in time for this deploy), would be to check the window.location.host for a substring `sfserviceguide`– e.g. `window.location.host.indexOf(config.MOHCD_DOMAIN) > -1` 

i've tested that code against staging.sfserviceguide.org, http://sfserviceguide.org, and www.sfsergiveguide.org and it works for all three. 

i will update my PR

PR #655, opened 2019-03-14
Add Intercom.

richardxia:
I added the Intercom widget to askdarcel-web. It looks like we had an existing Workspace in Intercom, which is where the App ID comes from. I figured it's not too dangerous to have the app ID key here, and if people steal it or something, we can just generate a new one.

Here are some screenshots of it in our app:

Screen Shot 2019-03-13 at 7 30 26 PM
Screen Shot 2019-03-13 at 7 30 42 PM
Screen Shot 2019-03-13 at 7 31 20 PM

Here's what it looks like in the Intercom interface:

Screen Shot 2019-03-13 at 7 34 11 PM


richardxia:
It looks like this is unfortunately causing our Testcafe tests to fail. Will debug tonight.

richardxia:
I couldn't figure out how to make TestCafe work with the Intercom widget, so I added some code that disables Intercom in TestCafe: https://github.com/ShelterTechSF/askdarcel-web/pull/655/commits/a8392c0bbe8fe85ef62dd70d6b2b9f380da4da39

I think the issue is that Intercom attaches to the `window` object internally, which bypasses the TestCafe hammerhead proxy that allows the CLI tool to control the web browser.

My hack is not that great either, since the way we run TestCafe locally is to just have it run against your normal, locally-running webpack development server, but Intercom is enabled by default there. In order to run the tests locally now, you have to stop your docker-compose instances, and then start them with the `TESTCAFE_RUNNING` environment variable set.

PR #668, opened 2019-03-22
663 Set dynamic page title

richardxia:
This PR pulls in the [react-helmet](https://github.com/nfl/react-helmet) library, which makes it really easy to add things to the `<head>` tag including the `<title>` as well as various metadata tags. For now, this just sets the title based on whether the MOHCD domain is present. Here is a screenshot that proves that it works for me locally when I set the `MOHCD_DOMAIN` config variable to `localhost`.

Screen Shot 2019-03-21 at 8 31 28 PM

### DIFF
--- a/app/assets/index.js
+++ b/app/assets/index.js
@@ -22,7 +22,6 @@ if (isSFServiceGuideSite()) {
     mohcdSeal: require('../assets/img/sf-seal.png'),
     icon,
   };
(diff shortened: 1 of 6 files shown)


richardxia:
Also, note that I reverted @lgarofalo's change, since it doesn't make sense to try to set the page title in two different places.

PR #672, opened 2019-03-24
fix scrollbar of the Discover resources by category section

PPzhen000:


richardxia:
This makes the fix as described by https://github.com/ShelterTechSF/askdarcel-api/issues/390, but at the cost of making two changes to the design:

1. Scrolling has been removed at all screen sizes, including on mobile, even though the original intent was for the categories to overflow offscreen and be scrollable.
2. The category boxes are now center aligned, which makes it look odd for two reasons:
  - On mobile, if there is an odd number of boxes, the last row is center aligned, which looks strange
  - On desktop, the category boxes are the only thing on the page that aren't left-aligned

Screen Shot 2019-04-01 at 11 23 29 PM
Screen Shot 2019-04-01 at 11 27 43 PM

Is there a way to preserve the old behavior but still get rid of the scroll bars on Windows desktop when unnecessary?

derekfidler:
Yes, you can hide horizontal scrollbars cross browser and preserve horizontal scroll with this html and css trick. https://jsfiddle.net/ehp3qnty/576/

Op 2 apr. 2019 om 09:10 heeft Richard Xia <notifications@github.com> het volgende geschreven:

This makes the fix as described by ShelterTechSF/askdarcel-api#390, but at the cost of making two changes to the design:

Scrolling has been removed at all screen sizes, including on mobile, even though the original intent was for the categories to overflow offscreen and be scrollable.
The category boxes are now center aligned, which makes it look odd for two reasons:
On mobile, if there is an odd number of boxes, the last row is center aligned, which looks strange
On desktop, the category boxes are the only thing on the page that aren't left-aligned




Is there a way to preserve the old behavior but still get rid of the scroll bars on Windows desktop when unnecessary?

richardxia:
> Yes, you can hide horizontal scrollbars cross browser and preserve horizontal scroll with this html and css trick. https://jsfiddle.net/ehp3qnty/576/

I don't have an easily accessible Windows machine, but does this allow the horizontal scrollbar to still appear when the user actually wants to scroll?

I think another way of "solving" the problem is to replace the `overflow: scroll` properties with [`overflow: auto`](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow#Values), which only adds in the scroll bars if the inner content is wider than the parent.

derekfidler:
yes, it preserves scrolling behaviour but hides the scroll bar and it works across browsers and devices.

PPzhen000:
> yes, it preserves scrolling behavior but hides the scroll bar and it works across browsers and devices.
> 
> I've actually implemented this before. If you want I can take over this task.

I can give it a try first and see if I need help with this

lgarofalo:
I did an auto-rebase here since this feature branch was out of date, Github just gives a button to do that here, once the tests pass I can merge.

PR #684, opened 2019-03-28
Use fixed version of Chrome.

richardxia:
I think the SauceLabs/Travis tests are failing because SauceLabs slightly changed their API. Previously we were telling SauceLabs to use the `beta` version of Chrome, which would allow us to always be running on a new version of Chrome. At the time the only alternative was to run on a fixed version of Chrome.

It seems that since then, you are now able to specify `latest`, `latest-1`, or `latest-2` for the latest, second to latest, or third to latest versions of Chrome, so now I've updated our TravisCI script to run on the latest version: https://wiki.saucelabs.com/display/DOCS/Test+Configuration+Options

### DIFF
--- a/travis-testcafe.sh
+++ b/travis-testcafe.sh
@@ -51,7 +51,7 @@ sleep 60
 # Print out container logs in case if an error occurs
 docker logs api
 
-npm run testcafe -- 'saucelabs:chrome@latest:Windows 10' \
+npm run testcafe -- 'saucelabs:Chrome@latest:Windows 10' \
   --quarantine-mode \
   --skip-js-errors \
   --assertion-timeout 50000 \
--- a/travis-testcafe.sh
+++ b/travis-testcafe.sh
@@ -51,7 +51,13 @@ sleep 60
 # Print out container logs in case if an error occurs
 docker logs api
 
-npm run testcafe -- 'saucelabs:Chrome@latest-1:Windows 10' \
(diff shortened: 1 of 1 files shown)


richardxia:
This does not work, so I'll have to debug what the actual correct browser + version string to use now. Will try to do so later tonight.

richardxia:
It actually looks like SauceLabs was having issues last Wednesday: https://status.us-west-1.saucelabs.com/
Screen Shot 2019-04-01 at 9 34 47 PM

I'm just going to close this out.

richardxia:
OK, so bad news about this. I tried kicking off a rebuild of https://github.com/ShelterTechSF/askdarcel-web/pull/683, which still failed with the same SauceLabs issue. I eventually found https://github.com/DevExpress/testcafe-browser-provider-saucelabs/issues/39, which was someone with a similar issue, and now I've submitted a support ticket to SauceLabs, since I don't think we've done anything special here.

richardxia:
Reopening this for testing purposes. SauceLabs responded back to me and asked me to try a non-beta version of Chrome, since the beta and dev releases often break compatibility with Selenium.

richardxia:
It looks like choose a fixed, stable version of Chrome rather than a "beta" version makes this work. Will merge this in if it finally passes Travis CI and squash merge to get rid of all the intermediate attempts.

PR #686, opened 2019-04-02
642 helmet meta tags

chrisco23:
Dynamically add meta title from name and meta description from long_description.

### DIFF
--- a/app/pages/OrganizationListingPage.jsx
+++ b/app/pages/OrganizationListingPage.jsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import ReactMarkdown from 'react-markdown';
 import { Link } from 'react-router';
+import Helmet from 'react-helmet';
 import PropTypes from 'prop-types';
 import {
   AddressInfo, PhoneNumber, ResourceCategories, Website, Email, StreetView,
@@ -80,6 +81,11 @@ export class OrganizationListingPage extends React.Component {
     // console.log(resource, window.google);
     return (!resource || !window.google ? <Loader />
       : (
+        <div>
+          <Helmet>
+            <title>{ resource.name }</title>
+            <meta name="description" content={ resource.long_description }></meta>
+          </Helmet>
         <div className="org-container">
           <article className="org" id="resource">
             {
@@ -224,6 +230,7 @@ export class OrganizationListingPage extends React.Component {
               </div>
             </div>
           </article>
+        </div>
         </div>
       )
     );
--- a/app/pages/OrganizationListingPage.jsx
+++ b/app/pages/OrganizationListingPage.jsx
@@ -82,7 +82,7 @@ export class OrganizationListingPage extends React.Component {
     // console.log(resource, window.google);
     return (!resource || !window.google ? <Loader />
       : (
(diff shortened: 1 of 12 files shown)


app/pages/OrganizationListingPage.jsx
```diff
@@ -80,6 +81,11 @@ export class OrganizationListingPage extends React.Component {
     // console.log(resource, window.google);
     return (!resource || !window.google ? <Loader />
       : (
+        <div>
```
richardxia: Rather than creating a `<div>` to wrap everything just for the sake of being able to put the `<Helmet>` component somewhere, let's use [`<React.Fragment>`](https://reactjs.org/docs/fragments.html), which is the official way of allowing a component to return multiple sibling components without needing to create an unnecessary HTML `<div>` element.

app/pages/OrganizationListingPage.jsx
```diff
@@ -80,6 +81,11 @@ export class OrganizationListingPage extends React.Component {
...
+          <Helmet>
+            <title>{ resource.name }</title>
```
richardxia: Can you also put the site name after the resource name?

app/pages/ServiceListingPage.jsx
```diff
@@ -106,97 +106,103 @@ class ServicePage extends React.Component {
...
-              <header>
-                <h1>{service.name}</h1>
-                {service.alsoNamed ? <p>Also Known As</p> : null}
-                <p>
-                  A service
-                  {/* TODO Implement rendering/popover when programs exist */}
-                  {program ? (
+      <div>
+        <Helmet>
+          <title>{ service.name } - AskDarcel</title>
```
richardxia: The site name itself should be parameterized based on whether we are on the whitelabeled version of the site, which uses the name "SF Service Guide". There's a helper function somewhere that abstracts that away, since it's something that depends on the domain name you view the site from.

Also, I would suggest perhaps using a vertical bar (`|`) instead of a dash here, since some resource/service names actually have dashes in them already, but it's much less likely that a vertical bar would be used in a resource/service name.

chrisco23:
I have made these changes, but React.Fragment isn't working for me--service page sticks on spinner, spits out a warning about React.createElement.  Probably we can't use this yet because React is still on 15.6.2 in master?

richardxia:
> I have made these changes, but React.Fragment isn't working for me--service page sticks on spinner, spits out a warning about React.createElement. Probably we can't use this yet because React is still on 15.6.2 in master?

Ah, I forgot that we were on such an old version of React. In that case, we can stick to `<div>`s for now, and we should consider upgrading React soon. Feel free to push your latest changes to this branch, and we can get this merged in as soon as we can.

chrisco23:
I'm still failing npm tests on the various open/close things though.  Do I need to do something more?

chrisco23:
At one point I asked Alex about testcafe etc. and he suggested I run 'npm
run test', which produces: https://pastebin.com/zMBptU8G

Looks to me like it would be the timestamp issues you mention.

richardxia:
Yep, LGTM. I'll go ahead and merge this in.

PR #697, opened 2019-04-07
625 fix cached hours on service page. No need to create state on chil…

schroerbrian:
…d MapOfLocations component.

Per https://reactjs.org/docs/lifting-state-up.html – it seems best to let the MapOfLocations inherit state from its parent component rather than store the values locally since the local values won't automatically update when parent props change.


Maxastuart:
@lgarofalo

PR #698, opened 2019-04-07
Refactor OrganizationEditPage

richardxia:
I fixed up the lint errors on the OrganizationEditPage and the EditScheduleDay component, and as a result, was able to remove a bunch of dead code. I tested this by making a bunch of edits locally and checking that the edit page still functions.


richardxia:
@jjfreund, fixed the bug that TestCafe caught, and it should be passing now. Let me know if this looks good to merge.

app/pages/OrganizationEditPage.jsx
```diff
@@ -224,70 +245,94 @@ export class OrganizationEditPage extends React.Component {
...
+        postSchedule(currentService.scheduleObj, promises);
+        delete currentService.scheduleObj;
+        if (!_.isEmpty(currentService)) {
+          promises.push(dataService.post(uri, { change_request: currentService }));
+        }
+      }
+    });
+
+    console.log('newServices');
+    console.log(newServices);
```
AnnaKalkanis: You still have some console.logs in here

richardxia: Nice catch! I just removed them, and I can merge as soon as Travis passes.

PR #699, opened 2019-04-07
666 favicon

PPzhen000:
Change SFserviceguide favicon to lifesaver ring


richardxia:
@PPzhen000, you have a number of linter failures that you should fix. In addition, you should add the rest of the images for the various mobile devices in a similar way.

Issue #637, opened 2019-02-24
Add checkbox to allow a new service to inherit its schedule from its parent resource


jjfreund:
A new flag is being added to the service object, inherit_schedule, which will determine whether or not it inherits its schedule from its parent resource. Add this field to the add / edit service screen.

Maxastuart:
If this is moved back to to-do, is it also open for reassignment? (feel free to unassign yourself @AnnaKalkanis ) Or is it blocked by a dependency? I reckon it can remain in In Progress if you're happy to do it after a dependency is cleared

AnnaKalkanis:
I moved this back to ‘to do’ because https://github.com/ShelterTechSF/askdarcel-api/issues/176 blocks it. I’m still planning on doing it once that is done, but seemed weird to have it in progress if there’s no actual progress being made on it.

PR #701, opened 2019-04-13
#637 - Add ability for service to inherit schedule from parent resource

AnnaKalkanis:
When editing a service, there will now be an option to have the service inherit its schedule from its parent:

Screen Shot 2019-04-13 at 9 42 06 AM

When selected, the edit hours table will be hidden:
Screen Shot 2019-04-13 at 9 42 13 AM

Note that resources (top level) won't have this option.

Screen Shot 2019-04-13 at 9 42 19 AM

### DIFF
--- a/app/components/edit/EditSchedule.jsx
+++ b/app/components/edit/EditSchedule.jsx
@@ -1,4 +1,5 @@
 import React, { Component } from 'react';
+import _ from 'lodash';
 import { stringToTime } from '../../utils/index';
 import EditScheduleDay from './EditScheduleDay';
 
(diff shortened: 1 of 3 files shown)


app/components/edit/ProvidedService.jsx
```diff
@@ -192,6 +192,7 @@ Service
           </li>
 
           <EditSchedule
+            canInheritFromParent
```
richardxia: Just wondering, when would we not want to set this flag to true?

AnnaKalkanis: Ya, that's a good question. We reuse `EditSchedule` for organizations (resources?) and for services. We only want services to be able to inherit schedules since organizations don't have a 'parent' to inherit from.

app/components/edit/EditSchedule.jsx
```diff
@@ -1,4 +1,5 @@
 import React, { Component } from 'react';
+import _ from 'lodash';
```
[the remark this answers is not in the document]
AnnaKalkanis: Ya, that's a good call!

PR #702, opened 2019-04-14
Add hack to make times display correctly on Resource show page.

richardxia:


richardxia:
I had to skip a bunch of unit tests that are failing because they are testing for behavior that is only correct during Daylight Savings. The tests are probably running in UTC because they're running in Node.js on Travis.

Issue #191, opened 2017-04-08
Add check box for "no physical location" when adding/editing resource


jjfreund:
Should remove the address object when deselected.

quanhuynh:
@jjfreund what should be the behavior? does the address section disappear or do the text boxes just become uneditable?

derekfidler:
I think it would be good to remove the address section for online-only resources. i'm not sure how we display these in the search results though. are they just centered on SF?

Maxastuart:
@quanhuynh whichever is easier for the time being, and make a new ticket if making the boxes disappear remains to be done. 

about derek's query, on the backend if the default doesn't automatically become "San Francisco, CA" let's give them city hall's address? or the library for online access? A later update to make search result ranking not purely based on location is coming.

quanhuynh:
i added the checkbox to remove the address field. currently it just sends an empty address to the backend which is the behavior as if the user didn't put in an address. i will make a ticket to make the default address in the backend city hall's address

PR #716, opened 2019-04-21
add no location checkbox to edit form

quanhuynh:
see https://github.com/ShelterTechSF/askdarcel-web/issues/191
unchecked
checked

### DIFF
--- a/app/components/edit/EditAddress.jsx
+++ b/app/components/edit/EditAddress.jsx
@@ -3,8 +3,9 @@ import React, { Component } from 'react';
 class EditAddress extends Component {
   constructor(props) {
     super(props);
-    this.state = { address: {} };
+    this.state = { address: {}, noLocation: hasNoLocation(props.address) };
     this.handleAddressChange = this.handleAddressChange.bind(this);
+    this.handleNoLocationChange = this.handleNoLocationChange.bind(this);
   }
 
   handleAddressChange(e) {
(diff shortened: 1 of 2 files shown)


app/components/edit/EditAddress.jsx
```diff
@@ -17,79 +18,142 @@ class EditAddress extends Component {
...
     return (
       <li key="address" className="edit--section--list--item">
         <label htmlFor="address">Address</label>
+        <input
+          type="checkbox"
+          className="input-checkbox"
+          checked={noLocation}
+          onChange={handleNoLocationChange} 
+        />
+        <div onClick={handleNoLocationChange}>No Physical Location</div>
```
richardxia: Looks like this is playing the role of a label. How about you actually make it a `<label>` tag? If you have the label tag be a parent of the `<input>`, then when you click on the label text, it'll natively check the checkbox without you having to pass in an `onClick` handler. This is what I mean:

```html
<label>
  <input ...>
  No Physical Location
</label>
```

If you don't want to nest the `<input>` inside of the `<label>`, then you can give the `<input>` an `id` and have the `<label>`'s `htmlFor` attribute point to that id (note that the way we're using htmlFor everywhere else on the page is actually wrong and doesn't do anything).

app/components/edit/EditAddress.jsx
```diff
@@ -17,79 +18,142 @@ class EditAddress extends Component {
...
+    return (
+      <li key="address" className="edit--section--list--item">
+        <label htmlFor="address">Address</label>
+        <input
+          type="checkbox"
+          className="input-checkbox"
+          checked={noLocation}
+          onChange={handleNoLocationChange} 
+        />
+        <div onClick={handleNoLocationChange}>No Physical Location</div>
```
richardxia: This seems to be an exact duplicate of the stuff above. Could you structure the `if` statement to make the common HTML always present and only conditionally hide the other stuff? You can do [conditional rendering](https://reactjs.org/docs/conditional-rendering.html#inline-if-with-logical--operator) of components directly in JSX instead of in a normal JavaScript `if` statement.

quanhuynh: doing conditional rendering on all adjacent input tags except one (the checkbox) caused the error "adjacent jsx elements must be wrapped in an enclosing tag". the alternative is to wrap all the conditional inputs into one <div>. i'll make that change and you can check it out

PR #717, opened 2019-04-21
Hoist state out of FormTextArea

richardxia:
## Overview
As I continue with refactoring the frontend app in order to fix the schedule handling code, I've been trying to hoist state up the component hierarchy. I think a pretty big problem that we have with our current code is that there is React Component state that is duplicate at several different layers, and state changes have to be propagated upwards and applied at every layer.

## Main Change
This PR represents one step of the refactoring, where I've hoisted state out of the `<FormTextArea>` component. In this component the state is less obvious, but because we're using a normal HTML form element, there is actually implicit state on the `<textarea>` element. React does give you the ability to let a parent component manage the state of a form element, which is what they call a [controlled component](https://reactjs.org/docs/forms.html#controlled-components). This is done by simply passing in the `value` prop into the form element and having the `onChange` prop be in charge of updating the state from higher up.

## Cleaner Architecture
A second thing I am trying out is having a cleaner API for interacting between the different layers of the React components. We already have a pattern of passing around `handleChange` or `onChange` functions between components, but they are often coupled to the implementation details of the lower components.

For example, in this PR, `<ProvidedService>` is a parent of `<FormTextArea>`. `<ProvidedService>` previously had a `handleFieldChange()` method that looked like this:

```jsx
  handleFieldChange(e) {
    const { service } = this.state;
    service[e.target.dataset.field] = e.target.value;
    this.handleChange(service);
  }
```

This `handleFieldChange` would get passed to `<FormTextArea>` or other child components, and it was expected to be attached directly to the `onChange` prop of a form input element.

This is a fragile API because it's tightly coupled to a few things at the lower levels:

- It assumes that there is a single argument which is the DOM event (which is expected to have a `.target` attribute)
- It assumes that the form input element is the DOM event target
- It assumes that the form input element has a [`data-field` attribute](https://developer.mozilla.org/en-US/docs/Learn/HTML/Howto/Use_data_attributes) on it that holds the name of the key on the `service` state/prop corresponding to the field being updated. This is what that `.dataset.field` thing is.

I'm proposing a new convention which is modeled after the new [React Hooks](https://reactjs.org/docs/hooks-overview.html) API, even though it does not use React Hooks today. The idea is that you pass a function as a prop to the lower component which only takes in the new value of the state and does not have to do any "unpacking" of the event object. This is modeled after the `const [count, setCount] = useState()` API of React Hooks, where the `value` and `setValue()` always come in pairs. This is only slightly different than what we had before with the `handleFieldChange()` function described above, in that the arguments are a bit different, but I think it has major benefits:

Because the arguments are only the new field value, the lower component doesn't _have_ to use DOM event, and the DOM event is handled in the lower component so that the upper component doesn't have to deal with it. This offers a cleaner separation of responsibilities where the upper component doesn't even need to be aware of the DOM at all.

Most importantly, I think this nests much better than our previous approach, and this is probably how the React team came to this design. As I hoist more and more state to the top of the component hierarchy, we can just repeat this pattern where each layer doesn't need to know how the lower layer will implement the code, but provides the most minimal API for getting and setting state.

### DIFF
--- a/app/components/edit/EditServices.jsx
+++ b/app/components/edit/EditServices.jsx
@@ -32,12 +32,13 @@ class EditServices extends Component {
    * @returns {void}
    */
   handleServiceChange(key, service) {
+    const { handleServiceChange } = this.props;
     const { services } = this.state;
     services[key] = service;
     this.setState({
       services,
     }, () => {
-      this.props.handleServiceChange(this.state);
+      handleServiceChange(this.state);
     });
   }
 
@@ -45,8 +46,8 @@ class EditServices extends Component {
    * @description Creates a brand new service
    */
   addService() {
-    const { existingServices } = this.state;
-    const newUUID = this.state.uuid - 1;
+    const { existingServices, uuid } = this.state;
+    const newUUID = uuid - 1;
 
     existingServices.push({
       id: newUUID,
@@ -63,22 +64,25 @@ class EditServices extends Component {
   }
 
   render() {
+    const { handleDeactivation } = this.props;
+    const { existingServices } = this.state;
     return (
       <li className="edit--section--list--item">
         <ul className="edit--section--list--item--sublist edit--service--list">
           {
-            this.state.existingServices.map((service, index) => (
+            existingServices.map((service, index) => (
               <ProvidedService
                 key={service.key}
                 index={index}
                 service={service}
                 handleChange={this.handleServiceChange}
-                handleDeactivation={this.props.handleDeactivation}
+                handleDeactivation={handleDeactivation}
               />
             ))
           }
         </ul>
         <button
+          type="button"
           className="edit--section--list--item--button new-service"
           id="new-service-button"
           onClick={this.addService}
--- a/.eslintignore
+++ b/.eslintignore
@@ -8,9 +8,7 @@ app/components/edit/EditCollection.jsx
 app/components/edit/EditNotes.jsx
 app/components/edit/EditPhones.jsx
 app/components/edit/EditSidebar.jsx
-app/components/edit/FormTextArea.jsx
 app/components/edit/MultiSelectDropdown.jsx
-app/components/edit/ProvidedService.jsx
 app/components/layout/CategoryItem.jsx
 app/components/layout/CategoryList.jsx
 app/components/layout/FindHeader.jsx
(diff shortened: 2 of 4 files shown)


app/components/edit/EditServices.jsx
```diff
@@ -32,21 +31,16 @@ class EditServices extends Component {
...
   handleServiceChange(key, service) {
-    const { services } = this.state;
-    services[key] = service;
-    this.setState({
-      services,
-    }, () => {
-      this.props.handleServiceChange(this.state);
-    });
+    const { handleServiceChange } = this.props;
+    handleServiceChange({ services: { [key]: service } });
```
AnnaKalkanis: I think this might be changing behavior. We were previously pulling all the services off state, updating the one we changed, and then passing all the services back into `handleServiceChange`. Here we are creating an empty services object with just the service that is changing.

I personally prefer the new approach of just passing the service that is changing (I might even argue we can just pass the key & service as parameters instead of wrapping them in a 'services' object). But, I think we need to update the `handleServiceChange` (from props) that is handling this to make sure it isn't overwriting the services that exist at that parent component level and instead just merging this changed one in with its existing services.

app/components/edit/EditServices.jsx
```diff
@@ -32,21 +31,16 @@ class EditServices extends Component {
...
   }
 
   /* @method addService
    * @description Creates a brand new service
    */
   addService() {
-    const { existingServices } = this.state;
-    const newUUID = this.state.uuid - 1;
+    const { existingServices, uuid } = this.state;
+    const newUUID = uuid - 1;
```
AnnaKalkanis: Wait a minute... I know this is existing code, but this is funky. We should probably just generate a new random UUID here.

AnnaKalkanis: Oh wait, these aren't even UUIDs, it starts as -1 and then each time you add a service it subtracts 1 from it. I have no idea what is going on here or why. I guess we can just ignore this for now - wasn't really the point of this MR anyway.

AnnaKalkanis:
This is great! Just unsure about the one piece where the behavior looks like it may have changed a bit.

app/components/edit/EditServices.jsx
richardxia: Good call. Let me revert this one change back, and I can take care of refactoring the handlers in a different PR.

app/components/edit/EditServices.jsx
richardxia: Yeah, I went through the exact same thought process here. I dug a little further and it's relying on the `-1` factor, since we have some _other_ code the partitions the services by ones with a positive "uuid" and ones with a negative "uuid". What it's really trying to do is keep track of new services which don't have an ID, so we really should handle that more explicitly.

PR #721, opened 2019-04-23
Componentize resource listing actions bar

AnnaKalkanis:
The new designs for the Organization Listing Page and the Service Listing Page are very similar. So, in implementing the Organization Listing Page redesign, I'm attempting to pull out as many shared components as possible. I started with the actions bars (we have one that we display on desktop, and another that we display on mobile).

Designs can be found on figma, here: https://www.figma.com/file/H0kD6ZZPhPR2RMyNJHWWeTtI/Web-Client?node-id=1456%3A1223

These components were partially implemented in the Service Listing Page, already. In this PR I further componentized them by moving more logic into the components and cleaning up the css. Once they were more self-contained, I was able to replace the actions bar on the Organization Listing Page with these components.

### Service
Screen Shot 2019-04-22 at 9 35 26 PM
### Org
Screen Shot 2019-04-22 at 9 35 09 PM

### Service
Screen Shot 2019-04-22 at 10 37 58 PM
### Org
Screen Shot 2019-04-22 at 10 38 07 PM

### DIFF
--- a/app/components/listing/ActionSidebar.jsx
+++ b/app/components/listing/ActionSidebar.jsx
@@ -2,32 +2,46 @@ import React from 'react';
 import PropTypes from 'prop-types';
 
 import { Link } from 'react-router';
+import { getResourceActions } from 'utils/ResourceActions';
+
+const getSidebarActions = resource => {
+  const resourceActions = getResourceActions(resource);
+  const sidebarActions = [
+    resourceActions.edit,
+    resourceActions.print,
+  ];
+  if (resource.address) {
+    sidebarActions.push(resourceActions.directions);
+  }
+  return sidebarActions;
+};
+
+const renderButtonContent = action => (
+  <span>
+    <i className="material-icons">{ action.icon }</i>
+    { action.name }
+  </span>
+);
 
 class ListPageSidebar extends React.Component {
-  renderButtonContent(action) {
-    return (
-      <span>
-        <i className="material-icons">{ action.icon }</i>
-        { action.name }
-      </span>
-    );
-  }
-
   render() {
+    const { resource } = this.props;
+    const actions = getSidebarActions(resource);
+
     return (
       <ul className="actions">
-        {this.props.actions.map(action => (
+        {actions.map(action => (
           <li key={action.name}>
             {
               action.to || action.handler
                 ? (
                   <Link to={action.to} onClick={action.handler} className={`listing--aside--${action.name.toLowerCase()}`}>
-                    { this.renderButtonContent(action) }
+                    { renderButtonContent(action) }
                   </Link>
                 )
                 : (
-                  <a href={action.link} target="_blank" className={`listing--aside--${action.name.toLowerCase()}`}>
-                    { this.renderButtonContent(action) }
+                  <a href={action.link} rel="noopener noreferrer" target="_blank" className={`listing--aside--${action.name.toLowerCase()}`}>
+                    { renderButtonContent(action) }
                   </a>
                 )
             }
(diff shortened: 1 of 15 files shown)


app/components/listing/MobileActionBar.jsx
```diff
@@ -2,27 +2,45 @@ import React from 'react';
 import PropTypes from 'prop-types';
 
 import { Link } from 'react-router';
+import { getResourceActions } from 'utils/ResourceActions';
 import { images } from '../../assets';
 
+import './MobileActionBar.scss';
+
+const getMobileActions = resource => {
```
AnnaKalkanis: Just moved this from ServiceListing.jsx

app/components/listing/ActionSidebar.jsx
```diff
@@ -2,32 +2,48 @@ import React from 'react';
...
-  renderButtonContent(action) {
-    return (
-      <span>
-        <i className="material-icons">{ action.icon }</i>
-        { action.name }
-      </span>
-    );
+import './ActionSidebar.scss';
+
+const getSidebarActions = resource => {
```
AnnaKalkanis: Just moved this from ServiceListing.jsx

app/styles/components/_listing-page.scss
```diff
@@ -262,101 +262,16 @@
...
 ////////////////////////////////////////////////////////////////////////////////////////
 // ORG ASIDE
 
 .listing--aside {
-	max-height: 100vh;
-	padding: 0 0 0 calc-em(30px);
-	width: 100%;
-	max-width: calc-em(240px);
-	&--content {
-		width: 100%;
```
AnnaKalkanis: As far as I can tell, none of this was being used anywhere

app/styles/st-base/_layout.scss
```diff
@@ -64,37 +64,6 @@ pre {
   }
 }
 
-// Listing page action buttons
```
AnnaKalkanis: This was all specific to the ActionsSidebar, so I moved it into there

app/utils/ResourceActions.js
```diff
@@ -0,0 +1,30 @@
+export const getResourceActions = resource => ({
+  // TODO Edit should add service ID header
```
AnnaKalkanis: Just pulled this out of ServiceListing.jsx

app/components/listing/MobileActionBar.jsx
```diff
@@ -2,27 +2,45 @@ import React from 'react';
...
+  const resourceActions = getResourceActions(resource);
+  const mobileActions = [
+    { ...resourceActions.edit, icon: 'edit-blue' },
```
richardxia: Non-blocking comment: I wonder if it would be better for the ResourceActions list to include both mobile and desktop icons. Something feels a little bit off about having to override the icons with the mobile, blue icons here.

app/components/listing/MobileActionBar.jsx
```diff
@@ -2,27 +2,45 @@ import React from 'react';
...
+  ];
+  if (resource.address) {
+    mobileActions.unshift({ ...resourceActions.directions, icon: 'directions-blue' });
```
richardxia: Also non-blocking comment: I think this `unshift`ing logic would be easier to follow if we stuck to just `push`ing onto the array. Another way we could write this is to initialize an empty array, and then conditionally append in order rather than prepending. E.g.

```jsx
const mobileActions = [];

if (resource.phones && resource.phones.length > 0) {
  mobileActions.push({ ... });
}

if (blah) {
  mobileActions.push(blah);
}

mobileActions.push(thingThatShouldAlwaysBePresent);
```

PR #722, opened 2019-04-24
Remove map from resource page

AnnaKalkanis:
In implementing the organization page redesign, I'm attempting to break it up into somewhat contained PRs. This one just deals with removing the map from the page header.

Designs are in figma here: https://www.figma.com/file/H0kD6ZZPhPR2RMyNJHWWeTtI/Web-Client?node-id=1456%3A1223

### Before
Screen Shot 2019-04-23 at 9 08 16 PM
Screen Shot 2019-04-23 at 9 08 10 PM

### After
Screen Shot 2019-04-23 at 9 07 32 PM
Screen Shot 2019-04-23 at 9 07 45 PM


Maxastuart:
Thanks for this Anna! On a double check, it looks like this wasn't supposed to be completely taken out, but moved to lower on the page (the same as it's currently on the Service pages) @AnnaKalkanis

AnnaKalkanis:
@Maxastuart My bad, I meant to follow up with a PR to add the new map and totally forgot. Here it is:

https://github.com/ShelterTechSF/askdarcel-web/pull/750

PR #729, opened 2019-04-29
Refactor ProvidedService

richardxia:
This PR does some refactoring to the `ProvidedService` component on the edit page. I think the main things I succeeded at improving are:

1. Changing various simple text input fields from stateful components to controlled form components, thereby hoisting the state into the `ProvidedService` component,
2. Collapsing the `handleFooChange()` methods into a single method, and
3. Adding a `flattenedService` variable to the `render()` method that collapses both the version of the service passed in as a prop with the version of the service which is part of the state. For some reason, much of our code only stores the changes to objects in the state, but not the whole object, so it makes it trickier to control children components when the prop and the state have separate, incomplete info. With this `flattenedService` variable, I hope to simplify the code by being able to pass in a view of the object with all changes applied to it.

### DIFF
--- a/app/components/edit/FormTextArea.jsx
+++ b/app/components/edit/FormTextArea.jsx
@@ -1,17 +1,17 @@
-import React from 'react';
+import React, { Fragment } from 'react';
 import PropTypes from 'prop-types';
 
 const FormTextArea = ({
   label, placeholder, value, setValue,
 }) => (
-  <li className="edit--section--list--item">
+  <Fragment>
     <label htmlFor="textarea">{label}</label>
     <textarea
       placeholder={placeholder}
       value={value}
       onChange={evt => setValue(evt.target.value)}
     />
-  </li>
+  </Fragment>
 );
 
 FormTextArea.propTypes = {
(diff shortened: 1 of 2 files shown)


app/components/edit/FormTextArea.jsx
```diff
@@ -4,14 +4,14 @@ import PropTypes from 'prop-types';
 const FormTextArea = ({
   label, placeholder, value, setValue,
 }) => (
-  <li className="edit--section--list--item">
+  <React.Fragment>
```
AnnaKalkanis: woo! 16!

AnnaKalkanis: I kind of prefer (but totally up to you)

`import React, { Fragment } from 'react';`

```
<Fragment>
...
</Fragment>
```

richardxia: I have no strong preference, so I went ahead and changed it to `<Fragment>`.

I am kind of interested in the `<></>` syntax for fragments, but it looks like that'll require a Babel upgrade.

PR #740, opened 2019-05-05
Fix lint on the remaining Edit components.

richardxia:
I just went ahead and fixed all the lint issues on all the Edit components. The diff probably looks less scary when you ignore whitespace changes. I did some basic testing locally to make sure that the app still seems to function.


richardxia:
Thanks for the review!

PR #742, opened 2019-05-05
373 blank fields in services

shadycat:


shadycat:
Ah, I forgot to run lint locally! I just fixed them up -- hoping everything else in CI passes!

shadycat:
I see some test cafe errors, i'll look into that.

app/components/listing/Notes.jsx
```diff
@@ -2,7 +2,7 @@ import React from 'react';
 import ReactMarkdown from 'react-markdown';
 
 export default function Notes(props) {
-  return (
+  return props.notes && (
```
richardxia: It looks like I made a conflicting change in https://github.com/ShelterTechSF/askdarcel-web/pull/795, where I hoisted the guards to the calling component. I'm guessing this will require updating the tests you added here.

PR #745, opened 2019-05-09
Bump Node JS version to 10.15.

richardxia:
I think these are the only two places where the Node JS version is specified. The .nvmrc file is what Travis CI reads, and the docker-compose.yml file is where our local Docker-based development workflow sets the Node JS version.

I haven't done super thorough testing, but from just randomly clicking around on the site, it seems to be working fine.

Note that when this merges in, everyone using the Docker development environment will need to blow away their `node_modules/` directory and reinstall everything. This is because some of the things in there required compiling native extensions, which are dynamically linked to libraries on the host OS, which might have changed when bumping the Node JS Docker image version.


richardxia:
Looks like TestCafe is hanging for some reason. I'll look into this later.

lgarofalo:
What prompted you to want to upgrade this, out of curiosity?

richardxia:
Mostly in order to get the latest version of NPM that is bundled with the node Docker image. Today if people use different versions of NPM and run `npm install`, it might touch 90% of the lines in package-lock.json because they slightly changed the format of that file, and it gets kind of annoying to ping pong back and forth between newer and older versions of that file in PRs.

But also, Node 8 is EOL at the end of this year, so I figured it's not a bad idea to try to upgrade earlier.

richardxia:
Looks like it was a transient failure, since rebuilding it succeeded.

PR #747, opened 2019-05-12
Fix browser console warnings on Edit page.

richardxia:
I got a little tired of the console warnings on the Edit page, so I went ahead and fixed all the ones that I encountered. This should make it a bit easier to develop, since the console won't be cluttered with warnings.

### DIFF
--- a/app/components/edit/ProvidedService.jsx
+++ b/app/components/edit/ProvidedService.jsx
@@ -14,7 +14,7 @@ const InputField = ({
     <input
       type={type}
       placeholder={placeholder}
-      value={value}
+      value={value || ''}
       onChange={evt => setValue(evt.target.value)}
     />
   </Fragment>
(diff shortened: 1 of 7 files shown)


app/components/edit/ProvidedService.jsx
```diff
@@ -14,7 +14,7 @@ const InputField = ({
     <input
       type={type}
       placeholder={placeholder}
-      value={value}
+      value={value || ''}
```
drcaramelsyrup: Do you still require this when you've already set the empty string as the default argument in `defaultProps`?

richardxia: Yeah, apparently React chooses to [handle `null` differently from `undefined`](https://github.com/facebook/react/issues/2166), and only `undefined` or not attempting to pass in the prop at all will trigger the default prop. If you pass in a `null`, then it will still get the value of null.

PR #748, opened 2019-05-12
Hoist state out of ProvidedService.

richardxia:
I think this is the last Component that I need to move state out of before I can finally work on EditSchedule. With this change, there are no layers of state between the OrganizationEditPage and the EditSchedule, so then I can more easily hoist state out of EditSchedule, and then afterwards finally fix the multiple implementations of timezone handling.

This PR is much less scary to look at if you look at the diff with whitespace changes ignored.


AnnaKalkanis:
lgtm!

PR #751, opened 2019-05-13
fix error on resource display when no telephone

AnnaKalkanis:
Noticed that after creating a resource without a telephone number, it was throwing an error when trying to display the resource. This is a simple fix for that error.

Before:
Screen Shot 2019-05-12 at 11 35 13 PM

After:
Screen Shot 2019-05-12 at 11 35 40 PM

### DIFF
--- a/app/utils/ResourceActions.js
+++ b/app/utils/ResourceActions.js
@@ -1,3 +1,5 @@
+import _ from 'lodash';
+
 export const getResourceActions = resource => ({
   // TODO Edit should add service ID header
   edit: {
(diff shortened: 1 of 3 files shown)


app/utils/ResourceActions.js
```diff
@@ -22,7 +24,7 @@ export const getResourceActions = resource => ({
   phone: {
     name: 'Call',
     icon: 'call',
-    link: `tel:${resource.phones[0].number}`,
+    link: `tel:${_.get(resource, 'phones[0].number')}`,
```
richardxia: Should we not display this button at all if the resource doesn't have a phone number?

AnnaKalkanis: Good call! Updated

PR #753, opened 2019-05-14
[735] Upgrade to Babel 7 from Babel 6

drcaramelsyrup:
Originally I had looked into this in order to use Jest. Regardless of our unit testing framework, Babel has deprecated a lot of old presets (stage 0 experimental features, yearly presets like es2015, etc.) more than two years ago, and it seems sensible to upgrade our major Babel version in addition to the more future-proof Babel presets.

I've tested our mocha unit tests, testcafe, and resource / service edit flow.


drcaramelsyrup:
Gotcha, thanks for fixing that!

PR #754, opened 2019-05-15
746 add axios config

tsyaeger:
This adds the basic axios config and all requests from the Postman collection. Config does not yet include auth headers. There is no abstraction but for auxiliary objects, so may need refactoring.


app/api/resourceService.js
```diff
@@ -0,0 +1,44 @@
...
+  const changeRequest = { verified_at: new Date().toISOString() };
+  return api.post(`/resources/${id}/change_requests`,
+    { change_request: changeRequest });
+};
+
+export const certifyResourceHAP = id => {
+  api.post(`/resources/${id}/certify`);
+};
+
+export // let query = { change_request: { status: '2' }
```
schroerbrian: nit but maybe this comment should be deleted so that the deactivateResource definition is on the same line as the export statement

schroerbrian:
LGTM! Just a small nit or two about leftover comments

app/api/resourceService.js
tsyaeger: Thanks! all done.

Issue #736, opened 2019-05-05
Add filter containing eligibilities to restrict search results


jjfreund:
Add a search filter at the top of the search results page containing a simple dropdown of eligibilities. Selecting any of these should hide all results not flagged with that eligibility.

MOHCD will provide a list of these eligibilities. Some will be compound, and should restrict results to those services that match any of a list of eligibilities in our db. A single entry in the dropdown "Youth" might correspond to the two eligibilities "Children" and "Youth", for example).

Until they provide us these eligibilities, populate the dropdown with the current featured eligibilities from the landing page.

PR #757, opened 2019-05-16
#736: additional multi-level filters for search

quanhuynh:
first draft of the filtering functionality added to search results page

**Initial filters when clicked on a category/eligibility from the home page (box already checked)**
initial-filters2

**Closed filter bar (it opens on click)**
closed-filters-bar

**Multiple filters selected across eligibilities and categories**
multiple-refines

a few other stuff need to be figured out, particularly:
- not sure if this is optimal for mobile
- a few alignment issues with check box labels
- featured categories/eligibilities in filters need to be modified in the code. this means when MOHCD provides us with what filters they want, we'll need another PR.

i may need someones help with the front-end fixes, but this should be a start in satisfying MOHCD's filtering requirement


quanhuynh:
updated PR based on the eligibilities MOCHD wants to filter by. mapping from these terms to our backend can be found at https://github.com/ShelterTechSF/askdarcel-web/pull/757/files#diff-7ddc92e2d26f916f457b186186fa037aR15

PR #758, opened 2019-05-17
746 add action creators

tsyaeger:
This adds the boilerplate action creators based only on requests in the Postman collection. I'm wondering if this action creator pattern still looks okay or if it is maybe too verbose. The action types file is pretty long but as I understand it, the convention is to keep them housed in one file. Regardless, I'm happy to split it up if needed. Also, there's a chance that I misinterpreted "boilerplate", so just let me know if I should make any changes.


jmpainter:
This follows the configuration that you had presented before and I think it is clear and organized. I had used the term 'boilerplate' to mean just the basic layout without implementation goes beyond that now. This would be a good time to get the approval for this setup from Joe and Richard.

Issue #647, opened 2019-03-10
Fix remaining issues with improper open and closed status being displayed


jjfreund:
I see the following issues still (on 2:12, Sunday):

1) Certain services neither display as open or closed, such as Harm Reduction offered by Temenos Catholic Worker. This service is neither open nor closed, but should be open:
https://www.askdarcel.org/services/1822

2) Certain services are displayed as closed, but remain listed as open when the open filter is clicked -- different logic is used to determine whether they are open. In the following case:
https://www.askdarcel.org/services/94
 I believe it is because the hours are meant to indicate 24 hour availability (11:00pm - 10:59 pm)

<img width="1435" alt="Screen Shot 2019-03-10 at 2 13 15 PM" src="https://user-images.githubusercontent.com/637355/54091727-951a2f80-4340-11e9-93aa-1d5991cbf1b1.png">

cliffcrosland:
This appears to be working great now. @jjfreund Is this ok to close? Thanks!

cc: @Maxastuart 

![Screen Shot 2019-03-27 at 8 09 46 PM](https://user-images.githubusercontent.com/581562/55127132-54eed700-50cc-11e9-8285-ff35d0cc97c9.png)

![Screen Shot 2019-03-27 at 8 10 24 PM](https://user-images.githubusercontent.com/581562/55127147-689a3d80-50cc-11e9-827e-0da65c84ce40.png)

Maxastuart:
Might work, but need the hour-offset fixed to truly test and close

Maxastuart:
#418 #415 related

cliffcrosland:
Relevant PR in askdarcel-api from @jjfreund https://github.com/ShelterTechSF/askdarcel-api/pull/400

jjfreund:
@richardxia's refactor will take care of this -- but not in this milestone.

fredciaramaglia:
Other issues related to this bug:
1) On Friday at 11:15pm, resource Workforce Development Program displayed Open Now when the hours in the details say Friday 8:00am to 5:30pm.  Note: The Open Now button correctly filters out this resource.
2) On Friday at 11:33pm, resource Early Intervention And School-Based Program displayed Closed in 20 minutes when the details say Friday 9:00am to 7:00pm.  Note: The Open Now button correctly filters out this resource.  The same is true for Casa Corazón.
[...]

PR #759, opened 2019-05-18
Hoist edit schedule state up to the top

richardxia:
## Overview

This is hopefully the final of all the refactoring PRs that I'm doing before I can finally work on https://github.com/ShelterTechSF/askdarcel-web/issues/647.

With this PR, all of the React state related to schedules on the Edit page will be moved up to the top level, which ensures a single canonical representation of schedule-based state on the page. This, I believe, is a vast improvement over the old version of code, where the schedule state was represented at four different levels, and they all had to be kept in sync with each other, which made changing the representation of the schedule very difficult.

## What was changed

Just to recap, here's what the Component hierarchy on the Edit page looks like:

- OrganizationEditPage
  - EditSchedule (for the resource)
  - EditServices (wraps the collection of all services)
    - ProvidedService (represents just a single service)
      - EditSchedule (for the service)

EditSchedule itself has the following hierarchy:

- EditSchedule
  - EditScheduleDay (which represents each day of the week, and can actually contain more than one "ScheduleDay", where "ScheduleDay" in this context actually means a single time range. Lots of confusion arises from reusing the same word to sometimes mean just one time range and othertimes multiple time ranges)

Previously, EditServices had the state that represents the whole schedule, and I moved it up to the top level. The original execution order was confusing because the state was also already being duplicated back to the top-level component, but only open actually making an edit to a schedule. Now the state lives _only_ at the top level and it is created at component construction time, not only when a schedule is actually edited.

There were several event handler methods that would mutate the state of EditServices, and I have split them up such that each event handler lives in the component that actually uses them, but each component has an internal helper method that gets passed in that allows them to modify a slice of the parent component's state. Let me explain with an example.

The EditSchedule component is passed in a prop that lets it set the values of the entire schedule state, which is mostly represented with a mapping from day of week to a list of ScheduleDays. The EditSchedule component passes a setter method prop to EditSchedule day which is only allowed to modify the list of ScheduleDays for a particular day of the week, without allowing it to modify any of the other days of the week. EditScheduleDay itself now has a new subcomponent named TimeInputRow, which is only responsible for a single ScheduleDay, and EditScheduleDay passes a setter prop to TimeInputRow that only allows it to modify a single ScheduleDay.

This hierarchy allows us to pass a smaller set of state editing functions to each subcomponent while also allowing for the flexibility of subcomponents to construct more specific state editing methods. With a single `setScheduleDaysForDayOfWeek()`, a subcomponent can add/remove new time ranges, edit existing time ranges, etc.

I have hopefully done all this correctly, which was at least simplified via my earlier efforts of getting rid of state in the EditServices and ProvidedService components.

### How this was tested

Here's the checklist of scenarios that I have tested:

- Edit Resource
  - [x] Add a second ScheduleDay to a day of the week
  - [x] Remove a second ScheduleDay from a day of the week
  - [x] Remove all ScheduleDays from a single day of the week
  - [x] Edit a ScheduleDay's open time
  - [x] Edit a ScheduleDay's close time
  - [x] Change a day of the week with one ScheduleDay to be 24 hours
  - [x] Change a day of the week with multiple ScheduleDays to be 24 hours (all but one ScheduleDay should be deleted)
  - [x] Change a day of the week from being 24 hours to having a set time
- Edit Service
  - [x] Add a second ScheduleDay to a day of the week
  - [x] Remove a second ScheduleDay from a day of the week
  - [x] Remove all ScheduleDays from a single day of the week
  - [x] Edit a ScheduleDay's open time
  - [x] Edit a ScheduleDay's close time
  - [x] Change a day of the week with one ScheduleDay to be 24 hours
  - [x] Change a day of the week with multiple ScheduleDays to be 24 hours (all but one ScheduleDay should be deleted)
  - [x] Change a day of the week from being 24 hours to having a set time
  - [x] Set service schedule to inherit from resource
  - [x] Set service schedule to no longer inherit from resource
  - [x] Creating a new Service should allow for the above operations to still succeed
- Create New Resource
  - [x] Only sanity checked that the page isn't broken (because it actually was and I at least fixed it enough to pass the one failing TestCafe test)

### DIFF
--- a/app/components/edit/EditSchedule.jsx
+++ b/app/components/edit/EditSchedule.jsx
@@ -2,60 +2,9 @@ import React, { Component } from 'react';
 import _ from 'lodash';
 import { stringToTime } from '../../utils/index';
 import EditScheduleDay from './EditScheduleDay';
+import { buildScheduleDays } from '../../pages/OrganizationEditPage';
 
 import './EditSchedule.scss';
-
-function buildSchedule(schedule) {
-  const scheduleId = schedule ? schedule.id : null;
-  const currSchedule = {};
-  let finalSchedule = {};
-
-  const is24Hours = {
-    Monday: false,
-    Tuesday: false,
-    Wednesday: false,
-    Thursday: false,
-    Friday: false,
-    Saturday: false,
-    Sunday: false,
-  };
-
-  const tempSchedule = {
-    Monday: [{ opens_at: null, closes_at: null, scheduleId }],
-    Tuesday: [{ opens_at: null, closes_at: null, scheduleId }],
-    Wednesday: [{ opens_at: null, closes_at: null, scheduleId }],
-    Thursday: [{ opens_at: null, closes_at: null, scheduleId }],
-    Friday: [{ opens_at: null, closes_at: null, scheduleId }],
-    Saturday: [{ opens_at: null, closes_at: null, scheduleId }],
-    Sunday: [{ opens_at: null, closes_at: null, scheduleId }],
-  };
-
-  if (schedule) {
-    schedule.schedule_days.forEach(day => {
-      const currDay = day.day;
-      if (!is24Hours[currDay]) {
-        // Check to see if any of the hour pairs for the day
-        // indicate the resource/service is open 24 hours
-        // if there is a pair only have that in the day obj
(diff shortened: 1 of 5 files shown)


schroerbrian:
Looks solid. Approved

schroerbrian:
👍

PR #764, opened 2019-06-01
Add "Name" field to addresses (#305)

cliffcrosland:
Allow users to include a "Name" field when editing an address. This is
useful for resources/services with multiple locations.

See also: https://github.com/ShelterTechSF/askdarcel-api/pull/421

### DIFF
--- a/app/components/listing/ResourceInfos.jsx
+++ b/app/components/listing/ResourceInfos.jsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { Fragment } from 'react';
 import PropTypes from 'prop-types';
 import _ from 'lodash';
 import config from '../../config';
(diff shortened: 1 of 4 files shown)


cliffcrosland:
Screen Shot 2019-06-01 at 3 10 36 PM

Screen Shot 2019-06-01 at 3 10 51 PM

Screen Shot 2019-06-01 at 3 59 14 PM

app/components/listing/ResourceInfos.jsx
```diff
@@ -30,37 +30,53 @@ ResourceCategories.propTypes = {
...
   }
 
   return (
     <span>
-      {line1}
-      <br />
-      {line2}
+    {
+      lines.map((line, i) =>
+        <Fragment key={i}>
```
AnnaKalkanis: Generally, you want to avoid using index as a key. The key is usually a unique identifier to to the piece of data that element holds, whereas an index isn't necessarily linked to that piece of data forever. Usually you would use something like an id. In this case, it honestly isn't really going to matter - it's more of a problem with long lists. Just more of a general FYI since you mentioned you were newer to this stuff. :)

AnnaKalkanis: In react, you can return an array of elements and it knows how to render them. So, you could avoid mapping twice by doing something like:

```
function buildLocation(address) {
  if (!address) return;

  const fieldsOnEachLine = [
    ['name'],
    ['address_1', 'address_2'],
    ['city', 'state_province', 'postal_code'],
  ];

  return fieldsOnEachLine.map(fields => {
    const line = fields.map(field => address[field])
                   .filter(Boolean)
                   .join(', ');

    if (line.length > 0) {
      const key = fields.join('-');

      return (
        <Fragment key={key}>
          <span>{line}</span>
          <br />
        </Fragment>
      );
    }
  });
}
```

Again, given that this is a list of at most 3, it doesn't really matter and feel free to just leave it the way you have it now.

AnnaKalkanis:
Just some general react comments, but nothing critical. Looks good!

app/components/listing/ResourceInfos.jsx
cliffcrosland: Awesome! Thanks for the info about keys, and I love `filter(Boolean)`. Will make these changes

Issue #147, opened 2017-02-08
Display success or failure messages after editing and on admin panel


derekfidler:

Maxastuart:
@lexholden if you’re still working on this, can you please add a brief description/update? Otherwise feel free to unassign yourself

lexholden:
I general we will need to totally rewrite the edit page at some point, this is required but would be part of that epic.

PR #766, opened 2019-06-03
Display success/failure message after save (#147)

cliffcrosland:
When editing a resource or service, display a pop-up message indicating
success or failure.

Introduce new `PopUpMessage` system. This can be used by any component
to display a temporary pop-up message.


cliffcrosland:
## Screenshot 1: Mobile
Screen Shot 2019-06-02 at 7 08 31 PM

## Screenshot 2: Desktop success
Screen Shot 2019-06-02 at 7 15 10 PM

## Screenshot 3: Desktop error
Screen Shot 2019-06-02 at 7 17 01 PM

## Video 1: Desktop
[pop-up-message.mov.zip](https://github.com/ShelterTechSF/askdarcel-web/files/3245606/pop-up-message.mov.zip)

cliffcrosland:
Whoops. The linter found some issues. Will push up another commit shortly.

cliffcrosland:
Cool. Linter is happy now.

app/components/ui/PopUpMessage.scss
```diff
@@ -0,0 +1,41 @@
+.pop-up-message {
+  display: table;
```
AnnaKalkanis: Do you know if this is necessary? I just never use `display: table`, so I'm curious what it is doing in this context.

app/components/ui/PopUpMessage.scss
```diff
@@ -0,0 +1,41 @@
...
+  position: fixed;
+  margin: auto;
+  top: 100px;
+  bottom: 0;
+  left: 0;
+  right: 0;
+  color: white;
+  height: 50px;
+  width: 300px;
```
AnnaKalkanis: Maybe a min-width makes more sense here? I'm just thinking if the message that is passed in is long, it is going to overflow this 300px.

app/components/ui/PopUpMessage.scss
```diff
@@ -0,0 +1,41 @@
...
+  text-align: center;
+  border-radius: 4px;
+  padding-top: 4px;
+  padding-bottom: 4px;
+  transition: opacity 0.5s;
+  -webkit-transition: opacity 0.5s; /* Safari */
+
+  &.success {
+    background-color: #5BBA81;
```
AnnaKalkanis: We have some css variables defined in `_helpers.scss`. If possible, it would be great to add these colors as variables in there and then we can just reference the variable here. I noticed that we also use these colors in `_layout.scss` so we could replace the hardcoded hex values with these new variables as well. That way, if we ever want to change the colors, they are only defined in one place.

AnnaKalkanis:
Nice! Would be great to get the colors in the css into variables, but otherwise looks great!

app/components/ui/PopUpMessage.scss
cliffcrosland: Haha, yes. Would love your help on how to do this properly. I was trying to center text vertically in the pop-up message, and after a ton of experimenting and failing, I found a StackOverflow comment about using `display: table` to perform vertical centering.

I'll bet there is something much better, like `display: flex` or similar.

Would you recommend `display: flex`? Or is there an even better approach?

In the meantime, I'll try to find a better alternative and push another attempt

app/components/ui/PopUpMessage.scss
cliffcrosland: Great idea! I'll go ahead and do that.

app/components/ui/PopUpMessage.scss
cliffcrosland: Cool, min-width looks a lot better with large messages. Will use that.

app/components/ui/PopUpMessage.scss
AnnaKalkanis: Ahh gotcha! Ya, display: flex with align-items: center should do the trick! (It might be justify-content, I always mix these up...)

cliffcrosland:
Quick update. Was able to change `display: table` to `display: flex` and simplify the style and layout a lot.

Unfortunately, changing `width` to `min-width` caused the pop-up to fill the full width of the screen. Tried replacing `width` with `min-width` and `max-width`, but it just used always used `max-width`. Tried using `flex-basis` as well, but it seemed to have no effect. Hence, I kept the `width` property.

PR #783, opened 2019-06-23
682 service net attribution

schroerbrian:


app/components/listing/ServiceAttribution.jsx
```diff
@@ -0,0 +1,36 @@
...
+  isVerified() {
+    const { attribution, status } = this.props;
+    return attribution === 'service_net' && status === 'approved';
+  }
+
+  render() {
+    const benetechLink = 'https://medium.com/@Shelter_Tech/sheltertech-is-participating-'
+      + 'in-the-benetech-service-net-pilot-in-the-san-francisco-bay-area-b28645d3dee6';
+
+    return this.isVerified() ? (
```
jjfreund: Is this the right name for the method? We really want to include ByServiceNet in the name or something similar right? There's a different field is_verified which we use sounds ambiguous to this method.

schroerbrian: yes, that is a very good point. i think this was leftover confusion as to whether i was working on HAP or ServiceNet or both.  i have updated the method name as well as some markup class names and img name for greater clarity.

app/components/listing/ServiceAttribution.jsx
```diff
@@ -0,0 +1,36 @@
...
+  isServiceNetAttribution() {
+    const { attribution, status } = this.props;
+    return attribution === 'service_net' && status === 'approved';
+  }
+
+  render() {
+    const benetechLink = 'https://medium.com/@Shelter_Tech/sheltertech-is-participating-'
+      + 'in-the-benetech-service-net-pilot-in-the-san-francisco-bay-area-b28645d3dee6';
+
+    return this.isServiceNetAttribution() ? (
```
richardxia: Nit: I think usually we prefer to use the `predicate && <JSX stuff>` pattern over `predicate ? <JSX Stuff> : null` when the "else" case should be to return nothing.

Hopefully you don't mind, but I'll make that minor tweak before merging this in.

PR #785, opened 2019-06-27
[784] Fixed incorrect font sizes on landing page

drcaramelsyrup:
It looks like this was caused by importing _helpers.scss into _layouts.scss in order to get _colors for the success / failure messages and open / close times. This also caused other unintended effects including underlines in the h2 tags. I decided to take the _colors out into a separate file.


app/styles/st-base/_layout.scss
```diff
@@ -1,4 +1,4 @@
-@import '../utils/_helpers.scss';
```
schroerbrian: i think we will still need to import the ` _helpers.scss` file as the _layout.scss file still references some vars from _layout.scss

drcaramelsyrup: That's odd. I see the dependent variables you mention, but prior to a change a week ago, this file wasn't actually included in this .scss file at all - even though it contained these variables.

I'll try to extract those variables from `_helpers.scss` as well.

drcaramelsyrup: I've moved the section containing those dependent variables from `_helpers` to a different file. I could be even more minimal about this change and only move the variable declarations.

schroerbrian: i see. that's strange. i'm perplexed by the .scss tree of dependencies sometimes.

Issue #638, opened 2019-02-24
Edit service, add/remove service bug


Maxastuart:
When removing a newly added unsaved service from an edit page, the service should be removed from the display of the page. This bug occurs when someone editing a service accidentally clicks “add a service” and then immediately clicks “remove service,” - without having saved anything, the blank or incorrect false service form fields stay displayed.

Maxastuart:
Update:
 - when you click "remove service" a confirm dialog pops up that says "are you sure you want to DEACTIVATE this RESOURCE" (wrong wording) 
 - Additionally, then clicking yes in the dialog does nothing but close the dialog.

PR #786, opened 2019-06-28
638 - fix edit service bug

quanhuynh:
Look at #638 for more details
Fix: now removing a service that was just added (but hasn't saved) makes that service disappear. Also reworded the confirmation prompt to refer to service instead of resource 

Screen Shot 2019-06-28 at 3 58 05 PM

### DIFF
--- a/app/pages/OrganizationEditPage.jsx
+++ b/app/pages/OrganizationEditPage.jsx
@@ -620,17 +620,26 @@ class OrganizationEditPage extends React.Component {
 
   handleDeactivation(type, id) {
     const { router } = this.props;
+    let confirmMessage = null;
+    let path = null;
+    if (type === 'resource') {
+      confirmMessage = 'Are you sure you want to deactive this resource?';
+      path = `/api/resources/${id}`;
+    } else if (type === 'service') {
+      confirmMessage = 'Are you sure you want to remove this service?';
+      path = `/api/services/${id}`;
+    }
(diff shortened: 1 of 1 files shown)


app/pages/OrganizationEditPage.jsx
```diff
@@ -620,28 +620,38 @@ class OrganizationEditPage extends React.Component {
...
-            // Service successfully deactivated. Mark deactivated in local state.
-            this.setState(state => {
-              state.deactivatedServiceIds.add(id);
-              return state;
-            });
-          }
+    if (confirm(confirmMessage) === true) {
+      if (id < 0) {
+        this.setState(state => {
+          state.deactivatedServiceIds.add(id);
```
richardxia: You're not supposed to directly mutate the values in a React `state` object, since React is unable to detect when those changes happen. Instead, React only wants you to ever modify state by calling the `setState()` method. See https://reactjs.org/docs/state-and-lifecycle.html#do-not-modify-state-directly

The subtle distinction here is that although you are calling `setState()` and returning the updated state, the way you are doing it involves directly changing the previous object.

The correct way of doing what you're trying to do here is to create a new array that contains the old `state.deactivatedServiceIds` array plus a new element at the end. Let me modify your code here to show what it should look like.

quanhuynh: Thanks @richardxia !

PR #792, opened 2019-06-30
Karla 712 basic needs block

Raawr:


app/pages/legal/PrivacyPolicy.jsx
```diff
@@ -6,7 +6,7 @@ export class PrivacyPolicyPage extends React.Component {
   static get content() {
     return `
 # Privacy Policy
-We do everything within our power to protect your privacy given the sensitive nature of many of the services listed on AskDarcel. 
+We do everything within our power to protect your privacy given the sensitive nature of many of the services listed on SF Service Guide.
```
jjfreund: Same comment as before -- it seems like changes unrelated to what you were trying to do (pare down the basic need block) made it into this PR. Can you remove those and add them to a separate PR? That way we can review and approve them one at a time.

jjfreund: Oh one more thing -- Travis is failing this commit. It looks like you based your commit on an old version of master, and an incorrect version of node is being used. I'd suggest rebasing to the current master.

richardxia: I just cherry-picked and rebased the basic block change on top of master, force-pushing to this branch, so this PR should be good to merge now.

Issue #374, opened 2018-01-11
Open Now/Closed Now incorrect for resources that close after midnight


richardxia:
On staging I edited a resource such that it is open on Wednesday from 11PM to 11AM (which I assume implicitly means 11AM Thursday). It is currently 10:30AM on a Thursday, but the summary at the top of the resource page says "Closed Now":

<img width="800" alt="screen shot 2018-01-11 at 10 32 10 am" src="https://user-images.githubusercontent.com/1002748/34840787-eb0f2004-f6ba-11e7-85dc-ea4ff8992748.png">
<img width="785" alt="screen shot 2018-01-11 at 10 32 17 am" src="https://user-images.githubusercontent.com/1002748/34840788-ec8f63f8-f6ba-11e7-9c4b-b2ff107b0151.png">

Note that on Resources listing page, the correct Open Now/Closed Now summary is displayed:

<img width="658" alt="screen shot 2018-01-11 at 10 34 10 am" src="https://user-images.githubusercontent.com/1002748/34840831-137ce378-f6bb-11e7-8b8b-97a28f33e7ee.png">

Whatever logic is used for that page should get refactored into a common function that the individual Resource page can reuse.

PR #793, opened 2019-07-01
Begin cleaning up open/close times 

richardxia:
This is the beginning of actually attempting to fix up the open/close times across the site. What I've done here is define a few data structures that are meant to centralize the logic for handling open/close times and begin to migrate the codebase to use them. The data structures all live in `app/utils/RecurringSchedule/models.js` and are named:

- `RecurringTime`, which represents the 3-tuple of (day of week, hour, minute)
- `RecurringInterval`, which represents the 2-tuple of (open time, close time)
- `RecurringSchedule`, which represents an array of `RecurringInterval`s.

I've added a couple new functions to the `app/utils/DataService.js` module — `getResource()` and `getService()` — which makes API requests and, before returning them to the caller, parses the results into the above data structures. In order to make the changes backwards compatible, instead of overwriting the existing `schedule` properties on the resource and service, I set the parsed schedule into a property named `parsedSchedule`.

My goal is to gradually migrate the whole app to use these new data structures and to delete all the old time handling code, since there were too many different implementations. Note that this new set of data structures does not rely on the builtin `Date` object or any other third party datetime library; this is because we don't really have datetime objects here, as our times represent recurring times each week. It was simpler to just represent them as plain data structures and do modular arithmetic on them than to attempt to force them into `Date` objects, and this also has the benefit of not having to deal with time zones.

The one component that I've updated to use these new data structures is the `TableOfOpeningTimes`, which is used on the "show" page for both Resources and Services.

Note that I'm not sure if this PR is good enough to merge just yet. I think I want to double check all the error handling in cases where the data coming back from the API doesn't match what I expect it to be.

### DIFF
--- a/app/pages/ServiceListingPage.jsx
+++ b/app/pages/ServiceListingPage.jsx
@@ -24,14 +24,14 @@ import HAPBadge from 'components/listing/HAPBadge';
 import { isSFServiceGuideSite } from '../utils/whitelabel';
 
 // TODO This should be serviceAtLocation
-const getServiceLocations = (service, resource, schedule) => (resource.address
+const getServiceLocations = (service, resource, recurringSchedule) => (resource.address
   ? [resource.address].map(address => ({
     id: address.id,
     address,
     name: service.name,
-    schedule: schedule && schedule.schedule_days.length ? schedule : resource.schedule,
+    recurringSchedule,
     // Just to make it clear this is inherited from the resource
-    inherited: !schedule && resource.schedule,
+    inherited: !recurringSchedule && resource.schedule,
   }))
   : []);
 
@@ -73,9 +73,9 @@ class ServicePage extends React.Component {
     const { activeService: service } = this.props;
     if (!service) { return <Loader />; }
 
-    const { resource, program, schedule } = service;
+    const { resource, program, recurringSchedule } = service;
     const details = this.generateDetailsRows();
-    const locations = getServiceLocations(service, resource, schedule);
+    const locations = getServiceLocations(service, resource, recurringSchedule);
     return (
       <div>
         <Helmet>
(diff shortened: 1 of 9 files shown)


Maxastuart:
REVIEWER: please cross-check with #374 (Open Now/Closed Now incorrect for resources that close after midnight)

app/pages/ServiceListingPage.jsx
```diff
@@ -28,7 +28,7 @@ const getServiceLocations = (service, resource, schedule) => (resource.address
     id: address.id,
     address,
     name: service.name,
-    schedule: schedule && schedule.schedule_days.length ? schedule : resource.schedule,
+    schedule,
```
shadycat: is this supposed to be set to parsed schedule?

richardxia:
Made a few changes to this PR after chatting with @shadycat earlier today:

- Renamed `WeeklyTime` to `RecurringTime`, `ScheduleDay` to `RecurringInterval`, and `Schedule` to `RecurringSchedule` to better emphasize that these all represent recurring, weekly schedules. Despite our conversation IRL today, I stuck with using the word "time" in `RecurringTime` because I felt that "point" would have been more confusing to use.
- Replaced all names of `parsedSchedule` with `recurringSchedule`, which I think makes it clearer that it holds a `RecurringSchedule` instance
- Removed all methods on the above data structures that are not yet in use. This includes all of the overlap detection logic. I'll bring these back in a future PR, when they are actually being used.

I very lightly retested this stuff locally by visiting both a Resource show page and a Service show page, making sure that the `<TableOfOpenTimes>` still renders properly.

richardxia:
I'm going to go ahead and merge this now, assuming that we'll find any bugs with it in QA.

PR #794, opened 2019-07-01
787 - disable edit and add buttons

quanhuynh:
commented out the add and edit buttons (they do not show up on inspect element)
let me know if there is any other parts or links i missed
Screen Shot 2019-06-30 at 6 47 05 PM
Screen Shot 2019-06-30 at 6 46 56 PM


quanhuynh:
CI failing because tests still expect the edit functionality. should we disable it in the test suite for now?

PR #795, opened 2019-07-01
Fix lint on listing pages.

richardxia:
As the title says. I also made a few minor changes to the ResourceInfos file to remove the `if (!foo) return null;` guards and pushed that up higher in the component tree, since I'd argue that it's up to the parent to handle the null cases.

### DIFF
--- a/app/components/listing/ResourceInfos.jsx
+++ b/app/components/listing/ResourceInfos.jsx
@@ -1,26 +1,22 @@
 import React, { Fragment } from 'react';
 import PropTypes from 'prop-types';
 import _ from 'lodash';
-import config from '../../config';
-import { RelativeOpeningTime } from './RelativeOpeningTime';
 
-function Cat(props) {
-  return <p>{props.category}</p>;
-}
+const Category = ({ category }) => <p>{category}</p>;
 
-Cat.propTypes = {
+Category.propTypes = {
   category: PropTypes.string.isRequired,
 };
 
 
-function ResourceCategories(props) {
-  if (props.categories.length) {
-    const categories = _.uniqBy(props.categories, 'id');
-    const cats = categories.map(cat => <Cat category={cat.name} key={cat.id} />);
-    return <span className="categories">{cats}</span>;
-  }
-  return null;
-}
+const ResourceCategories = ({ categories }) => {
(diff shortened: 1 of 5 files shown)


app/components/listing/ResourceInfos.jsx
```diff
@@ -131,96 +104,30 @@ PhoneNumber.propTypes = {
 };
 
 
-function ExternalLink(props) {
-  return <a href={props.to} target="_blank" rel="noopener noreferrer">{props.children}</a>;
-}
+const ExternalLink = ({ children, to }) => <a href={to} target="_blank" rel="noopener noreferrer">{children}</a>;
```
drcaramelsyrup: Nitpick: this line seems a good bit longer than the rest. I feel like it might be good to give this one the paren-wrapped multi-line treatment for readability.

app/components/listing/ResourceInfos.jsx
```diff
@@ -131,96 +104,30 @@ PhoneNumber.propTypes = {
...
-  if (!props.email) {
-    return null;
-  }
-  return (
-    <span className="email">
-      <ExternalLink to={`mailto:${props.email}`}>{props.email}</ExternalLink>
-    </span>
-  );
-}
+export const Email = ({ email }) => (
```
drcaramelsyrup: Why is it preferable to export these components with their declaration? Is there a linter complaint / style reason why we don't want the export block at the bottom of the file? Personally I thought it made it clearer which components this file offers.

app/components/listing/ResourceInfos.jsx
richardxia: Agreed, I can make this change.

app/components/listing/ResourceInfos.jsx
richardxia: There isn't a linter rule that enforces this; my personal style preference was to put the export declarations in the definitions of the components, since it makes it easier to tell whether or not a component is being exported while you're reading the definition of the component itself. However, the tradeoff is that you don't have a single place that lists all of the exported components.

I don't have a strong opinion either way, and I think this particular file is a bit of an exception because it defines so many components, some of which are exported and some of which are not, so I can revert back to the old style of declaring the exports at the bottom.

PR #797, opened 2019-07-01
fixed about us link

daniel-owens:


package-lock.json
```diff
@@ -10066,12 +10066,14 @@
         "balanced-match": {
           "version": "1.0.0",
           "bundled": true,
-          "dev": true
+          "dev": true,
+          "optional": true
```
richardxia: Looks like you committed some unnecessary changes to `package-lock.json`. I'm going to amend this commit to revert these changes, but the change otherwise LGTM!

PR #803, opened 2019-07-14
796 discover panel heading link to see more

shadycat:
This resolves: Make each Discover panel heading link to see more search results.

### DIFF
--- a/app/components/ui/LandingPageResourceBlock.jsx
+++ b/app/components/ui/LandingPageResourceBlock.jsx
@@ -4,7 +4,8 @@ import PropTypes from 'prop-types';
 const HOST_QUERY = '/search?query=';
 
 class LandingPageResourceBlock extends Component {
-  render() {
+    render() {
+    const url = `/search?refinementList[categories][0]=${encodeURIComponent(this.props.config.TITLE.BLUE_WORD)}`
     return (
       <div className="landing-page-resource-block">
         <div className="landing-page-resource-block__resources">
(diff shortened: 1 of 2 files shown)


app/components/ui/LandingPageResourceBlock.jsx
```diff
@@ -4,7 +4,8 @@ import PropTypes from 'prop-types';
 const HOST_QUERY = '/search?query=';
 
 class LandingPageResourceBlock extends Component {
-  render() {
+    render() {
+    const url = `/search?refinementList[categories][0]=${encodeURIComponent(this.props.config.TITLE.BLUE_WORD)}`
```
richardxia: I feel like I see this replicated in several different pages, and I feel like it'd be useful to consolidate it into a single helper function that knows the URLs of various search pages.

app/components/ui/LandingPageResourceBlock.jsx
```diff
@@ -14,11 +15,16 @@ class LandingPageResourceBlock extends Component {
               {this.props.config.TITLE.BEFORE_BLUE_WORD}
               <span className="blue-word">
                 {' '}
+                <a href={url}>
                 {this.props.config.TITLE.BLUE_WORD}
```
richardxia: Nit: You should indent the children of an HTML tag so that the visual indentation makes it easier to see which tags are inside of which other tags.

derekfidler:
I believe this has already been completed satisfactorily so I'm going to close this ticket

PR #805, opened 2019-07-17
[714] Casey landing page block and carousel

drcaramelsyrup:
Created landing page block for the Casey Youth Homelessness guide. The Figma spec outlines how this block (and the related related changes here) should look: https://www.figma.com/file/H0kD6ZZPhPR2RMyNJHWWeTtI/Web-Client?node-id=0%3A2
- Created a carousel component with navigation buttons to display multiple cards.
- Added links to the Legal, Basic Needs & Shelter, and Youth Homelessness blocks according to the Figma spec.
- Cleaned up linter errors on the Landing Page Resource Block.


app/components/ui/Carousel.jsx
```diff
@@ -0,0 +1,92 @@
...
+    const { children, numSlots } = this.props;
+    const slotWidth = 100 / numSlots;
+
+    return (
+      <CarouselOuter>
+        <CarouselContainer>
+          <CarouselSlider slotWidth={slotWidth} activeIndex={activeIndex}>
+            {
+              children.map(child => ((
+                <CarouselSlot width={slotWidth}>
```
richardxia: React complains that the `key` property is not set on any of these.

app/components/ui/Carousel.jsx
```diff
@@ -0,0 +1,92 @@
...
+    };
+
+    const advanceLeft = () => {
+      const { activeIndex } = this.state;
+      this.setState({ activeIndex: Math.max(0, activeIndex - 1) });
+    };
+
+    const { activeIndex } = this.state;
+    const { children, numSlots } = this.props;
+    const slotWidth = 100 / numSlots;
```
richardxia: Something's a bit odd with this when viewed on mobile. I can't actually scroll to the right far enough to see all of the items:

Screen Shot 2019-07-21 at 7 44 00 PM

app/components/ui/Carousel.jsx
```diff
@@ -0,0 +1,92 @@
...
+const CarouselSlider = styled.div`
+  position: relative;
+  display: flex;
+  flex-wrap: nowrap;
+  transition: transform 1s ease;
+  transform: translateX(${props => -props.activeIndex * props.slotWidth}%);
+`;
+
+const CarouselNavButton = ({ dir, onClick }) => (
+  <button type="button" className={`carousel-nav btn-${dir}`} onClick={onClick} />
```
richardxia: Since you've built the rest of the components on this page with Styled Components, maybe it makes sense to translate these classes to Styled Components as well?

drcaramelsyrup: I'll do that. That question about Styled Components - I was hesitant to introduce them, but I didn't see an easy way to implement a scalable Carousel without the help of dynamically-generated CSS. If there's an idiomatic way to do this in Sass I should read up on, I'd be interested!

richardxia: I'm not opposed to using Styled Components, since I think it makes it more difficult to create overly "global" styles, but in this particular case. I think you may just be able to use (S)CSS for normal styling, and then use inline styles via the `style` prop for anything that must be dynamic.

Also, have you considered using an existing library for creating the carousel and tweaking the styles to match our design?

drcaramelsyrup: Cool, was able to do that. It seemed like existing libraries that matched what we needed were pretty scarce. I think one that could show multiple items at once also had a couple of extra features that would be pretty expensive (e.g. triggering rerenders via a timer to cycle through the slides), so I figured it wouldn't be too bad to try to just do it on my own. Turns out it took a bit longer to get the carousel to work with the mobile view!

PR #808, opened 2019-07-22
Fix lint on app/components/layouts

richardxia:
I fixed up the lint errors on files under app/components/layouts as well as got rid of a bunch of unused, commented out references to RelativeOpeningTime.


richardxia:
I'm going to assume that this is OK to merge since it's a pure refactor and no new behavior is expected.

Issue #771, opened 2019-06-13
Map icon in wrong location after search and moves when zooming in or out


fredciaramaglia:
From the AskDarcel home page click on the Legal category.  The first resource is Legal Advocacy, 3543 18th Street.  See the first screenshot and notice that icon 1 is on Valencia near 20th, about 2-3 blocks from the correct location.

![image.png](https://images.zenhubusercontent.com/5cdcbb3d22376b7717f71776/497d8d18-c114-4dc3-9b13-5ac58be47c40)

Now click the '-' zoom out button and notice that icon 2 has moved another few blocks, close to 22nd and Mission.

![image.png](https://images.zenhubusercontent.com/5cdcbb3d22376b7717f71776/4b63f8d5-f291-4320-bbf0-adcd555cc0e0)

If you keep zooming in or out you will see all of the icons move on the map although the icons seem to maintain their relative positions.

These two bugs may be related since  when you zoom in or out the map layer and the icon layer are moving relative to each other almost as if they had different center points, and when you zoom in close enough icon 1 is in the correct location.

fredciaramaglia:
Related to Issue:
https://github.com/ShelterTechSF/askdarcel-web/issues/251

May be duplicate.

fredciaramaglia:
Verified fixed on Staging

PR #815, opened 2019-07-27
771 Fix bug where map icons were drifting in and out on zoom. Update …

schroerbrian:
…google-map-react which also fixes weird teleporting behavior on zoom.

This specific CSS fix was recommended in the threads of the google-map-react package that we're using: https://github.com/google-map-react/google-map-react/issues/470

I also updated the google-map-react package. We were a whole version behind; this fixes a separate, minor but annoying issue: https://github.com/google-map-react/google-map-react/issues/636 It's not strictly necessary for #771, but I thought it'd be nice to get the update in. 

~P.S. What's our policy on committing package-lock.json? I wasn't sure, so didn't include it in this PR. I can create a separate commit and push if we want to include it.~

P.P.S. Travis has helpfully informed me that package-lock.json must be updated as well

### DIFF
--- a/app/components/search/SearchMap.jsx
+++ b/app/components/search/SearchMap.jsx
@@ -55,7 +55,15 @@ const SearchMap = ({ hits, userLocation }) => {
     return null;
   }
 
-  const markers = hits.map((hit, index) => <CustomMarker lat={hit._geoloc ? hit._geoloc.lat : 0} lng={hit._geoloc ? hit._geoloc.lng : 0} key={hit.objectID} index={index} />);
+  const markers = hits.map(function(hit, index) {
+    return <CustomMarker
+      lat={hit._geoloc ? hit._geoloc.lat : 0}
+      lng={hit._geoloc ? hit._geoloc.lng : 0}
+      key={hit.objectID}
+      index={index} />
+  }
+
+  );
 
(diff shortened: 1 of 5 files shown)


app/components/search/SearchMap.jsx
```diff
@@ -55,7 +55,15 @@ const SearchMap = ({ hits, userLocation }) => {
     return null;
   }
 
-  const markers = hits.map((hit, index) => <CustomMarker lat={hit._geoloc ? hit._geoloc.lat : 0} lng={hit._geoloc ? hit._geoloc.lng : 0} key={hit.objectID} index={index} />);
+  const markers = hits.map(function(hit, index) {
```
richardxia: In terms of code style, we prefer using arrow functions in many cases, but especially when passing anonymous functions to functional programming functions like `map()`. If you want to break it across multiple lines, you can put parentheses around the arrow function body, like so:

```js
  const markers = hits.map((hit, index) => (
    <CustomMarker
      lat={hit._geoloc ? hit._geoloc.lat : 0}
      lng={hit._geoloc ? hit._geoloc.lng : 0}
      key={hit.objectID}
      index={index} />
  ));
```

richardxia: I'm going to go ahead and make the minor formatting changes here and push it up to this branch, and then I can merge when Travis completes.

PR #816, opened 2019-07-28
809 gray out spurious refinements

schroerbrian:
the categoriesRefinementList and the eligiblitiesRefinementList can pretty easily be collapsed into one file/component, but i just wanted to complete this task and perhaps refactor the two at a later time

### DIFF
--- a/app/components/search/CategoriesRefinementList.jsx
+++ b/app/components/search/CategoriesRefinementList.jsx
@@ -67,6 +67,11 @@ class CategoriesRefinementList extends Component {
     }
   }
 
+  refinementHasResults(key) {
+    const { items } = this.props;
+    return items.find(item => item.label === key);
+  }
+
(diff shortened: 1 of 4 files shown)


app/components/search/CategoriesRefinementList.jsx
```diff
@@ -67,29 +67,42 @@ class CategoriesRefinementList extends Component {
...
-                />
-                {key}
-              </label>
-            </li>
-          ))}
+            return (
+              <li
+                key={key}
+                className={`refine-li ${isChecked[key] ? 'active' : ''
+                }${!refinementHasResults ? 'refine-li-disabled' : ''}`}
```
richardxia: Minor: I'd suggest breaking out this logic for computing the CSS class names into pure JS, since that might make it easier to read.

Issue #755, opened 2019-05-16
Move save button or make it floating & fix bug on services list


Maxastuart:
Move save button (or make it floating) on edit page to make it easier to save when a resource has lots of services. (see below comments on the bug)

Maxastuart:
(Notes from the duplicate Joe made: "Save" button should always be visible on the left-hand side of the edit screen. Currently, the "Save" button on the edit screen is anchored to the bottom -- the user must scroll all the way down to access it. The "Save" button should instead always be visible.)

quanhuynh:
Found another issue while working on this one:
The sidebar populates existing services twice, with the duplicates not shown. This makes the Services Section longer than it should be, and when a new service is added, there's a huge gap between existing services and the new service. see screenshot below:

![Screen Shot 2019-07-28 at 3 52 16 PM](https://user-images.githubusercontent.com/8869737/62013959-b04ed580-b14f-11e9-8f56-4c4a705d0dd5.png)

I am going to make a fix for this in the same PR

Fyi @Maxastuart

PR #820, opened 2019-07-28
755 - Edit Resource sidebar fixes

quanhuynh:
two fixes:
1. make edit sidebar float: original issue only requests for save button, but floating the whole sidebar may enhance navigation, especially with a lot of services (it's also easier to implement this for the whole sidebar unit)
2. remove invisible duplicate listings of existing services (refer to issue #755 for details)

sweet gif showing how it looks

a2b8563f2c4ba2399a63393124cd6384


app/components/edit/EditSidebar.jsx
```diff
@@ -69,6 +69,12 @@ const EditSidebar = ({
       </button>,
     );
   }
+  // populate existing services so they show up on the sidebar
+  const allServices = newServices;
+  for (let i = 0; i < resource.services.length; i++) {
+    const resourceService = resource.services[i];
+    allServices[resourceService.id].name = resourceService.name;
```
richardxia: What you're doing here isn't safe, even if it happens to work now. You're mutating the values of an object that is being passed into this function (`newServices`), which causes the values to change from underneath the caller. This could cause unintentional results if the caller assumed that the values passed in as arguments would not change. You really should be making a copy of each object and modifying that copy rather than mutating the array that you passed in.

On a related note, the C-style for loop you've written is unidiomatic JavaScript. Instead of creating a loop counter, defining a loop invariant, and incrementing the loop counter, it's more encouraged these days to use the functions that allow you to directly iterate over the elements of an array.

You can address both of these points I've mentioned by using the functional programming methods `.reduce()` and `.forEach()`. This would look like the following:

```js
// Do a 2-level-deep clone of the newServices object
const allServices = newServices.entries().reduce(
  (acc, [id, service]) => ({...acc, [id]: {...service}}),
  {},
);
resource.services.forEach(service => {
  allServices[service.id].name = service.name;
});
```

app/components/edit/EditSidebar.scss
```diff
@@ -7,8 +7,12 @@
 	width: calc-em(250px);
 	background-color: $color-grey1;
   display: flex;
-	justify-content: space-between;
 	flex-direction: column;
+  position: sticky;
+  position: -webkit-sticky;
```
richardxia: You probably shouldn't be explicitly writing the vendor prefix properties here, since we really ought to be running a CSS postprocessor that outputs the complete set of vendor prefixes. If we don't have one set up yet, can we install and configure [Autoprefixer](https://github.com/postcss/autoprefixer) to add these prefixes?

quanhuynh: @richardxia I can set up the autoprefixer in a separate issue/PR? to keep the scope of this PR clear

app/components/edit/EditSidebar.jsx
quanhuynh: @richardxia thanks for the notes! javascript is new territory for me :p

quanhuynh:
@richardxia made the changes and tests passed :)

PR #829, opened 2019-08-08
774 - fix checkbox bugs & bunch of other things

quanhuynh:
Fixing recent bug - some checkboxes not checkable. The issue was that algolia by default only gives 10 refinement values, that's why most of the filters couldn't be checked. Made a bunch of other changes to improve the filtering function as well

Changes:
1. increased the limit of refinement values returned by algolia. now all checkboxes are checkable!
2. @schroerbrian i changed your **refinementHasResults()** logic. originally this looks only at the key (Disability) to determine if a checkbox can be refined. but this key label is only a key that maps to the actual eligibilities to be refined (e.g. Learning Disabilities, Intellectual Disabilities, etc.) so i changed it to check that for each key, at least one eligibility has results
3. changed the URLs that the eligibility card on home page leads to - now the URLs includes all underlying eligibilities (e.g originally Families only leads to Families eligibility, now it includes both Families and Families with Babies; same with other multi-eligibility cards)
4. new refinement logic - if a checkbox is not checked, checking it will refine all underlying eligibilities; if a checkbox is already checked, checking it will un-refine all underlying eligibilities
5. separated the mapping structure out onto a new util .js file

anddd here's a gif

612f857494ed683ce32213d9b933f0bc


app/components/ui/LandingPageEligibilityBlock.jsx
```diff
@@ -1,8 +1,19 @@
...
 class LandingPageEligibilityBlock extends Component {
+
+  buildBlockUrl(eligibility) {
+    let url = '/search?'
+    const eligibilityElements = eligibilitiesMapping[eligibility];
+    eligibilityElements.forEach((element, index) => {
+      url += `refinementList[eligibilities][${index}]=${encodeURIComponent(element)}&`;
+    });
+    return url;
+  }
```
richardxia: I think it'd be better to use a library for this, since it's more likely that they'll handle the edge cases when it comes to query parameter escaping. [`URLSearchParams`](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams) is built in to modern browsers, although if we want to maintain IE11 support, we'll probably want to install a [polyfill](https://www.npmjs.com/package/url-search-params-polyfill).

app/components/search/CategoriesRefinementList.jsx
```diff
@@ -13,68 +14,61 @@ class CategoriesRefinementList extends Component {
...
-        const val = this.categoriesMapping[key][i_1];
-        if (currentRefinement.includes(val)) {
-          atLeastOneRefined = true;
-          break;
-        }
-      }
-      checks[key] = atLeastOneRefined;
-    }
+    mapKeys.forEach(key => {
+      checks[key] = this.keyHasAtLeastOneRefined(key);
```
richardxia: I see that this was already like this before, but `checks` is declared as an array, but for some reason we are treating it as a key-value mapping. It's probably clearer to declare it as `checks = {}`.

app/components/search/Filtering.jsx
```diff
@@ -74,8 +74,8 @@ class Filtering extends Component {
               Open now
             </button>
             <div className={"custom-refinement " + (filtersActive ? 'active' : '')}>
-              <EligibilitiesRefinementList attribute="eligibilities" />
-              <CategoriesRefinementList attribute="categories" />
+              <EligibilitiesRefinementList attribute="eligibilities" limit={100} />
+              <CategoriesRefinementList attribute="categories" limit={110} />
```
richardxia: Could you add a comment describing why you have explicitly set the limits? This will help future readers of the code understand why the code is the way it is.

quanhuynh:
> It looks like there is a good amount of duplication of functionality between CategoriesRefinementList and EligibilitiesRefinementList. Is there a way to capture the common functionality into a shared component/module?

thought the same thing when i was making edits. i just haven't thought about if i wanted to include that work in this PR. since it doesn't seem like much work i can include it as well

quanhuynh:
@richardxia CI says i'm getting a dependency cycle error on pages that i didn't edit. do you know why this might have shown up now?

richardxia:
> @richardxia CI says i'm getting a dependency cycle error on pages that i didn't edit (EditServices, ProvidedService, etc.). do you know why this might have shown up now? seems like no change has been made to these pages for a while

I noticed that your `package-lock.json` ended up changing the versions of almost every dependency that we have, and I think that may have been related. When I reverted all the changes to `package-lock.json` and regenerated a new one with the `url-search-params-polyfill` library, it seemed to have fixed the problem.

PR #830, opened 2019-08-11
[714] Host Casey youth guide on AskDarcel

drcaramelsyrup:
Although this change has dependencies on the linked branch, I decided this task involved sufficiently different changes (including changes to where we publicly serve files in our webpack config) to warrant another pull request. The desire is to have a convenient link e.g. sfserviceguide.org/youthguide to host the Casey youth guide pdfs and future guides. 

Since there isn't an intuitive location to cram the pdf in the Casey block, I decided to add a new link for now under the link to the FAQ guide.


richardxia:
Since I merged https://github.com/ShelterTechSF/askdarcel-web/pull/805 in, I went ahead and rebased this branch on top of master, and I edited the PR to target master instead of the branch from https://github.com/ShelterTechSF/askdarcel-web/pull/805.

AnnaKalkanis:
Ya, same for me. This wasn't loading when i ran it locally. It did seem to work, however, if I added the file extension back into the webpack file loader for pdfs. Don't know if that helps 🤷‍♀

drcaramelsyrup:
Bummer :/ I had been hoping that that would work, but for now if including the webpack extension does it I think that's what we should do for now.

PR #831, opened 2019-08-11
Change RelativeOpeningTime time representation

richardxia:
This is an almost complete rewrite of the `<RelativeOpeningTime>` component to have it make use of the new `RecurringSchedule` class that I created for unifying the time representation. Fortunately, there were a lot of unit tests written for this component, so I was able to make sure that most of the existing tests still pass after this rewrite.

For those whore are unfamiliar, the `<RelativeOpeningTime>` component is the "Open Now"/"Closed Today"/"Open Until ..." widget that appears on the search results page as well as the resource/service detail page.

Screen Shot 2019-08-11 at 12 27 56 PM

## Data model changes

I added a new class, `Duration`, to the set of Recurring* classes in order to represent time durations. The internal unit of representation is meant to be opaque, and consumers of this class are meant to use the helper methods in order to get the representation in terms of a familiar unit like minutes.

I also several more helper methods to the existing classes for performing operations such as taking the difference between two `RecurringTime`s, detecting overlaps between two `RecurringInterval`s, and finding the next closest `RecurringInterval` when given a `RecurringSchedule`.

## Glue code changes

I had to port over some of the code I wrote for handling the askdarcel-api API responses to the Algolia API. The purpose of this code is to take the raw representation of schedules and convert them over to the Recurring* classes, and since some of the data comes from Algolia, I needed to handle responses from Algolia as well.

## RelativeOpeningTime changes

Most of the `<RelativeOpeningTime>` component has been completely rewritten. Most of the "math-y" logic has been pulled into the RelativeSchedule models, but what remains is the logic for determining which scenario we are in and what to display.


AnnaKalkanis:
LGTM!

I pulled your branch and did some local testing and looks to be working correctly. Nice code clean up 💯

richardxia:
Thanks for the review! Going to merge this in now.

PR #832, opened 2019-08-11
Update DetailedHours to use new time helpers

richardxia:
This one is relatively straightforward compared to some of my recent PRs. This updates the `<DetailedHours>` component, which is used on the Resource show page, to use the new RecurringSchedule data model. I was able to simplify the component by pushing the conditionals down further into the component, which merges together most of the HTML structure and conditionalizes just the text content.

I was also able to remove a bunch of now unused time-related helper function from utils/index.js.

### DIFF
--- a/app/components/listing/Services.jsx
+++ b/app/components/listing/Services.jsx
@@ -80,7 +80,8 @@ updated
                 result={service.fee}
               />
               {service.notes.length ? <Notes notes={service.notes} /> : null }
-              <WeeklyHours schedule={service.schedule} />
+              {service.recurringSchedule.intervals.length > 0
+                && <WeeklyHours recurringSchedule={service.recurringSchedule} />}
             </ul>
             <div
               role="button"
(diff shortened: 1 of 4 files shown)


app/components/listing/Services.jsx
```diff
@@ -80,7 +80,8 @@ updated
                 result={service.fee}
               />
               {service.notes.length ? <Notes notes={service.notes} /> : null }
-              <WeeklyHours schedule={service.schedule} />
+              {service.recurringSchedule.intervals.length > 0
```
AnnaKalkanis: is it possible that recurringSchedule won't have intervals? Might want to guard against that either here by using lodash get, or in the RecurringSchedule class by setting it to an empty array if intervals aren't passed into the constructor.

AnnaKalkanis:
Nice! Just the one question, otherwise LGTM

app/components/listing/Services.jsx
richardxia: Good idea, added that in 3303699.

Issue #807, opened 2019-07-19
Add heading to tags list on right side of service/org pages, "Relevant Related Searches"


Maxastuart:
On the tags list on the right side of both the Service and Organization Pages:

- Add a heading for people browsing the site to understand the tags list's context and that the links are clickable.

- Should say something like Relevant Related Searches / Relevant Searches / Related Categories / etc. (dev's choice until @derekfidler returns)

- Considering adding a border, font matching, font size matching, or some other contextual help to separate this section of the page from the buttons above it.

- Make sure it looks good in all screen sizes (i.e. desktop & mobile)

Requested by Barry on July 17th, 2019.

derekfidler:
removed tags from UI until a better solution is found for them

PR #834, opened 2019-08-11
tags header on sidebar

daniel-owens:


app/pages/ServiceListingPage.jsx
```diff
@@ -183,6 +183,9 @@ class ServicePage extends React.Component {
               </div>
               <div className="listing--aside">
                 <ActionSidebar resource={resource} service={service} />
+                {service.categories ? (
+                  <p className="related-searches">Related Searches</p>
+                ) : null}
```
richardxia: Thanks for submitting this! Two comments:

1. You're failing the linter, which is complaining that you have trailing space characters at the end of this line. A good practice is to run the linter before you submit a PR so that you can correct the lint failures before having to wait for the automated tests to fail.
2. For ternary operator expressions in JSX, if you're just returning a `null` in the "else" case, then it's better to just use the `&&` operator, like `{service.categories && <p className="related-searches">Related Searches</p>}`.

daniel-owens: Thank you for those tips! Duly noted.

daniel-owens: Richard, I'm worried about any unwanted negative consequences of that ternary operator. I didn't notice anything unusual when I was testing it out but it's ultimately going to affect the display of every listing that is clicked on. My thinking was that, if a listing contained indexes from the categories array then go ahead and display the header. Otherwise, don't list the header. It almost seemed to easy of a fix but perhaps that's all we need. I'm working on the rendering of this feature on smaller screens as well.

richardxia: @daniel-owens, I think the conditional logic is fine. I believe the intent of #807 is to affect the display of every listing, so it makes sense that this logic would affect all pages.

Looking at it more closely, however, I realized that we probably want to make a slight modification to the conditional. In JavaScript, an empty array still evaluates to `true`, so what we really should do is use `service.categories && service.categories.length != 0` to decide whether there are any categories.

PR #835, opened 2019-08-11
824 - Remove HAP badge from resource page

AnnaKalkanis:
Remove HAP verified badge.

## Before

Screen Shot 2019-08-11 at 3 41 51 PM

## After

Screen Shot 2019-08-11 at 3 41 05 PM

Note that this component was also in the service listing page, but it doesn't look like it was ever actually rendering since the verification lives at the resource level. Went ahead and removed the component anyway.

### DIFF
--- a/app/components/layout/OrganizationCard.jsx
+++ b/app/components/layout/OrganizationCard.jsx
@@ -26,7 +26,6 @@ class OrganizationCard extends React.Component {
             {/* TODO Walking distance */}
           </h4>
           {/* TODO Add Rating */}
-          {/* TODO HAP Certification */}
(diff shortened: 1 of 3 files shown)


derekfidler:
It was mentioned that we wanted to keep the "Verified DATE" but remove the "by HAP" part. That way people accessing the site would know when the content was last reviewed but it just wouldn't mention HAP as per MOHCD's request.

Maxastuart:
We discussed it live at Sunday's Worksesh and realized that date data
wasn't recorded the way I previously thought. Neither certified_at nor
verified_at are updated when people at HAP or Datathons review entries, so
we decided to hide it for now.

PR #841, opened 2019-08-25
[442] Resources / services with "hours unknown" to display "Call for Hours."

drcaramelsyrup:
To match with the corresponding API pull request connected to the same issue, which introduces an `hours_known` db field.

The frontend changes to actually implement this as a checkbox in the "Edit" page seemed pretty involved, so this is what I have for now.

For comparison:

Screen Shot 2019-08-25 at 12 35 25 PM
Screen Shot 2019-08-25 at 12 50 03 PM


derekfidler:
I reviewed Q Foundation, which is our patient zero for this bug, on staging and still saw some weirdness with the hours. In search results it says open 24 hours, the resource page says open 24 hours up top, followed by "null", but the location component says Call for Hours.

Screen Shot 2019-08-28 at 13 53 56

PR #853, opened 2019-09-06
Fix lint on search components.

richardxia:
Just gradually whittling down the list of files being ignored by ESLint.


richardxia:
> Some of these are really interesting. Thanks for fixing these.

Some of these changes were beyond what the linter preferred and was just me rewriting some code in a way that I personally prefer. For example, changing any class-based React Components that don't have any state into function-based Components, or changing that `<a>` tag to a `<button>` while updating the CSS rules to make sure link still appeared the same.

PR #854, opened 2019-09-07
Remove unused packages

richardxia:
I went through and removed a bunch of unused NPM packages. This included completely removing Karma and Jasmine, since it looks like we're now using Mocha as our testing framework. I double checked that even with all of these packages removed, the built bundle.js has the exact same MD5 checksum, so none of these had an effect on the actual build JS.


richardxia:
Thanks!

PR #855, opened 2019-09-07
Confirm adding resource fields

alexanderturinske:
confirms
- website
- email
- description
- phone

There is a bug with adding a note for a new resource

Relates to #427 
Relates to https://github.com/ShelterTechSF/askdarcel-api/issues/449

### DIFF
--- a/testcafe/newResource.js
+++ b/testcafe/newResource.js
@@ -25,7 +25,7 @@ test('Add new resource, basic', async t => {
     website: 'https://theBestResource.best',
     email: 'theBestResource1337@aol.com',
     description: 'A description of this resource',
-    notes: 'A new note about this resource',
+    note: 'A new note about this resource',
   };
 
   // Add name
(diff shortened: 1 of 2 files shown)


testcafe/newResource.js
```diff
@@ -7,23 +7,63 @@ fixture`Add New Resource`
...
+  await t
+    .typeText(phone.number, data.phone.number, { replace: true })
+    .typeText(phone.serviceType, data.phone.type, { replace: true });
+
+  // Add website
+  await t.typeText(newResourcePage.website, data.website, { replace: true });
+
+  // Add email
+  // For some reason, testCafe puts the email in backwards
+  const reversedEmail = data.email.split('').reverse().join('');
```
richardxia: Are you sure about this? It looks like you're failing the Travis CI test because it's being input backwards:

```
AssertionError: expected 'moc.loa@7331ecruoseRtseBeht' to deeply
      equal 'theBestResource1337@aol.com'
```

alexanderturinske: Well, it fails locally without this, but fails Travis CI with it, but Travis CI is more important.
Fixed.

image

richardxia:
BTW, I saw that I'm running into the backwards text bug too when running TestCafe locally. It looks like it's a bug in TestCafe starting with Chrome 75 (https://github.com/DevExpress/testcafe/issues/3865), but I think it's supposed to be fixed in TestCafe 1.2.1 (we're still on 0.23.3).

I can try to upgrade us to the latest TestCafe and latest Chrome.

Maxastuart:
Richard, sounds like a good thing moving forward, I'm for it

PR #856, opened 2019-09-07
Fix lint on components/ui files.

richardxia:
Note that this is probably easier to review with whitespace diffs ignored.

This should be the last of the lint fixes for files that are actually used in production. Everything else left in the `.eslintignore` is either the old admin pages or debugging pages.

### DIFF
--- a/app/components/ui/Accordion/AccordionItem.jsx
+++ b/app/components/ui/Accordion/AccordionItem.jsx
@@ -2,33 +2,19 @@ import React from 'react';
 import PropTypes from 'prop-types';
 
 // TODO Enforcing the user to add their own dropdown with a custom renderer is a bit clunky
-class AccordionItem extends React.Component {
-  render() {
-    return (
-      <div>
-        <header>
-          {
-            this.props.headerRenderer
-              ? this.props.headerRenderer(this.props.title)
-              : (
-                <div>
-                  { this.title }
-                  <div className="selector">
-                    <i className="material-icons">keyboard_arrow_down</i>
-                  </div>
-                </div>
-              )
-          }
-        </header>
-        <section>{ this.props.children }</section>
-      </div>
-    );
-  }
-}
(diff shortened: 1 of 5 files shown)


app/components/ui/LandingPageEligibilityBlock.jsx
```diff
@@ -1,46 +1,40 @@
...
-                count={eligibility.service_count}
-                url={this.buildBlockUrl(eligibility.name)}
-              />
-            ))
-            }
-          </div>
-          {/* <a className="scroll-button scroll-button-right">&gt;</a> */}
-        </div>
+      </h2>
+      {/* <a className="scroll-button scroll-button-left">&lt;</a> */}
```
drcaramelsyrup: Should the "Properly implement horizontal scroll buttons" comment be preserved / moved here if this other comment is also being preserved?

app/components/ui/Accordion/AccordionItem.jsx
```diff
@@ -2,33 +2,19 @@ import React from 'react';
...
-        </header>
-        <section>{ this.props.children }</section>
-      </div>
-    );
-  }
-}
+const AccordionItem = ({ children, headerRenderer, title }) => (
+  <div>
+    <header>
+      {headerRenderer(title)}
```
drcaramelsyrup: Should this be null-checked still (or the headerRenderer set as an `isRequired` prop type)?

richardxia: I thought the same thing at first, but actually it looks like there's a `defaultProp` for this property that renders just the title as a string.

app/components/ui/LandingPageEligibilityBlock.jsx
richardxia: Yeah, sorry, I probably got overzealous deleting and moving around code. I'll add the comment back here.

PR #857, opened 2019-09-07
Restore disabled ESLint rules

richardxia:
We had a number of ESLint rules that I believe were probably disabled in an attempt to get ESLint running at all on our codebase. This restores several rules that we disabled, some of which didn't even require any code changes.


richardxia:
I'm going to merge this in since I don't think anyone is going to have a problem with these changes.

PR #862, opened 2019-09-08
Fixed spacing issue

azhelle16:


app/pages/OrganizationEditPage.jsx
```diff
@@ -784,7 +784,7 @@ class OrganizationEditPage extends React.Component {
             />
             <p>
 If you&#39;d like to add formatting to descriptions, we support
-              <a href="https://github.github.com/gfm/" target="_blank" rel="noopener noreferrer">Github flavored markdown</a>
+              <a href="https://github.github.com/gfm/" target="_blank" rel="noopener noreferrer"> Github Flavored Markdown</a>
```
drcaramelsyrup: I actually think we'd prefer not to make the space part of the link.
React really tries to trim all whitespace off the ends, like tabs and spaces, but you can force it to render a space with something like this:

```
If you&#39;d like to add formatting to descriptions, we support
              {' '}
              <a href="https://github.github.com/gfm/" target="_blank" rel="noopener noreferrer">Github flavored markdown</a>
```

which makes that space an actual JSX expression and tells React, "no srs render this whitespace please."

azhelle16: Can I use the same pull request for my recent push? I know that this is only minor. I take off the space from the link and instead put &nbsp; before it.

drcaramelsyrup: Sure, that's fine.

richardxia:
Thanks for making the change, @azhelle16! I added one small commit to revert changes to package-lock.json, but once Travis CI passes, I'll merge this in.

Issue #420, opened 2018-03-05
Confirm all service fields are displayed correctly on resource/service pge


jjfreund:
The following fields of each service should be displayed:

--Service Name
--Service Description
--Application Process, if not null
--Required Documents, if not null
--Fees, if not null
--Notes, as list
--Service Schedule

alexanderturinske:
There are tests to check for some of these, but not all.

alexanderturinske:
- [ ] Service Name
- [ ] Service Description
- [ ] Application Process, if not null
- [ ] Required Documents, if not null
- [ ] Fees, if not null
- [ ] Notes, as list
- [ ] Service Schedule

PR #863, opened 2019-09-08
Issue #420: Verify service data

alexanderturinske:
- name
- description
- address
- phone
- note

Relates to #420

### DIFF
--- a/testcafe/editResource.js
+++ b/testcafe/editResource.js
@@ -1,8 +1,10 @@
 import ResourcePage from './pages/ResourcePage';
 import EditResourcePage from './pages/EditResourcePage';
+import ServicePage from './pages/ServicePage';
 
 const resourcePage = new ResourcePage();
 const editResourcePage = new EditResourcePage();
+const servicePage = new ServicePage();
 
 fixture`Edit Resource`
   .page(ResourcePage.url(1));
@@ -204,6 +206,19 @@ test('Delete Resource Note', async t => {
 });
 
 test('Add new service', async t => {
+  const SERVICE_DATA = {
+    NAME: 'Test Service',
+    NICKNAME: 'Best Service',
+    EMAIL: '1800bestservice@best.com',
+    DESCRIPTION: 'I go here when I want the best best service.',
+    APPLICATION_PROCESS: 'Walk in and ask for assistance',
+    REQ_DOCS: 'No ID required',
+    INTERPRETATION_SERVICES: 'ASL provided',
+    COST: 'Free',
+    WAIT_TIME: 'N/A',
+    WEBSITE: 'https://bestservice.com',
+  };
+
   // Navigate to edit page
   await resourcePage.clickEditButton(t);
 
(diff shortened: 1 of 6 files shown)


testcafe/service.js
```diff
@@ -4,20 +4,30 @@ import EditResourcePage from './pages/EditResourcePage';
 const servicePage = new ServicePage();
 const editResourcePage = new EditResourcePage();
 
-// Define service id to navigate to
-// Tests depend on service id
-const serviceId = 5;
+// Define service information to test
+// Tests depend on service information
+// TODO: Create a new service and add all this information in the future
+//  to not rely on hardcoded data
```
richardxia: Looks like most of the tests are failing because this data doesn't match. You probably want to update this part of the DB fixture populator to populate the same values: https://github.com/ShelterTechSF/askdarcel-api/blob/518efa307db3c7d4cc1ef00831358b0640dabfdd/lib/sheltertech/db/fixture_populator.rb#L148

alexanderturinske: Ah, I was going off of the data we populate in the our local DBs. I will put up an MR for the testcafe data

alexanderturinske: Change of plans; since we already had a test as to whether adding a service to a resource worked, I just extended that test to also verify the information on the service page.

alexanderturinske:
This was failing for some weird reason that I couldn't debug before, but I pulled in master and now it is failing for some other reason that seems debug-able, so I will fix this.

testcafe/editResource.js
```diff
@@ -203,7 +205,20 @@ test('Delete Resource Note', async t => {
     .eql(originalCount - 1);
 });
 
-test('Add new service', async t => {
+test.only('Add new service', async t => {
```
alexanderturinske: Whoops! Remove `.only`. That was for testing purposes

alexanderturinske: Done

testcafe/pages/ServicePage.js
```diff
@@ -1,16 +1,23 @@
 import { ReactSelector } from 'testcafe-react-selectors';
 import config from '../config';
 
-export default class ResourcePage {
+export default class ServicePage {
```
alexanderturinske: This was a copy/paste error that the class was called `ResourcePage` when it was in the `ServicePage` file. It didn't error before because `export default` lets you import it with whatever name you want. Updated for consistency

testcafe/pages/ServicePage.js
```diff
@@ -1,16 +1,23 @@
...
     this.details = baseSelector.find('.listing--main--left--details');
+    this.directionsButton = baseSelector.find('.action-sidebar--directions');
     this.editButton = baseSelector.find('.action-sidebar--edit');
+    this.email = baseSelector.find('.listing--main--left--contact tr').nth(1).find('td');
+    this.name = baseSelector.find('.listing--main--left > header h1');
+    this.note = this.details.find('tbody td');
+    this.phone = baseSelector.find('.listing--main--left--contact tbody td');
     this.printButton = baseSelector.find('.action-sidebar--print');
-    this.directionsButton = baseSelector.find('.action-sidebar--directions');
+    this.requiredDocs = baseSelector.find('.listing--main--left--details tr').nth(1).find('td');
```
alexanderturinske: `directionsButton` was moved above to alphabetization.  The diff this is confusing to read.

PR #878, opened 2019-10-10
add typeform integration; fix lint

quanhuynh:
added typeform popup to the guided pathway links

8e10903f096ae3fe690d5bad546e5d20

### DIFF
--- a/app/pages/HomePage/components/GuideList/GuideList.jsx
+++ b/app/pages/HomePage/components/GuideList/GuideList.jsx
@@ -24,41 +24,35 @@ function typeform(event, link) {
   typeformReference.open();
 }
 
-const GuideCard = props => {
-  const { name } = props;
-  const { img } = props;
-  const { link } = props;
-  const { partner } = props;
-  const { partnerImg } = props;
-
-  return (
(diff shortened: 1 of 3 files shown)


app/pages/HomePage/components/GuideList/GuideList.jsx
```diff
@@ -12,33 +13,57 @@ import ImgYouthHomelessnessPartner from './assets/logos/larkin.png';
...
+  );
+  typeformReference.open();
+}
+
+const GuideCard = props => {
+  const { name } = props;
+  const { img } = props;
+  const { link } = props;
+  const { partner } = props;
+  const { partnerImg } = props;
```
richardxia: Nit: My preference is to do all this in a single destructuring assignment in the function argument definition itself. Something like this: https://github.com/ShelterTechSF/askdarcel-web/blob/099d26fdf4f2395de8cd1c2aeddca25a57ba07d9/app/components/ui/HamburgerMenu.jsx#L37-L44

richardxia: I'm going to go ahead and make this change right now, since it seems like we're interested in getting the landing page deployed to staging soon.

PR #880, opened 2019-10-14
Remove unnecessary stuff (including admin pages)

richardxia:
I just wanted to do a bit of cleanup of stuff that's no longer necessary.

The best way to review this PR is to go commit by commit, which I tried to keep independently reviewable:

- https://github.com/ShelterTechSF/askdarcel-web/pull/880/commits/737c61ee8b5ec2690e9dfafaef2010d1557bbb51 - This removes the `name` prop from all of the `<Route>`s, since it's not defined in the API docs and we're pretty inconsistent with how we named the routes anyway. This also happened to make the routes easier to read because the `path` prop is now left-aligned.
- https://github.com/ShelterTechSF/askdarcel-web/pull/880/commits/32a9e9638addf032e2ca11867b14b89e28c8686b - Here I removed all of the admin pages, the associated components, and the associated CSS files. I think we're going to need to do a big overhaul of the admin pages anyway, so it might be better to throw away all of that code and rebuild them from scratch.
- https://github.com/ShelterTechSF/askdarcel-web/pull/880/commits/16ddfc8ed6a1d1becedff7253c77eb5f168e58ce - This removes the `react-autosize-textarea` package, which was only being used by the admin pages.
- https://github.com/ShelterTechSF/askdarcel-web/pull/880/commits/00a29477949431512df1abb756a3089b4a636b4b - This removes the pages under the `app/pages/_unused` directory, which, as the name implies, are unused.
- https://github.com/ShelterTechSF/askdarcel-web/pull/880/commits/5e7f0f753265e78c3787f4b98bf8a051cb371c15 - This one is not removing code. Instead this is me change the URLs for the Organization Edit and New pages from `/resource/edit?resourceid=` to `/organizations/:id/edit` and `/resource/new` to `/organizations/new`, respectively. This makes them match the view Organization page's URL change in https://github.com/ShelterTechSF/askdarcel-web/pull/874. This also removes another query parameter in favor of a path parameter, which is easier to handle in React Router v4.
- https://github.com/ShelterTechSF/askdarcel-web/commit/2f79c786fd038a3007a137afd5c377b5e18de583 - Lastly, this just alphabetizes the URLs in the routes component to make them easier to read.


quanhuynh:
LGTM. thanks for doing the work @richardxia
I believe Karla was working on some admin stuff, should let her know that the old admin code is scrapped @lgarofalo @jjfreund 
for the url change, do you know if there are any off-site references to those URLs anywhere else? may be good to sync with the rest of ShelterTech and update those as well (e.g i think some datathon material links directly to /resource/ urls)

richardxia:
> I believe Karla was working on some admin stuff, should let her know that the old admin code is scrapped @lgarofalo @jjfreund

@Raawr, are you working on any of the admin code that I removed in this PR? I had figured that most of this code was so old and unused that we'd be better off rewriting it from scratch. Let me know if I should hold off from making any of these changes.

> for the url change, do you know if there are any off-site references to those URLs anywhere else? may be good to sync with the rest of ShelterTech and update those as well (e.g i think some datathon material links directly to /resource/ urls)

I want to say I don't think there would be anything that links to those pages. Note that I only changed the edit and create new resource URLs, which I don't think are URLs that would typically be linked to by anywhere else. What I could do is add in redirects, much like how I did with the organization/resource show page, which will at least make sure that the old URLs still work.

Good call on the datathon links. I'll send a message to the datathon team as well as the org in general about the updated URLs so that we can correct them.

richardxia:
Pushed up a new commit that adds in legacy redirects from the old edit and new resource URLs to the new ones: https://github.com/ShelterTechSF/askdarcel-web/pull/880/commits/170b6551c029edb0c63f2933abc59fa65b930c65

AnnaKalkanis:
This is awesome. Those new resource create/edit routes make me so happy. Thanks for doing this!

LGTM!

richardxia:
Thanks for the reviews!

Raawr:
I think everything was answered already in slack just in case though it's no problem for me (and thank you!)

PR #882, opened 2019-10-20
Fix styling of MOHCD Funded badge on map tooltips

AnnaKalkanis:
This is a bit of a continuation of https://github.com/ShelterTechSF/askdarcel-web/pull/875, where I added tooltips with some details when you click on a map marker. I just noticed that these tooltips needed a little styling love for when they are displaying the `MOHCD Funded` badge.

Notice that with the current styling, the header overlaps with the MOHCD badge.
Screen Shot 2019-10-20 at 1 23 57 PM

Moved the header & badge into the same flexbox so they play a little nicer with each other.
Screen Shot 2019-10-20 at 1 59 15 PM

Also noticed that for some reason we were showing the badge regardless of whether or not a resource actually had the `is_mohcd_funded` property as true on the backend. Fixed that while I was in there.

### DIFF
--- a/app/components/search/ResourceEntry.jsx
+++ b/app/components/search/ResourceEntry.jsx
@@ -19,7 +19,18 @@ class ResourceEntry extends Component {
       <Link to={`/organizations/${hit.resource_id}`}>
         <li className="results-table-entry resource-entry">
           <div className="entry-details">
-            <h4 className="entry-headline">{`${hitNumber}. ${hit.name}`}</h4>
+            <div className="entry-header">
+              <h4 className="entry-headline">{`${hitNumber}. ${hit.name}`}</h4>
+              {hit.is_mohcd_funded
+                ? (
+                  <div className="mohcd-funded">
+                    <img src={images.mohcdSeal} alt="MOHCD seal" />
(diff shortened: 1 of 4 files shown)


app/components/search/ResourceEntry.jsx
```diff
@@ -19,7 +19,18 @@ class ResourceEntry extends Component {
       <Link to={`/organizations/${hit.resource_id}`}>
         <li className="results-table-entry resource-entry">
           <div className="entry-details">
-            <h4 className="entry-headline">{`${hitNumber}. ${hit.name}`}</h4>
+            <div className="entry-header">
+              <h4 className="entry-headline">{`${hitNumber}. ${hit.name}`}</h4>
+              {hit.is_mohcd_funded
+                ? (
```
richardxia: Nit: If the "else" case just returns a `null`, then we can just use `&&`. E.g.

```jsx
{hit.is_mohcd_funded && <div>...</div>}
```

AnnaKalkanis: oh ya, good call. I'll merge this one and update in the consolidated component!

PR #888, opened 2019-10-20
Upgrade React Router to v4

richardxia:
## Overview

This is a pretty big change, since React Router 4 is not compatible with 3. Some of the biggest backwards incompatible changes are:

- Package name has been renamed from `react-router` to `react-router-dom`. (It's because they also created a `react-router-native` package for React Native)
- The Router and Routes have been changed so that they are both plain React components instead of some magic thing.
- History is completely changed, so you have to choose from one of three Router types to get a specific type of history, and you can no longer use the global `browserHistory` object. Instead, you need to use `withRouter` in order to shove the current history object into your own React Components' props
- `react-router-redux` has been removed, and there's a separate library that you can use called [`connected-react-router`](https://github.com/supasate/connected-react-router) for combining a React Router history with a Redux store. React Router actually [recommends not trying to link your routes to your Redux store](https://reacttraining.com/react-router/web/guides/redux-integration).

Note that we should probably do pretty extensive testing once this is merged in, since I don't have a great way to systematically test everything.

## Changes in this PR

Most of the changes here are related to router history changes to React Router. Let me try to break down and explain the biggest changes.

### New Browser History Store (https://github.com/ShelterTechSF/askdarcel-web/commit/683a0f8303795ad9ccd01e2940268447d8b7276e)

This follows the instructions in the README for [`connected-react-router`](https://github.com/supasate/connected-react-router) for setting up your Redux store with a React Router 4 history. One other related change here was changing up the component hierarchy from:

init -> Router -> routes -> App + individual Routes

to

init -> Router -> App -> routes -> individual Routes

I could use a second set of eyes on these changes, since I'm not really familiar with setting up Redux, so I'm hoping that I did it correctly.

### Hamburger Menu (https://github.com/ShelterTechSF/askdarcel-web/commit/be17a7818c0be1a0f2c9694d86d115374d99aa80)

Here most of the changes are related to React Router having a new `<NavLink>` component that is like a normal `<Link>` except it can tell whether the current Link is the active route and style it accordingly. This allowed us to avoid implementing our own logic for that and also not have to pass in the `location` object into the component.

### Changing how `history` is accessed

The following commit is probably the best example of the kind of change that I had to make in many places:

https://github.com/ShelterTechSF/askdarcel-web/commit/337486927a3cf4bfaa0a1f4f37d9dac14763642d

Basically, instead of importing `browserHistory`, which no longer exists, we have to wrap the React component with `withRouter` so that it injects the `history` prop into the component. This affects anywhere we do something like `history.push()` to navigate to a new URL, since we can't depend on the global variable anymore.

### Alert prompt for unsaved changes (https://github.com/ShelterTechSF/askdarcel-web/commit/70c94320c2e69a41b3089ebc3fc8d469b7ad128a)

In React Router 4, there's a new way of detecting whether someone is leaving a page and presenting them with a prompt, via the [`<Prompt>`](https://reacttraining.com/react-router/core/api/Prompt) component. This replaces React Router 3's `setRouteLeaveHook()`.

### Scroll to top on navigating to page (https://github.com/ShelterTechSF/askdarcel-web/commit/d887cc6c3002adbd750ec5bff6a989bacde9b974)

There's a new way to do scrolling to the top of the page when navigating to a new route.


richardxia:
I'm going to merge this in now to avoid this getting too stale. If this causes any issues, hopefully we catch them in staging.

PR #889, opened 2019-10-22
Add multiple addresses edit page

ashleyabrooks:
Having trouble getting the previously saved address to populate after I add the `Add` button from the `editCollectionHOC` component. Trying to re-use what we already have similar to how we're doing it for the `EditPhones` component [here](https://github.com/ShelterTechSF/askdarcel-web/blob/master/app/components/edit/EditPhones.jsx) but it's not working properly. I'm probably doing something silly because my React knowledge is pretty rusty. 

I also tried starting again from master and just adding the `editCollectionHOC` component but I'm running into the same issue.

### DIFF
--- a/app/components/edit/EditAddress.jsx
+++ b/app/components/edit/EditAddress.jsx
@@ -107,7 +107,7 @@ EditAddress.propTypes = {
 };
 
 
-const EditAddressCollection = editCollectionHOC(EditAddress, 'Addresses', {}, true);
+const EditAddressCollection = editCollectionHOC(EditAddress, 'Addresses', {}, 'Add Address');
 EditAddressCollection.displayName = 'EditAddressCollection';
 
 
--- a/app/components/edit/EditAddress.jsx
+++ b/app/components/edit/EditAddress.jsx
@@ -1,124 +1,130 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import editCollectionHOC from './EditCollection';
 
-const EditAddress = ({ address, setAddress, setHasLocation }) => (
-  <AddressForm
-    setAddress={setAddress}
-    setHasLocation={setHasLocation}
-    address={address}
-  />
+const HasPhysicalLocationToggle = ({ hasLocation, setHasLocation }) => (
+  <label className="inline-checkbox">
+    <input
+      type="checkbox"
+      className="input-checkbox"
+      checked={!hasLocation}
+      onChange={e => setHasLocation(!e.target.checked)}
+    />
+    No Physical Location
+  </label>
 );
 
-EditAddress.propTypes = {
-  address: PropTypes.object,
-  setAddress: PropTypes.func.isRequired,
-  setHasLocation: PropTypes.func.isRequired,
-};
 
-EditAddress.defaultProps = {
-  address: null,
-};
-
-const AddressForm = ({
-  setHasLocation, setAddress, address,
-}) => {
+const EditAddress = ({ index, item, handleChange }) => {
+  const address = item;
   const handleFieldChange = event => {
     const { target } = event;
     const { value, dataset: { field } } = target;
-    setAddress({ ...address, [field]: value });
+    handleChange(index, { ...address, [field]: value });
   };
   return (
-    <li key="address" className="edit--section--list--item">
-      <label htmlFor="address">Address</label>
-      <label className="inline-checkbox">
-        <input
-          type="checkbox"
-          className="input-checkbox"
-          checked={!address}
-          onChange={e => setHasLocation(!e.target.checked)}
-        />
-        No Physical Location
-      </label>
-      {address
-        && (
-          <div>
-            <input
-              type="text"
-              className="input"
-              placeholder="Name"
-              data-field="name"
-              value={address.name}
-              onChange={handleFieldChange}
-            />
-            <input
-              type="text"
-              className="input"
-              placeholder="Address 1"
-              data-field="address_1"
-              value={address.address_1}
-              onChange={handleFieldChange}
-            />
-            <input
-              type="text"
-              className="input"
-              placeholder="Address 2"
-              data-field="address_2"
-              value={address.address_2}
-              onChange={handleFieldChange}
-            />
-            <input
-              type="text"
-              className="input"
-              placeholder="Address 3"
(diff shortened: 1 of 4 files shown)


app/components/edit/EditAddress.jsx
```diff
@@ -51,120 +65,140 @@ class EditAddress extends Component {
...
-      <AddressForm
-        handleAddressChange={this.handleAddressChange}
-        handleNoLocationChange={this.handleNoLocationChange}
-        noLocation={noLocation}
-        address={address}
-      />
+      <div>
+        <label>
+          <input
+            type="checkbox"
```
ashleyabrooks: Moved this out of the `AddressForm` component because I don't want it to replicate with each child address, but it's not working as expected. It still re-renders every time you click the Add button.

app/components/edit/EditAddress.jsx
```diff
@@ -51,120 +65,140 @@ class EditAddress extends Component {
...
+                onChange={handleAddressChange}
+              />
           </div>
         )
     }
   </li>
 );
 
-export default EditAddress;
+const EditAddresses = editCollectionHOC(EditAddress, 'Addresses', {}, true);
```
ashleyabrooks: As soon as I add this (with the intention of reusing the `Add` button from other edit components like the `EditPhone` component, any previously saved address doesn't re-populate on the page.

richardxia:
The state management is kind of messy on the Edit page. Over time I've been trying to hoist state up to the top level so that there is no state in the lower level components, which I feel would make it easier to work on the code here.

The other related problem is that we have a bunch of duplicated state where each level of the Edit page component hierarchy seems to have a separate copy of the same state values, and the update hooks keep them in sync. I've also been trying to eliminate the duplicate state over time.

You may want to use the React Developer Tools to inspect the state on that page to get a better feeling for how state is stored at each layer of the page. You may also need to make some modifications to the state at the top level of the app to change the address from being a single property to being a property that contains an array of addresses.

app/pages/OrganizationEditPage.jsx
```diff
@@ -737,7 +737,7 @@ class OrganizationEditPage extends React.Component {
             />
           </li>
 
-          <EditAddress
+          <EditAddresses
             address={resource.address}
```
richardxia: Does this need to be somehow changed to an array of addresses?

AnnaKalkanis:
I'm not super familiar with the `editCollectionHOC` but I think you will need to pass an array of addresses as a `collection` prop in `EditAddresses`. Ie `<EditAddresses collection={[resource.addresses]}>`

app/components/edit/EditAddress.jsx
```diff
@@ -51,45 +65,54 @@ class EditAddress extends Component {
...
-        No Physical Location
-    </label>
+
+
     {!noLocation
         && (
           <div>
+            <label htmlFor={htmlID}>Address</label>
             <input
+              id={htmlID}
```
richardxia: BTW, this is not a good idea because in HTML, every `id` needs to be unique, and if you put the same ID on multiple `<input>` elements, there's no way to distinguish between them.

app/components/edit/EditAddress.jsx
richardxia: It looks like this is because anything that gets passed to the `editCollectionHOC` component gets rendered once per item. We'll have to split up the form into the part that should only render once and the part that should render once per item, and then make sure that the `editCollectionHOC` only gets called on the latter.

richardxia:
After staring at these files for like an hour, I realized that the state management is really crazy for this data, and it's kind of similar to a lot of the issues I had last year with changing the representation of open/close times. Sorry to assign this to you, @ashleyabrooks, as one of your first larger tickets!

I think I want to create a separate PR where I refactor the state management for this component in order to make it easier to convert it into multiple addresses. The main challenge with making this change without refactoring it first is that there are multiple copies of the state across different components, and keeping them in sync across a large refactoring is challenging. I think it'll be much simpler when we aren't duplicating state management.

For example, the following statement is actually quite complex and subtle:

https://github.com/ShelterTechSF/askdarcel-web/blob/00cd64bc58691d5d75be1cb9c528bd32a54f60bd/app/components/edit/EditAddress.jsx#L29-L31

It's first calling `updateAddress()`, which calls the callback that the parent component passed in to update the parent component's state, and then taking the result of that and setting the current state of this component to match.

richardxia:
We've got a new set of design mocks for this feature, so I'm going to try to revive this branch. We're going to have new modals for for actually entering in address info, and the main Organization and Service sections will have buttons to open up the modals.

### Resource Location
Screen Shot 2020-03-22 at 4 05 23 PM

### Service Location
Screen Shot 2020-03-22 at 4 05 45 PM

### Modal
Screen Shot 2020-03-22 at 4 07 51 PM

richardxia:
@jjfreund, @azhelle16 

I think this branch is now up to date. The primary goal of this PR is to get the multiple addresses data model working on the edit page, but this only updates the UI to accommodate it using the existing design, so it doesn't introduce a modal yet. I think it's still worth getting this in so that it's easier to make the new UI changes with the modal-based UI. @quanhuynh and @davidagustin should be aware of this in case if this touches any of the code they're working on.

richardxia:
This shouldn't go in until we actually get the two API changes in for adding a brand new address and removing an existing address.

azhelle16:
@richardxia, just to clarify the mock UI, the new address is added one at a time? or will it be hard for the API to save multiple addresses altogether?

richardxia:
The UI mocks show one address being added at a time, using the modal, but you can save multiple changes at the same time. We currently do this using multiple API calls, so we don't need to worry about having a single API call support adding or editing multiple addresses at the same time.

richardxia:
I updated this to work with @jjfreund's API changes in https://github.com/ShelterTechSF/askdarcel-api/pull/520, and they seem to work for me. I probably need to do much more extensive testing and add a TestCafe test, though.

richardxia:
Since this is passing CI checks against the latest API changes from @katerina-kossler's PR in https://github.com/ShelterTechSF/askdarcel-api/pull/571, I'm going to go ahead and merge this in. Thank you to everyone who took a part in working on this!

PR #890, opened 2019-10-22
landing page updates from MOHCD

derekfidler:
Updated landing page styling and copy from MOHCD requests

PR #891, opened 2019-10-23
Tweaks to CSS for new landing page

richardxia:
Here's what I'm suggesting we could do in #890 instead of creating a separate `<div>` for the background and absolutely positioning it.


derekfidler:
Nice! This seems like a much more elegant solution. Thanks.

PR #894, opened 2019-10-31
Multiple Addresses - update search page, resource listing and service listing

AnnaKalkanis:
This updates the places we reference addresses so that they can handle multiple addresses per resource. This is dependent on a few backend changes, and shouldn't be merged until those are ready. Namely:
1 ) Resources need to be updated so they have an `addresses` property that is an array of addresses for the give resource.
2) Services has a `resource` associated with them. That resource should have the same `addresses` property as when you GET just the resource itself.
3) Currently algolia returns resources and services as `hits`. These hits should have an `addresses` property with an array of addresses
4) Currently, the algolia hits have `_geoloc` property with lat and long data for the resource. Now that there are multiple locations for any resource, we should move these lat/long properties onto the address they are associated with in the `addresses` array.

Screen Shot 2019-10-30 at 7 54 59 PM


app/components/search/SearchEntry.jsx
```diff
@@ -7,6 +7,19 @@ import { RelativeOpeningTime } from '../listing/RelativeOpeningTime';
 import './SearchEntry.scss';
 
 class SearchEntry extends Component {
+  renderAddressMetadata() {
+    const { hit } = this.props;
+    const { addresses } = hit;
+    if (addresses && addresses.length) {
```
alexanderturinske: if you make ln 12 use a default value for destructuring (i.e. `const { addresses = [] } = hit;`), then you could un-nest this if statement and just have it as three cases instead of two, i.e.

```
if (address.length === 0) {
    return <span>No address found</span>;
} else if (addresses.length > 1) {
    return <span>Multiple locations</span>;
} else if (addresses[0].address_1) {
    return <span>{addresses[0].address_1}</span>;
}

File: app/components/search/SearchEntry.jsx [hunk 2]
```diff
@@ -7,6 +7,19 @@ import { RelativeOpeningTime } from '../listing/RelativeOpeningTime';
...
+      if (addresses.length > 1) {
+        return <span>Multiple locations</span>;
+      } if (addresses[0].address_1) {
```
alexanderturinske: `else if`

app/components/search/SearchEntry.jsx
```diff
@@ -42,12 +55,7 @@ class SearchEntry extends Component {
             </p>
             )}
             <p className="entry-meta">
-              {type === 'service' && (
-                <span>{hit.addresses && hit.addresses.address_1 ? hit.addresses.address_1 : 'No address found'}</span>
-              )}
-              {type === 'resource' && (
-                <span>{hit.address && hit.address.address_1 ? hit.address.address_1 : 'No address found'}</span>
-              )}
+              {this.renderAddressMetadata()}
```
alexanderturinske: nice change

app/components/search/SearchEntry.jsx
AnnaKalkanis: ohh good call. Updated!

richardxia:
Note that you are failing lint checks, so you should probably fix those even if you are expecting that some of the TestCafe tests will fail until the backend changes are merged in.

richardxia:
Got some of the TestCafe tests to run and pass after merging in https://github.com/ShelterTechSF/askdarcel-api/pull/462, but it looks like a couple are failing. Going to try to get these passing.

richardxia:
Determined that one of the failing TestCafe tests is due to stale data in Algolia. We'll need to do a full reindex in Algolia after these changes.

The second failing TestCafe test is due to the New Organization page being broken. Trying to see if there's a quick way to rescue it.

app/components/search/SearchEntry.jsx
```diff
@@ -7,6 +7,21 @@ import { RelativeOpeningTime } from '../listing/RelativeOpeningTime';
 import './SearchEntry.scss';
 
 class SearchEntry extends Component {
+  // eslint-disable-next-line consistent-return
```
richardxia: We probably shouldn't disable this. If we hit this case, then we'll silently not return anything. If we fall out of all of the `if` checks, then we probably want to return some "No address found" message.

app/components/search/SearchMap.jsx
```diff
@@ -84,17 +84,20 @@ const SearchMap = ({
...
-      hitsPerPage={hitsPerPage}
-      lat={hit._geoloc ? hit._geoloc.lat : 0}
-      lng={hit._geoloc ? hit._geoloc.lng : 0}
-      key={hit.objectID}
-      index={index}
-    />
-  ));
+  const markers = hits.map((hit, index) => {
+    const { addresses = [] } = hit;
+    return addresses.length ? addresses.map(address => (
```
richardxia: This `address.length` check is unnecessary. `Array.prototype.map()` will already return an empty array for an empty array input.

app/pages/ServiceListingPage.jsx
```diff
@@ -22,16 +22,17 @@ import MOHCDBadge from 'components/listing/MOHCDBadge';
 import { getSiteTitle } from '../utils/whitelabel';
 
 // TODO This should be serviceAtLocation
-const getServiceLocations = (service, resource, recurringSchedule) => (resource.address
-  ? [resource.address].map(address => ({
+const getServiceLocations = (service, resource, recurringSchedule) => {
+  const { addresses = [] } = resource;
+  return addresses.length ? addresses.map(address => ({
```
richardxia: Same thing here about the unnecessary array length check.

PR #895, opened 2019-10-31
Remove Edit button from individual organization listing page and add org button

ashleyabrooks:
Removes the `Edit` button from the organization listing page so it can't be edited by external users. BTW - The reason I have so many commits attempting to fix tests is my local TestCafe setup isn't working for some reason and I'm trying to get this out as quickly as possible :/ Will make sure to squash commits before merging! 

**Before:**
Screen Shot 2019-10-30 at 8 17 12 PM
**After:**
Screen Shot 2019-10-30 at 8 15 31 PM

### DIFF
--- a/testcafe/editResource.js
+++ b/testcafe/editResource.js
@@ -55,6 +55,8 @@ test('Edit resource address', async t => {
       (_t, prop) => _t.expect(resourcePage.address.textContent).contains(newProps[prop]),
       t,
     );
+  
+  await t.navigateTo(editResourcePage.url(1))
 
   // Check visibility of edits on edit page
   await Object.keys(newProps).reduce(
--- a/testcafe/editResource.js
+++ b/testcafe/editResource.js
@@ -90,12 +90,13 @@ test('Add resource phone number', async t => {
   const newServiceType = 'Added number';
 
   // Wait for page to load before counting phone numbers by using hover action.
-  await t
-    .navigateTo(editResourcePage.url(1))
+  await t.hover(resourcePage.phones);
   const originalCount = await resourcePage.phones.with({ boundTestRun: t }).count;
 
   // Make edits
-  await t.click(editResourcePage.addPhoneButton);
+  await t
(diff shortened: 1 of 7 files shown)


testcafe/editResource.js
```diff
@@ -11,7 +11,6 @@ fixture`Edit Resource`
 
 
 async function testEditTextProperty(t, showPageSelector, editPageSelector, newValue) {
-  await resourcePage.clickEditButton(t);
```
richardxia: If we're getting rid of the edit button, then you'll still need to navigate to the edit page if you want these tests to pass. You'll want to use the [`navigateTo()`](https://devexpress.github.io/testcafe/documentation/test-api/actions/navigate.html) action.

ashleyabrooks: Thanks for the tip! That makes sense. I tried this and I'm getting Test Cafe DNS (cannot resolve url) errors locally. I googled the error and it looks like an issue other people run into with test cafe on localhost so I'm pushing up the changes to see if Travis runs into it too. If it does, have you seen this error before?

testcafe/service.js
```diff
@@ -61,41 +57,40 @@ test('Confirm Page Loads with Information', async t => {
...
 
-  await modifySheduleTime(t);
+  await t.navigateTo(servicePage.url(serviceId));
+
+  await modifySheduleTime(t, 'add');
 
   await t
-    .navigateTo(servicePage.url(serviceId))
-    .hover(servicePage.editButton)
+    .navigateTo(editResourcePage.url(1))
```
richardxia: I finally got a chance to take a look at the error you were running into, @ashleyabrooks, and I think this is the bug. Previously we were looking at the (non-edit) Service Page to get the count of schedule days, but this line changed it to go to the Edit Page instead.

I'm not sure why the `.hover(servicePage.editButton)` was there (might have been to force a page scroll), but it wasn't actually clicking on the button, so we should still be on the Service Page after this action.

Let me submit a PR to fix this and unskip the tests.

Issue #594, opened 2018-11-26
Additional service schedule day cannot be deleted


alexanderturinske:
**Steps:**
1. Navigate to a service
2. Edit the service schedule by adding a second time on a day (i.e. two monday times, etcetera)
3. Save the change
4. Edit the service schedule by deleting the second time on a day (i.e. click on the red x next to the day)
  - EXPECTED: On save,  the day does disappear from the schedule on the service page
  - ACTUAL: On save, the day does NOT disappear from the schedule on the service page

So I wrote a test to ensure deleting a schedule day (i.e. for a service there is a schedule, say a user adds a second time of availability on Tuesday, now I want to delete it) works, and it broke. Looking at the code, it looks like our api supports change requests on schedule days (i.e. change the times of open/close), but I am not sure what I should send to the change request for a delete. Currently the code just changes all the times to `null` when a user tries to delete the time, but the row does not get deleted from the db.
Is this by design? Does our api support deleting a schedule day (i.e. if there are two Tuesday times, can I delete one)? If so, what is the api request to make to achieve this? I am digging around in the ruby now, but if someone knows, that would be rad.

Relates #441 
Relates #593 

@lgarofalo @jjfreund

azhelle16:
I'm unable to reproduce this issue.

PR #897, opened 2019-11-17
TestCafe: Fix Add Service Schedule Day test.

richardxia:
See https://github.com/ShelterTechSF/askdarcel-web/pull/895/files#r347166009, but it looks like the reason that the TestCafe test was failing for us is because we accidentally navigated to the wrong page.

I also added in some changes to refactor the `t.hover()` into a method with a name that more obviously describes what it is doing, which is to wait for the page to load.

### DIFF
--- a/testcafe/pages/ServicePage.js
+++ b/testcafe/pages/ServicePage.js
@@ -20,4 +20,17 @@ export default class ServicePage {
     this.schedule = baseSelector.findReact('TableOfOpeningTimes tbody tr');
     this.url = serviceId => `${config.baseUrl}/services/${serviceId}`;
   }
+
+  /**
+   * Wait until the page is fully loaded.
+   *
+   * Some TestCafe actions may attempt to run before a page is fully loaded, so
+   * this method can be called to force us to wait until the page is loaded
+   * first.
+   *
+   * @param t - A TestCafe test Promise.
+   */
+  async waitUntilPageLoaded(t) {
+    await t.hover(this.name);
+  }
 }
--- a/testcafe/service.js
+++ b/testcafe/service.js
@@ -57,9 +57,8 @@ test('Confirm Page Loads with Information', async t => {
     .ok();
 });
 
-test.skip('Confirm Service Schedule Day Can Be Added', async t => {
-  // Wait for page to load before counting services by using hover action.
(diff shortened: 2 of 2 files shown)


testcafe/pages/ServicePage.js
```diff
@@ -20,4 +20,17 @@ export default class ServicePage {
...
+   * Wait until the page is fully loaded.
+   *
+   * Some TestCafe actions may attempt to run before a page is fully loaded, so
+   * this method can be called to force us to wait until the page is loaded
+   * first.
+   *
+   * @param t - A TestCafe test Promise.
+   */
+  async waitUntilPageLoaded(t) {
+    await t.hover(this.name);
```
ashleyabrooks: So much better

ashleyabrooks: I think we use `t.hover` in a few other test modules but we can update that as needed

testcafe/service.js
```diff
@@ -68,29 +67,30 @@ test.skip('Confirm Service Schedule Day Can Be Added', async t => {
...
+  await t.navigateTo(servicePage.url(serviceId));
+  await servicePage.waitUntilPageLoaded(t);
   await t
-    .navigateTo(editResourcePage.url(1))
     .expect(servicePage.schedule.count)
     .eql(originalScheduleDayCount + 1);
 });
 
 // TODO: Uncomment with the completion of ISSUE #594
 test.skip('Confirm Service Schedule Day Can Be Deleted', async t => {
```
ashleyabrooks: Can this test be un-skipped now too?

testcafe/pages/ServicePage.js
richardxia: Yeah, we can update them as we go. We also might want to upgrade the version of TestCafe that we're using, since I think they have some of the bugs with not properly waiting for elements to appear before acting on them fixed.

testcafe/service.js
richardxia: Unfortunately, no. I tried uncommenting it, but it looks like that test is still broken for other reasons. I think #594 captures getting that bug fixed.

richardxia:
Thanks for reviewing! I'll merge this in now.

PR #900, opened 2019-11-28
Add social media meta tags

richardxia:
This adds some extra social media meta tags so that when we share links to the site, more info appears in the preview boxes.

Note that I also migrated us from react-helmet to react-helmet-async, since the former appears to be unmaintained, and the latter is a fork by the New York Times that is more up to date with newer React features. In particular, I wanted the ability for React Fragments to work inside of the `<Helmet>` tag.

I can't really fully test this locally, since to truly test it, I'd want to see what the preview links on Twitter/Facebook/etc. look like, but if we get this merged in to master, we'll be able to test it out on staging. I did at least check that the meta tags looked like they render properly locally, including with the conditional logic for the SF Service Guide branded version of the site.


derekfidler:
i don't see that the new image was added in the diff @richardxia

richardxia:
Sorry, I missed that you wanted that image here. Should it be displayed for both the SF Service Guide and the AskDarcel site, or do you want different ones for each?

richardxia:
I pushed up a new commit that adds the image for both variants of the site. Let me know if you want that changed.

derekfidler:
thanks yeah that image works for both sites

PR #908, opened 2020-01-06
Fixed issue for making phone numbers clickable

azhelle16:
Also made a minor change with a heading

### DIFF
--- a/app/components/listing/ResourceInfos.jsx
+++ b/app/components/listing/ResourceInfos.jsx
@@ -89,7 +89,7 @@ AddressInfo.propTypes = {
 
 const PhoneNumber = ({ phones }) => (
   <span className="phone">
-    { phones.map(phone => <p key={phone.id}><a href={`tel:${phone.number}`}>{`${phone.number} ${phone.service_type}`}</a></p>) }
+    { phones.map(phone => <p key={phone.id}><a href={`tel:${phone.number}`}>{`${phone.number}`}</a>{` ${phone.service_type}`}</p>) }
   </span>
 );
 
--- a/app/components/listing/ResourceInfos.jsx
+++ b/app/components/listing/ResourceInfos.jsx
@@ -89,7 +89,7 @@ AddressInfo.propTypes = {
 
 const PhoneNumber = ({ phones }) => (
   <span className="phone">
-    { phones.map(phone => <p key={phone.id}>{`${phone.number} ${phone.service_type}`}</p>) }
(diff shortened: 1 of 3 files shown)


app/components/listing/ResourceInfos.jsx
```diff
@@ -89,7 +89,7 @@ AddressInfo.propTypes = {
 
 const PhoneNumber = ({ phones }) => (
   <span className="phone">
-    { phones.map(phone => <p key={phone.id}>{`${phone.number} ${phone.service_type}`}</p>) }
+    { phones.map(phone => <p key={phone.id}><a href={`tel:${phone.number}`}>{`${phone.number} ${phone.service_type}`}</a></p>) }
```
richardxia: Minor suggestion: I think I'd probably make just the phone number itself clickable, not the service type as well, since I feel like that's usually how websites present clickable phone numbers.

azhelle16: huh, never thought of that! No problem though let me fix that :) Can I still use the same branch and same pull request?

app/components/listing/ResourceInfos.jsx
```diff
@@ -89,7 +89,7 @@ AddressInfo.propTypes = {
 
 const PhoneNumber = ({ phones }) => (
   <span className="phone">
-    { phones.map(phone => <p key={phone.id}>{`${phone.number} ${phone.service_type}`}</p>) }
+    { phones.map(phone => <p key={phone.id}><a href={`tel:${phone.number}`}>{`${phone.number}`}</a>{` ${phone.service_type}`}</p>) }
```
richardxia: Thanks! I have one last code change suggestion. The (` `` `) syntax in JavaScript (known as [template literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals)) allows you to define a string where you interpolate the value of a variable into the string. It doesn't make much sense to use it if the only in the string is the variable itself, so it'd be simpler to just return the variable in those cases. Specifically, for this line of code, I would suggest writing it like the following:

```suggestion
    { phones.map(phone => <p key={phone.id}><a href={`tel:${phone.number}`}>{phone.number}</a>{` ${phone.service_type}`}</p>) }
```

richardxia:
Thanks for addressing my comments and getting this PR passing Travis! I'll merge this in.

azhelle16:
Yay. Welcome. I'll start to look for other issues to fix. :)

PR #911, opened 2020-01-06
Refactor EditAddress state

richardxia:
As mentioned in https://github.com/ShelterTechSF/askdarcel-web/pull/889#issuecomment-570960935, this is my attempt at refactoring the state management on the Edit page with respect to addresses to simplify the supporting of multiple addresses.

This is branched from https://github.com/ShelterTechSF/askdarcel-web/pull/894, so the diff will be messy and include the stuff from that PR until that PR merges in to master. In the meantime, you can look at the following diff view to see what was added in this PR: https://github.com/ShelterTechSF/askdarcel-web/compare/akalkanis_885_resource_page_multiple_addresses...refactor-edit-address-state

Basically what I tried to do was remove all React state from the `<EditAddress>` component, and hoist whatever necessary state into the OrganizationEditPage itself. Previously there were four (!) different places that held state related to the address, and updates to them were very difficult to keep track of, not to mention one of them violated the React rule of not directly mutating the state object:

1. `OrganizationEditPage.state.resource.addresses[0]`: This was the state of the address as it came from the initial API request. However, due to this line, we were previously directly mutating it: https://github.com/ShelterTechSF/askdarcel-web/blob/5a85d49fd50ae2e731cc26f9fea34de5717a5733/app/components/edit/EditAddress.jsx#L28

    This is [not allowed by React](https://reactjs.org/docs/react-component.html#state), and it could lead to undefined behavior, since React does not know that we are modifying its value that we. We were unfortunately relying on this undefined behavior.

2. `OrganizationEditPage.state.address`: This usually represented any modifications to the address, since most of the time we didn't directly modify `this.state.resource.addresses[0]`.

3. `EditAddress.state.address`: This was the original EditAddress component's state of the address, and it wasn't always clear how it related to both of the other two states above. In addition, `<EditAddress>` also contained an extra `hasNoLocation` boolean state variable that was not present in `OrganizationEditPage` at all.

4. The actual `<input>` elements in `EditAddress`, since they are [uncontrolled components](https://reactjs.org/docs/uncontrolled-components.html). This mean that they kept track of their own state independent of the components that contain them.

I think I managed to collapse these all into really two places:
1. `OrganizationEditPage.state.resource.addresses[0]`, which really should be immutable now, and
2. `OrganizationEditPage.state.address`/`OrganizationEditPage.state.hasLocation`, which is expected to mutate as a user modifies fields.

There are a bunch of tricky edge cases to keep in mind, and I have a very long comment in the `handleSubmit()` method that describes 7 scenarios, 2 of which I am just throwing an exception for until we actually handle multiple addresses and 0 addresses.

Just note that this does not actually do anything to support multiple addresses, but I think by getting rid of the redundant state, it'll make it easier to actually implement that feature because we won't have to carefully keep four different representations of the same state in sync.

### DIFF
--- a/app/components/edit/EditAddress.jsx
+++ b/app/components/edit/EditAddress.jsx
@@ -1,92 +1,45 @@
-import React, { Component } from 'react';
+import React from 'react';
+import PropTypes from 'prop-types';
 
-const hasNoLocation = address => {
-  const someFieldExistsOrNewAddress = typeof address === 'undefined'
-    || address.name !== '' || address.address_1 !== ''
-    || address.address_2 !== '' || address.address_3 !== ''
-    || address.address_4 !== '' || address.city !== '' || address.postal_code !== ''
-    || address.state_province !== '' || address.country !== '';
-  if (someFieldExistsOrNewAddress) {
-    return false;
-  }
-  return true;
+const EditAddress = ({ address, setAddress, setHasLocation }) => (
+  <AddressForm
+    setAddress={setAddress}
+    setHasLocation={setHasLocation}
+    address={address}
+  />
+);
+
+EditAddress.propTypes = {
+  address: PropTypes.object,
+  setAddress: PropTypes.func.isRequired,
+  setHasLocation: PropTypes.func.isRequired,
 };
 
-class EditAddress extends Component {
-  constructor(props) {
-    super(props);
-    this.state = { address: {}, noLocation: hasNoLocation(props.address) };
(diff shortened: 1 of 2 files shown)


richardxia:
Removing WIP since this is now passing Travis CI, now that I've fixed up the new organization page.

richardxia:
I'm going to merge this in and start working on the next step of updating the actual edit page to support multiple addresses.

Issue #867, opened 2019-09-22
Add Toggle Map View button (mobile)


Maxastuart:
On the Search Results page, when in mobile view:

- add a button to the filters area to toggle Map View on and off (so users can see the search results from their page on the map instead of on a list

- make sure the map view interaction is a decent mobile experience, and filters are still accessible and returning to the list view is easy

(I believe this is a function for mobile or tablet - whenever the screen size is small enough not to display the map on the side by default as it does on desktop. This button/function should be hidden in desktop view)

Maxastuart:
Related #700 #868

PR #914, opened 2020-01-27
Add Map Button in Mobile View

rajsudeep:
As requested in #867, provides a map button in the mobile view (window width <= 767) that when is toggled, displays the google map within the search list container.

### DIFF
--- a/app/components/search/SearchResultsContainer.jsx
+++ b/app/components/search/SearchResultsContainer.jsx
@@ -47,7 +47,7 @@ function useWindowDimensions() {
 const SearchResultsContainer = ({ searchState, searchResults, searching }) => {
   const [getMapButton, setMapButton] = useState(false);
   const { width } = useWindowDimensions();
-  const isMobile = (width <= 767) ? true : false;
+  const isMobile = (width <= 767);
   let output = null;
   if (!searchResults && searching) {
     output = <Loader />;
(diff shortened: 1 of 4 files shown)


app/components/search/SearchResultsContainer.jsx
```diff
@@ -49,13 +70,27 @@ No results have been found for
...
+              isMobile
+            />
+          ) : (
+            <SearchTable
+              hits={hits}
+              page={searchResults.page}
+              hitsPerPage={searchResults.hitsPerPage}
+            />
+          )
+          }
```
richardxia: I think this conditional will completely go away if you make more of the media queries in CSS, like I recommended, but I wanted to point out a general tip about stuff like this. Since the result of the conditional is to just set or unset a single prop on the component, you don't actually need the ternary expression around it and you can instead just pass a boolean value to `isMobile`. For example, this could just be written as:

```jsx
<SearchMap
  hits={hits}
  page={searchResults.page}
  hitsPerPage={searchResults.hitsPerPage}
  isMobile={getMapButton && isMobile}
/>
```

rajsudeep: I see, thank you for the detailed input! I'll take a stab at implementing this using media queries if I get the time. Until then I'll leave this ticket open again on Clubhouse in case anyone else is able to get the job done before me. I'll be sure to contact you regarding any questions I have on implementing this. Cheers!

richardxia: Cool, thanks! Let me know if you ever wanted to spend a bit of time chatting about this and I'd be happy to help out.

PR #915, opened 2020-01-29
Fix New Org Creation

AnnaKalkanis:
See @Maxastuart 's report in channel for context: https://sheltertech.slack.com/archives/C0EGCTDCL/p1580195509055400

There were a few issues here. 
1) The first was that the API was throwing a 500 on submission because we were trying to send a resource with an `addresses` field and it only knows how to handle POSTs with an `address` field. 
2) Next, we weren't rendering an hours template for the new resource because we were generating the temporary schedule object and adding it to the component state nested in the `resource` object. It should be directly on component state.
3) Finally, the API errors if you send a schedule day with a null `opens_at` or `closes_at`. So, had to add some additional cleaning of the schedule before actually sending it up to remove any schedule days that had null times

### BEFORE
Screen Shot 2020-01-28 at 11 01 44 PM

### AFTER
Screen Shot 2020-01-28 at 11 02 02 PM

In addition to the functional changes necessary above, I also tweaked some styling on the checkboxes.

### BEFORE
Screen Shot 2020-01-28 at 11 00 09 PM
Screen Shot 2020-01-28 at 11 00 15 PM

### AFTER
Screen Shot 2020-01-28 at 10 59 38 PM
Screen Shot 2020-01-28 at 10 59 03 PM

### DIFF
--- a/app/components/edit/EditAddress.jsx
+++ b/app/components/edit/EditAddress.jsx
@@ -77,7 +77,7 @@ const AddressForm = ({
 }) => (
   <li key="address" className="edit--section--list--item">
     <label htmlFor="address">Address</label>
-    <label>
+    <label className="inline-checkbox">
       <input
         type="checkbox"
         className="input-checkbox"
--- a/app/components/edit/EditSchedule.jsx
+++ b/app/components/edit/EditSchedule.jsx
@@ -31,13 +31,16 @@ class EditSchedule extends Component {
         { canInheritFromParent
           && (
             <div className="inherit-schedule">
-              <input
-                id="inherit"
-                type="checkbox"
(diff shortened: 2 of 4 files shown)


richardxia:
> The first was that the API was throwing a 500 on submission because we were trying to send a resource with an addresses field and it only knows how to handle POSTs with an address field.

I think this might have only been failing for you locally, which would be the case if you hadn't pulled the latest `master` of askdarcel-api. We just merged in the multiple addresses changes, so the backend does expect `addresses` instead of `address` now, which I think is why your Travis CI build failed. I'm going to try reverting the change related to `addresses` to see if that passes.

PR #919, opened 2020-02-05
update travis to not push images for all branch builds

lgarofalo:


lgarofalo:
This PR caused dockerhub to push an image to latest, which is a bug, looking at that.

PR #920, opened 2020-02-16
Add more error checking for Travis CI failures and fix broken Travis Postgres error

richardxia:
Trying to debug why Travis CI tests are all currently failing.


richardxia:
Looks like the Docker postgres image changed its behavior and now requires you to explicitly set a password via environment variables. Fixing this in a sec.

PR #922, opened 2020-03-22
add temporary covid banner

AnnaKalkanis:
Temp banner on homepage:
Screen Shot 2020-03-22 at 3 43 13 PM

Banner will show up on all pages:
Screen Shot 2020-03-22 at 3 43 24 PM


app/components/ui/Banner/Banner.jsx
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+import styles from './Banner.scss';
+
+export default function Banner() {
+  return (
+    <div className={styles.bannerContainer}>
+      <strong>CORONAVIRUS (COVID-19) UPDATE: </strong>
+      <span>
```
richardxia: Why does this `span` need to exist? I feel like this shouldn't be necessary.

AnnaKalkanis: oops ya you're right! I'll go ahead and merge this in case we want to get this temporary one out tonight and will clean it up once I get actual designs from derek tomorrow.

PR #925, opened 2020-03-29
Update banner copy

AnnaKalkanis:
See conversation here: https://sheltertech.slack.com/archives/C010E72NA0Z/p1585426146299600

Laura asked that we update the banner copy so that it does not say that this resource guide is ours.

Screen Shot 2020-03-28 at 7 53 27 PM


drcaramelsyrup:
LGTM!

PR #927, opened 2020-04-01
Updated guide cards, covid page for covid launch

JacobDFrank:
- Increased accesibilty for the cards by adding more visible buttons and a border
- Made the banner and font sizes on the covid page smaller on mobile
- removed the intercom logo on mobile
- added padding to the google doc embed

### DIFF
--- a/app/pages/Covid/Covid.jsx
+++ b/app/pages/Covid/Covid.jsx
@@ -1,6 +1,6 @@
 import React from 'react';
-import Footer from '../../components/ui/Footer/Footer';
 import Iframe from 'react-iframe';
+import Footer from '../../components/ui/Footer/Footer';
 
 import styles from './Covid.scss';
 
@@ -16,10 +16,14 @@ export default class Covid extends React.Component {
         <article className={styles.covidPage} id="covid">
           <header className={styles.covidHeader}>
             <h1>
-            Coronavirus resources for Individuals-In-Need            
+            Coronavirus resources for Individuals-In-Need
             </h1>
             <p>
-              This list is compiled to link individuals experiencing homelessness to resources.  We will be updating the list continuously, when services change. Please email <a href="mailto:contact@sheltertech.org">contact@sheltertech.org</a> to suggest edits.
+              This list is compiled to link individuals experiencing homelessness to resources.  We will be updating the list continuously, when services change. Please email ]
+              <a href="mailto:contact@sheltertech.org">
+                contact@sheltertech.org
+              </a>
+                to suggest edits.
             </p>
           </header>
           <section>
--- a/app/pages/Covid/Covid.jsx
+++ b/app/pages/Covid/Covid.jsx
@@ -19,10 +19,10 @@ export default class Covid extends React.Component {
             Coronavirus resources for Individuals-In-Need            
             </h1>
             <p>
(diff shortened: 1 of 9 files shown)


app/pages/Covid/Covid.jsx
```diff
@@ -1,5 +1,6 @@
 import React from 'react';
 import Footer from '../../components/ui/Footer/Footer';
+import Iframe from 'react-iframe';
```
AnnaKalkanis, line 3: I'm not familiar with this library. Is there a reason we need it here? Was the regular iframe not working?

app/pages/Covid/Covid.jsx
```diff
@@ -15,30 +16,19 @@ export default class Covid extends React.Component {
...
-              of resources for individuals experiencing homelessness. We will be updating the list continuously, when services change, to ensure you have the most up to date information.
-              Please email
-              {' '}
-              <a href="mailto:contact@sheltertech.org">contact@sheltertech.org</a>
-              {' '}
-              to suggest edits.
-              See what types of places are open or closed at
-              {' '}
-              <a href="sf.gov/stay-home-except-essential-needs">sf.gov/stay-home-except-essential-needs</a>
+              This list is compiled to link individuals experiencing homelessness to resources.  We will be updating the list continuously, when services change. Please email <a href="contact@sheltertech.org">contact@sheltertech.org</a> to suggest edits.
```
AnnaKalkanis, line 22: Might be worth throwing a `mailto:` on here so this is recognized as an email address. ie `href="mailto:contact@sheltertech.org"`

app/pages/Covid/Covid.jsx
```diff
@@ -15,30 +16,19 @@ export default class Covid extends React.Component {
...
             </p>
           </header>
-          <section>
-            <iframe
+          <section className={styles.aboutSection}>
```
AnnaKalkanis, line 25: don't think we need this class here - I'm not seeing it defined in the css

AnnaKalkanis:
Definitely looks much better when resized! And I like the borders on the cards. The biggest thing for me is that it would be nice to avoid adding the react-iframe dependency if we don't need it.

AnnaKalkanis:
You've also got some failing tests. Looks like just linting. You can see what they are by clicking the `details` for `CodeFactor` and `continuous-integration/travis-ci/pr `

app/pages/Covid/Covid.jsx
JacobDFrank: It allows us to easily strip out the CSS of a google chrome embed and allow it to be more readable. There may be some better workarounds but I wanted to get this out sooner.

app/pages/Covid/Covid.jsx
JacobDFrank: good idea, i'll add that

app/pages/Covid/Covid.jsx
JacobDFrank: you're right, I accidentally left that in

JacobDFrank: I accidentaly added it.

app/pages/Covid/Covid.jsx
AnnaKalkanis: ah ok cool, sounds good!

PR #930, opened 2020-04-02
Made more mobile friendly

JacobDFrank:
added blue border to embed and made width smaller


JacobDFrank:
Add a blue outline to better view the embed

PR #932, opened 2020-04-06
Remove empty modules.

richardxia:
My [bugfix to eslint-plugin-import](https://github.com/benmosher/eslint-plugin-import/pull/1640) finally made it in the latest release, so I want to make use of it and detect a bunch of unused imports.

The first set of things I found I wasn't even really expecting, which is that we have some completely empty JavaScript files lying around. This PR deletes them (and fixes some code that was somehow importing one of them).


richardxia:
I'm going to just merge this in since I think it's a pretty safe change, as the empty files are clearly not actually being used.

PR #935, opened 2020-04-12
Covid food page

JacobDFrank:


app/pages/HomePage/components/GuideList/GuideList.jsx
```diff
@@ -69,19 +69,30 @@ const GuideList = () => (
         </a>
       </li>
       <li className={styles.item}>
+        <a
+          href="/FoodMap"
+          // target="_blank"
```
ruochenhuang, line 74: @JacobDFrank @richardxia can we also have the food pathway page open into a new tab in addition to the /covid page? Thank you!

app/components/ui/Banner/Banner.jsx
```diff
@@ -4,12 +4,10 @@ import styles from './Banner.scss';
...
-      <strong>CORONAVIRUS COVID-19: </strong>
-      Many organizations and services have reduced hours and availability. See our new
-      {' '}
-      <a className={styles.bannerLink} href="/covid">Resource Guide</a>
-      {' '}
-      - updated daily.
+      <span>
+        <strong>CORONAVIRUS UPDATE: </strong>
+        Many organizations and services have reduced hours and availability.
+      </span>
```
AnnaKalkanis, line 10: Oh are we removing the link to the resource guide here?

richardxia:
@JacobDFrank, it looks like you had started this branch from an older commit instead of the latest `master` branch, and as a result, you have merge conflicts (possibly with some of your own changes). The branch history looks like this:

```
* commit 5a739bb78fa1ab234c5fde220f247626be0ba1f5 (origin/covid-food-page)
| Author: Jacob <jfrank@adobe.com>
|
|     Draft of foodmap and foodlist pages, removed non-covid pathways
|
* commit 02d4416c681073b8ae287220d7ed78a4e556b781
| Author: Jacob <jfrank@adobe.com>
|
|     Added food page map embed jsx and scss
|
* commit e953ae4b5e02d2041fa78786209508a5cf18d221
| Author: Jacob <jfrank@adobe.com>
|
|     Made covid links open in a new tab
|
| * commit 5bc22fc6e017f872c106f9bb8f2eefdc49108699 (origin/master, origin/HEAD)
| | Author: Jacob Frank <frankjacob42@gmail.com>
| |
| |     Made covid links open in a new tab (#934)
| |
| * commit a69a3eb2d64d0fa6895919b8ce7d460cc6a1b5b8
|/  Author: Richard Xia <richardxia@richardxia.com>
|
|       Edit page CSS updates (#931)
|
* commit 2424705b19ff6b1cebdac90d12e36b4ef50c6561
| Author: Jacob Frank <frankjacob42@gmail.com>
|
|     Made more mobile friendly (#930)
|
```

Can you resolve the conflicts either by merging the latest master branch into this branch or rebasing this branch on top of master? You could also try the "Resolve conflicts" button from the GitHub UI, if you feel more comfortable with that.

---

Also, in general, I see that you have commented out a lot of stuff. We usually prefer to outright delete code rather than just comment it out, since the problem with the latter is that the commented out code accumulates over time and no one will remember why it's commented out or whether it's safe to uncomment. It's also less clear to reviewers whether the commented out code was an intentional change to be merged in or whether you meant to uncomment them before merging, so by deleting the code you more clearly communicate your intent of deleting the code.

---

Let me know if you want help with any of these, or if you just want me to go make some of this changes directly on this branch. I wasn't sure if you were still working on this and have unpushed changes, so I don't want to touch this branch. I should be mostly free for the remainder of the day, so just let me know when you want to hand this off for merging and getting it into staging.

app/components/ui/Banner/Banner.scss
```diff
@@ -2,16 +2,25 @@
 
 .bannerContainer {
   width: 100%;
-  padding: calc-em(20px) calc-em(30px);
+  padding: calc-em(16px) calc-em(28px);
   color: black;
   background-color: $color-yellow;
   font-size: calc-em(24px);
   text-align: center;
+  position: sticky;
```
AnnaKalkanis, line 10: Are you trying to get the banner to always be present at the top of the page & have the rest of the page scroll under it?

app/components/ui/Banner/Banner.scss
```diff
@@ -2,16 +2,25 @@
...
   strong {
     font-weight: bold;
   }
-  @media screen and (max-width: 800px) {
-    font-size: 16px;
+  @media screen and (max-width: 600px) {
+    font-size: calc-em(16px);
+    height: 120px;
+    line-height: 120px;
+    padding: 0 calc-em(28px);
```
AnnaKalkanis, line 23: I'm not sure that the styling on the span or anything in this media query (other than the font-size) is actually necessary. I wouldn't add it if we don't need it.

app/pages/HomePage/components/GuideList/GuideList.jsx
JacobDFrank: I commented that out because we haven't fixed the modal opening problem yet, it was discussed here. https://sheltertech.slack.com/archives/C0EGCTDCL/p1586646425205700

app/components/ui/Banner/Banner.jsx
JacobDFrank: Yes

app/components/ui/Banner/Banner.scss
AnnaKalkanis: Oh i see now, you were doing this so that you could specify the banner height. I'm gonna mess with this a bit and see if I can figure out a less brittle way to do this

app/components/ui/Navigation/Navigation.jsx
```diff
@@ -71,11 +71,11 @@ class Navigation extends React.Component {
             <button type="button" className={styles.hamburgerButton} onClick={toggleHamburgerMenu} />
           </div>
           <ul className={styles.navRight}>
-            <li>
+            {/* <li>
               <Link to="/covid">
                 Coronavirus
               </Link>
-            </li>
+            </li> */}
```
AnnaKalkanis, line 78: If we're removing this, I say just delete this.

app/pages/HomePage/components/GuideList/GuideList.jsx
ruochenhuang: Thanks @JacobDFrank! @AnnaKalkanis are you able to modify the GuideCard component to identify whether the link is a typeform or not? Want to have any non-typeform pages open a new tab on click from the homepage for google analytics tracking purposes. Thank you!

app/components/ui/GDocEmbed/GDocEmbed.jsx
```diff
@@ -0,0 +1,59 @@
...
+import Iframe from 'react-iframe';
+import Footer from '../Footer/Footer';
+
+import styles from './GDocEmbed.scss';
+
+
+export default class GDocEmbed extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {
```
richardxia, line 12: Defining an empty state is a bit misleading and unnecessary. If the component is stateless, then you don't need to declare this field. In fact, I would prefer that stateless components are declared as [function components](https://reactjs.org/docs/components-and-props.html#function-and-class-components), which are much simpler.

In this case, you can completely drop the `class` ceremony and define a function that is just the body of the `render()` method and call it a day. It'd look something like:

```jsx
const GDocEmbed = ({ pageHeader, GDocURL }) => (
  <React.Fragment>
    <article className={styles.GPage}>
      ...
);
```

app/components/ui/GDocEmbed/GDocEmbed.jsx
```diff
@@ -0,0 +1,59 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import Iframe from 'react-iframe';
+import Footer from '../Footer/Footer';
+
+import styles from './GDocEmbed.scss';
+
+
+export default class GDocEmbed extends React.Component {
```
richardxia, line 9: Nit on naming (which admittedly are my personal style preferences): I'd prefer to spell out GDoc as GoogleDoc, since it's clearer when spelled out all the way, and it's not even an official abbreviation that Google uses in public materials.

app/components/ui/GDocEmbed/GDocEmbed.jsx
```diff
@@ -0,0 +1,59 @@
...
+            <h1>
+              {pageHeader}
+            </h1>
+            <p>
+              This list is compiled to link individuals experiencing homelessness to resources.
+              We will be updating the list continuously, when services change.
+              <br />
+              Please email
+              <a href="mailto:contact@sheltertech.org">
+                {' '}
```
richardxia, line 32: I'd put the `{' '}` above and outside the `<a>` tag, since it's kind of weird to have the space character before a link to be clickable.

app/components/ui/GDocEmbed/GDocEmbed.jsx
```diff
@@ -0,0 +1,59 @@
...
+              url={GDocURL}
+              width="80%"
+              height="100%"
+              padding="0pt"
+              allowFullScreen="yes"
+              className={styles.GDocEmbed}
+            />
+          </section>
+        </article>
+        <Footer />
```
richardxia, line 51: I don't feel that this footer belongs here. The component should only be responsible for embedding a single Google Document. It should be the responsibility of the page component that includes this GDocEmbed to insert a footer.

One example of a problem with this approach is that if you wanted to include two GDocEmbeds into the same page, you'd end up getting two footers. Or, if you wanted to include a GDocEmbed on a different page, but you wanted a different, perhaps shorter footer, you'd be stuck using the same footer style on both GDocEmbeds.

app/pages/Covid/Covid.jsx
```diff
@@ -12,39 +9,12 @@ import styles from './Covid.scss';
...
-              height="100%"
-              padding="0pt"
-              allowFullScreen="yes"
-              className={styles.covidDocEmbed}
-            />
-          </section>
-        </article>
-        <Footer />
-      </div>
+      <React.Fragment>
```
richardxia, line 12: This `<React.Fragment>` is unnecessary since you only have a single child component.

app/pages/DomesticViolence/DomesticViolence.jsx
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+import GDocEmbedPage from '../../components/ui/GDocEmbed/GDocEmbed';
+
```
richardxia, line 3: I can't leave a comment on it, but the `DomesticViolence.scss` file is empty, and it's not imported here, so I'd remove it to avoid any confusion. Same goes for many of these.

app/pages/HomePage/components/GuideList/GuideList.jsx
```diff
@@ -4,12 +4,22 @@ import * as typeformEmbed from '@typeform/embed';
 
 import styles from './GuideList.scss';
 
-import ImgCovid from './assets/Covid19.jpg';
-import ImgEviction from './assets/EvictionPrevention.jpg';
-import ImgAffordableHousing from './assets/AffordableHousing.jpg';
+
+// import ImgEviction from './assets/EvictionPrevention.jpg';
+// import ImgAffordableHousing from './assets/AffordableHousing.jpg';
```
richardxia, line 9: Delete commented out code.

richardxia:
Merging this in now since we want this to show up on staging, but I'm going to take a pass at making the changes I suggested above in a separate PR.

app/pages/HomePage/components/GuideList/GuideList.jsx
AnnaKalkanis: Yup, I modified it and pushed it up on this PR. So, should be opening in a new tab now :)

PR #938, opened 2020-04-13
Refactor COVID food page

richardxia:
This PR refactors the code originally added in https://github.com/ShelterTechSF/askdarcel-web/pull/935.

At a high level, I tried to refactor the code such that similar or identical pages or components are represented with shared, common components. Most substantially, I've moved almost all of the code into the `app/pages/Covid` directory, and I've pulled out set of shared components. Here is the new directory tree:

```
Covid/
├── Covid.jsx
├── Covid.scss
├── components
│   ├── CovidHeader
│   │   ├── CovidHeader.jsx
│   │   ├── CovidHeader.scss
│   │   └── index.js
│   ├── FoodList
│   │   ├── FoodList.jsx
│   │   ├── FoodList.scss
│   │   └── index.js
│   ├── FoodMap
│   │   ├── FoodMap.jsx
│   │   ├── FoodMap.scss
│   │   ├── assets
│   │   │   └── SFGovLogo.png
│   │   └── index.js
│   ├── GoogleDocEmbed
│   │   ├── GoogleDocEmbed.jsx
│   │   ├── GoogleDocEmbed.scss
│   │   └── index.js
│   └── SimpleGuide
│       ├── SimpleGuide.jsx
│       └── index.js
└── index.js
```

Let me describe the shared components in a bit more detail:

- `CovidHeader` is the header block at the top of most of the COVID pages, which includes the page heading as well as the "This list is compiled to ..." text. If you want to update that, you can just update it in this one place
- `GoogleDocEmbed` is the component that just embeds the Google Doc into the page, including all of the parameters for the `<IFrame>` so that we don't let the parameters for that drift
- `SimpleGuide` is a page-level component that just puts the `CovidHeader` and the `GoogleDocEmbed` together. Almost all of the pages just instantiate this component and pass it the page title and the Google Doc embed URL.

There are two pages that are not shared but can still reuse the shared components:

- `FoodList` reuses the `CovidHeader` and the `GoogleDocEmbed`, but it also contains the extra snippet that links to the `FoodMap`.
- `FoodMap` is more or less the same as before and I guess doesn't actually reuse any of the other shared components.

The other big change is that I turned `Covid.jsx` into its own URL router. The original page under `Covid.jsx` has been turned into a `SimpleGUide`, and now the new `Covid.jsx` is a React Router `<Switch>` which routes between each of the sub-pages. This also contains the top-level page styling (e.g. font size) and it is also the one that includes the footer, so that none of the other components has to instantiate it.

My goal for all of these changes is to make it easier to update the individual pages without the risk of drift if we make a change to one page but forget to propagate it to all other pages. This still retains flexibility if we need to customize individual pages, since we can just follow the pattern that I used for the `<FoodList>`, which still reuses the header and Google doc embed components while being able to fully customize that page.


richardxia:
Thanks a lot for the kind feedback! I hadn't done nested routing before, but I'm glad it turned out as nicely as it did. Going to merge this in now to get this onto staging for tonight's deploy.

PR #948, opened 2020-05-18
sort services alphabetical so sidebar and page services can match

Akhtam:


app/components/edit/EditServices.jsx
```diff
@@ -8,8 +8,9 @@ const EditServices = ({
 }) => (
   <li className="edit--section--list--item">
     <ul className="edit--section--list--item--sublist edit--service--list">
-      {
-        services.map((service, index) => (
+      {services
+        .sort((a, b) => a.name - b.name)
```
richardxia, line 12: You probably don't want to subtract strings, since this mostly likely won't do what want it to do. The subtraction operator in JavaScript is only defined for numbers, so it'll attempt to cast both sides to a number before performing the subtraction operation.

Here's some examples:

```js
> "abc" - "bde"
NaN
> "3" - "2"
1
> "foo" - "42"
NaN
```

Yes, in the middle expression, those are both strings that contain only numbers and therefore get converted to actual numbers.

---

Since we use lodash, I'd recommend using the [`sortBy`](https://lodash.com/docs/#sortBy) function, which allows you to specify a property to sort by.

app/components/edit/EditSidebar.jsx
```diff
@@ -87,6 +87,12 @@ const EditSidebar = ({
       allServices[service.id].name = service.name;
     });
   }
+
+  const sortedServices = Object.entries(allServices).sort((a, b) => {
+    if (+a[0] < 0) return 0;
+    if (+b[0] < 0) return 1;
```
richardxia, line 93: Once again, this probably won't work well, since casting strings to numbers using the unary `+` operator will not work the way you want it to most of the time. I'd also recommend using lodash's `sortBy` here.

package-lock.json
```diff
@@ -340,7 +340,8 @@
             "code-point-at": {
               "version": "1.1.0",
               "bundled": true,
-              "dev": true
+              "dev": true,
```
richardxia, line 343: Since you haven't added any new packages to the project, I'd request that you revert all the changes to this file.

PR #950, opened 2020-05-21
Service Discovery Modal (1500)

AnnaKalkanis:
This PR adds a base modal, with the hope that we can use this for all our modals in the future. It will take a `isFullscreen` prop to determine whether it is displayed fullscreen or as a floating modal with an overlay. The non-fullscreen version has just basic styles, I haven't bothered fully implementing it since it is not being used yet. Here's what it looks like right now:
Screen Shot 2020-05-20 at 10 31 04 PM

Additionally, this PR makes use of that base modal fullscreen modal to implement the service discovery modal.
Screen Shot 2020-05-20 at 10 32 58 PM

This is still very much a WIP (note the hard coded category ids). But I figured it would be useful to get it merged since we'll need to fit @davidagustin's work into this.


app/pages/HomePage/components/GuideList/GuideList.jsx
```diff
@@ -82,31 +80,30 @@ const GuideList = () => (
...
-          isTypeform={false}
         />
       </li>
       <li className={styles.item}>
         <GuideCard
           name="Family Homelessness"
           link="https://sheltertech.typeform.com/to/GFEzl2"
           img={ImgFamilyHomelessness}
-          isTypeform
+          categoryId="5"
```
AnnaKalkanis, line 90: i've just hardcoded a random id in for now, until we have answers on https://sheltertech.slack.com/archives/C013NS0057A/p1590026728003000

app/components/ui/Modal/ServiceDiscoveryModal.jsx
```diff
@@ -0,0 +1,79 @@
...
+    });
+  }
+
+  handleCheckboxClick(optionId) {
+    const { checkedItems } = this.state;
+    checkedItems[optionId] = !checkedItems[optionId];
+  }
+
+  render() {
+    const { isEligibility } = this.props;
```
AnnaKalkanis, line 43: We'll pass this prop to determine whether we show the eligibility options or the subcategory options, but I still have an open question about how we determine which to use here: https://sheltertech.slack.com/archives/C013NS0057A/p1590026695002300

app/components/ui/Modal/ServiceDiscoveryModal.jsx
```diff
@@ -0,0 +1,79 @@
...
+        eligibilities,
+      });
+    });
+
+    dataService.get(`api/categories/subcategories?id=${categoryId}`).then(response => {
+      const { categories } = response;
+      this.setState({
+        subcategories: categories,
+      });
+    });
```
AnnaKalkanis, line 34: I originally implemented this so it just fetched whichever one of these is relevant to the modal (ie if we are showing eligibilities or subcategories). But then I realized that if we are using this same modal for david's view, we'll need to fetch both anyway.

app/pages/HomePage/components/GuideList/GuideList.jsx
```diff
@@ -21,25 +17,14 @@ import Imghygiene from './assets/hygiene.jpg';
...
 }) => {
-  // if this is typeform, we open the typeform modal on click.
-  // Otherwise, we just attach the link as an href.
-  const anchorTagProps = isTypeform ? {
+  const [modalOpen, setModalOpen] = useState(false);
+
+  const anchorTagProps = categoryId ? {
     role: 'button',
-    onClick: e => { typeform(e, link); },
+    onClick: () => setModalOpen(true),
```
AnnaKalkanis, line 27: I still have an open question about whether this is replacing all typeform guides or all guides. The answer to that may change this a bit. https://sheltertech.slack.com/archives/C013NS0057A/p1590026656001500

app/components/ui/Modal/Modal.jsx
```diff
@@ -0,0 +1,52 @@
...
+    </Modal>
+  );
+};
+
+BaseModal.propTypes = {
+  isOpen: PropTypes.bool.isRequired,
+  backButtonText: PropTypes.string,
+  closeModal: PropTypes.func,
+  modalContent: PropTypes.element,
+  isFullScreen: PropTypes.bool,
```
richardxia, line 44: Even if this isn't fully implemented yet, could you add some documentation on what the `isFullScreen` prop does? Your PR description of this was helpful for me, since I actually would have otherwise thought that `isFullScreen` would have referred to the "floating modal with overlay", since it almost cover the full screen. This is in contrast to modals I've seen on other sites which occupy a much smaller percentage of the screen.

Also, on mobile, do we always want to go full screen? If so, will setting the `isFullScreen` property have any effect?

app/pages/HomePage/components/GuideList/GuideList.jsx
AnnaKalkanis: turns out that this is not replacing any of the typeform guides. Those will remain as-is and we'll be creating new categories for this new flow. I've put all the typeform links back in and removed the hardcoded categoryIds for now. I'll hook up the right categories once they are available on the backend.

app/components/ui/Modal/Modal.jsx
AnnaKalkanis: Ah ya, I'm looking at some of the designs that use that standard floating modal and it definitely doesn't take up as much of the screen as I have right now. Think I'm gonna just leave it as-is for now and we can style it better once we actually use it for the edit address modal. That's a good point about it transitioning to fullscreen on mobile. We don't have any mobile designs for this view ([figma here](https://www.figma.com/file/tBRClGsxYN9m6ZvL2STB6t/WIP-Edit-Redesign---Unscheduled?node-id=2003%3A2581)) but I'll check in with design when I implement it.

Screen Shot 2020-05-31 at 2 32 41 PM

I'll add some documentation to the `isFullscreen` prop!

richardxia:
Looks like merging with master fixed it 🤷

Anyway, LGTM once this passes!

PR #951, opened 2020-05-31
Dummy commit to see if TestCafe tests are passing on master

richardxia:
I'm having a bit of trouble figuring out why https://github.com/ShelterTechSF/askdarcel-web/pull/950 is failing, so I'm creating a PR against master with a dummy commit editing only the README to see if the tests are passing here.


richardxia:
Hmm, that passed fine. Let me try just pushing up #950's latest commit to this branch to see if that works.

PR #958, opened 2020-06-16
Refactor RefinementListFilter to reuse existing Algolia functionality.

richardxia:
I was trying to debug some of the issues that we were seeing in prod when I noticed that there was a lot of stuff that we were doing in RefinementListFilter that's unnecessary, since the Algolia helpers already do a lot of that work. I redid the component to look more like [the example in the Algolia docs](https://www.algolia.com/doc/api-reference/widgets/refinement-list/react/#create-and-instantiate-your-connected-widget), where we don't define any custom props, since Algolia already has them.

I think this simplifies the component because there's no longer any state, as all the state is derived from the Algolia higher-order component helper that ties it to the context created by the containing `<InstantSearch>` component.

One of the more notable difference is that now the `<RefinementListFilter>` is a pretty dumb component, and the act of filtering down the selectable subcategories is in the instantiation of the `<RefinementListFilter>` in ServiceDiscoveryResults.jsx, via the [`transformItems`](https://www.algolia.com/doc/api-reference/widgets/refinement-list/react/#connector-param-exposed-transformitems) prop, which is the official API for filtering, sorting, or otherwise altering the filter options.

I'm not sure if this actually fixes things in prod, but I can confirm that at least locally, the Food pathway works for me.

### DIFF
--- a/app/pages/ServiceDiscoveryResults/RefinementListFilter.jsx
+++ b/app/pages/ServiceDiscoveryResults/RefinementListFilter.jsx
@@ -1,86 +1,36 @@
-import React, { Component } from 'react';
+import React from 'react';
 import PropTypes from 'prop-types';
 import { connectRefinementList } from 'react-instantsearch/connectors';
 import styles from './ServiceDiscoveryResults.scss';
 
-class RefinementListFilter extends Component {
-  static propTypes = {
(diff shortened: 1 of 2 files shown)


AnnaKalkanis:
+1 thanks for cleaning this up!

PR #961, opened 2020-06-22
Give service discovery forms and results their own pages.

richardxia:
This included some refactoring to make the ServiceDiscoveryResults component its own standalone react-router route.

## Design
Here's what it looks like (note the address bar of my browser):

Screen Shot 2020-06-21 at 9 48 01 PM

I could definitely use some more guidance on what the URLs should be (@JacobDFrank?). For now, it's just at `/service-discovery-results/:category_id`, but it causes it to have a kind of goofy URL because we started the category ID at 1,000,001.

## Technical details

I learned how to properly sync the Algolia InstantSearch state with the browser URL and react-router, which is covered in [a specific document in the Algolia docs](https://www.algolia.com/doc/guides/building-search-ui/going-further/routing-urls/react/). Their example made use of React Hooks, so I also went ahead and changed the class-based component to a function-based component with React Hooks, since it was easier to apply their example that way than to figure out how to thread both the React Router and Algolia state into the right places.

In terms of the internal structure of the `<ServiceDiscoveryResults>`, I decided to split it up into two separate components: the outer one (`<ServiceDiscoveryResults>`) is responsible for all the state, including syncing the react router state, the browser address, and the API requests to our backend service, and the inner one is a pure function component with no state.

The way that the flow works now is that the ServiceDiscoveryModal is still a modal on the home page, but instead of directly rendering the ServiceDiscoveryResults, it now serializes the important information from the form data into the URL query parameters and issues a react-router `<Redirect>` to the new `/service-discovery-results` route. The `<ServiceDiscoveryResults>` component is now responsible for deserializing that state out of the URL, as well as making some duplicated queries to the askdarcel-api service.

I had to make the redundant queries to the backend service, even though we just made them on the modal form pages, since it's on a brand new page. I don't think this is too bad, but it will now trigger a loading spinner when you first arrive at the search results page. In theory, we could probably cache the API results if we actually implemented a real "data service" that could cache results, but that's probably not an optimization that we need to make today.

### DIFF
--- a/app/components/ui/Modal/ServiceDiscoveryModal/ServiceDiscoveryModal.jsx
+++ b/app/components/ui/Modal/ServiceDiscoveryModal/ServiceDiscoveryModal.jsx
@@ -1,6 +1,7 @@
 import React, { Component } from 'react';
+import { Redirect } from 'react-router-dom';
+import qs from 'qs';
 import { STEPS } from 'components/ui/Modal/ServiceDiscoveryModal/constants';
-import ServiceDiscoveryResults from '../../../../pages/ServiceDiscoveryResults/ServiceDiscoveryResults';
 import { BaseModal } from '../Modal';
 import * as dataService from '../../../../utils/DataService';
 
@@ -67,7 +68,7 @@ class ServiceDiscoveryModal extends Component {
 
   render() {
     const {
-      closeModal, steps, categoryName, algoliaCategoryName,
+      closeModal, steps, categoryId,
     } = this.props;
     const {
       eligibilities, subcategories, selectedEligibilities, selectedSubcategories, currentStep,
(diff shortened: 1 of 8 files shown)


richardxia:
Screen Shot 2020-06-23 at 11 38 59 PM
Screen Shot 2020-06-23 at 11 39 05 PM

This should be good for a final review now. The changes I made since last time were:

- Update the URL of the results page to match the ones in the Clubhouse ticket
- Refactored the ServiceDiscoveryModal form such that it now has its own route rather than being nested within the GuideLists
- Refactored some of the other logic such that the main configuration of the pathways (name, ID, steps, etc.) is in a simple JavaScript array, which is referenced in multiple places rather than only being expressed in the GuideList React code

We should probably refactor the code such that the ServiceDiscoveryModal is no longer a modal and to reorganize the file directory structure to reflect the URL structure, so that they're easier to find, but I think these can be done later.

PR #963, opened 2020-06-29
Refactor ServiceDiscoveryForm to not be a modal

richardxia:
This PR does some refactoring of the ServiceDiscoveryForm to not make it a modal anymore, since it's a little bit odd to have a standalone page with nothing but a modal that cannot be closed.

I mostly just pulled together all of the components into the ServiceDiscoveryForm page, and I also refactored it to use React Hooks, which allowed me to reuse some of the hooks code I put in the ServiceDiscoveryResults page.

I also removed the dependency on react-modal, since it's now unused. Sorry about the confusion :/, we should have confirmed with the design team what they wanted.

I think doing this will make it a bit easier to iterate on some of the final changes that we're looking to make for this Service Discovery feature.


app/components/ui/Modal/Modal.jsx
```diff
@@ -1,60 +0,0 @@
-import React, { Fragment } from 'react';
-import PropTypes from 'prop-types';
-import { noop } from 'lodash';
-import Modal from 'react-modal';
-
-import styles from './Modal.scss';
-
-export const BaseModal = props => {
```
AnnaKalkanis, line 8: Any reason to keep this around for the modal work we were planning on doing on the edit pages?

AnnaKalkanis: Actually I take that back, I don't think this is worth hanging on to. It's quick enough to implement when we need it, and a lot of this was specifically geared towards this specific implementation that took over the full page. Would rather delete it all now so we don't end up with junk we don't need later.

richardxia: Yeah, and this will always be available in the Git history for us to take a look at in case if we want to reference it in the future.

richardxia:
Thanks!

PR #969, opened 2020-07-12
Update buttons and add progress bar.

richardxia:
This updates the footer of the service discovery form to match the mockups in https://www.figma.com/file/lhmZSzRw4mbDpil3CZ7tsC/SFSG-Topic-listings---For-production?node-id=0%3A1

I updated the buttons to match the guide, and I also added a progress bar. However, I hid the progress bar when there's only one page of questions (which I think is all of them), since it's kind of silly to see a full progress bar with 1/1 pages on every page. However, I can change this if @JacobDFrank wants the progress bar to always be there.

Here's what it looks like, including with some faked out data to make the progress bar show up:

## Chrome (with progress bar)
Screen Shot 2020-07-12 at 4 09 24 PM

## Firefox (with progress bar)
Screen Shot 2020-07-12 at 4 09 29 PM

## Chrome (without progress bar) (This is the only one that we'll actually see in prod, unless you want me to change anything)
Screen Shot 2020-07-12 at 4 17 36 PM

## Chrome (with progress bar, even with only 1 element)
Screen Shot 2020-07-12 at 4 23 58 PM

---

Also, this is the first time I ever learned about the HTML [`<progress>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/progress) element, which is kinda neat. I followed the [guide here](https://css-tricks.com/html5-progress-element/) on styling it, since you have to do different things in Chrome/Safari/webkit vs Firefox. I didn't bother with Internet Explorer because 1) it's harder for me to test that on a Mac and 2) the latest release of Edge now runs Chromium under the hood.

### DIFF
--- a/app/pages/ServiceDiscoveryForm/ServiceDiscoveryForm.jsx
+++ b/app/pages/ServiceDiscoveryForm/ServiceDiscoveryForm.jsx
@@ -117,7 +117,12 @@ const InnerServiceDiscoveryForm = ({
     <Fragment>
       <Header onGoBack={goBack} />
       <Content />
-      <Footer onGoBack={goBack} onNextStep={goToNextStep} />
+      <Footer
+        onGoBack={goBack}
+        onNextStep={goToNextStep}
+        currentStep={currentStep}
+        numSteps={steps.length}
+      />
     </Fragment>
   );
 };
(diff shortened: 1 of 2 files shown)


JacobDFrank:
I think @richardxia is right, there's no need for the progress bar when there is only one question.

PR #970, opened 2020-07-13
add icons to actions

AnnaKalkanis:
Just adding some icons to the actions you can take on a result.

Before:
Screen Shot 2020-07-12 at 6 14 13 PM

After:
Screen Shot 2020-07-12 at 6 18 52 PM

Note: I don't know how often we have an extension, but I went back and forth if it was better to do something like

```
    white-space: nowrap;
    overflow: scroll;
```
and have it truncated but scrollable, or to have it wrap to the next line like 
Screen Shot 2020-07-12 at 6 10 30 PM

I don't really like either, but for some reason my mind isn't coming up with anything better. I didn't want to use an ellipsis and cut anything off since the full number is useful. Let me know if you have an opinion/better idea.


richardxia:
I think the weird extension thing only exists in the dev environment, but in staging/prod, they're formatted a bit more nicely.

PR #971, opened 2020-07-13
Mobile display of search results page is unreadable

davidagustin:
Updated the css for the mobile view

Screen Shot 2020-07-12 at 8 48 13 PM


davidagustin:
Screen Shot 2020-07-12 at 8 53 36 PM
Also updated the view to "Clear All Resources" and took out the title "Filter Resources"

app/pages/ServiceDiscoveryResults/SearchResults/SearchResults.scss
```diff
@@ -26,22 +38,46 @@
...
   }
-  
+
   .address {
     font-size: 18px;
     margin-top: 12px;
+
+    @media screen and (max-width: 600px) {
+      width: 320px;
+      font-size: 14px;
```
richardxia, line 54: I think generally, the smallest font size that people recommend on mobile views is 16px. Any smaller and I think this will exclude older users or visually impaired users who may not have as keen eyesight.

app/pages/ServiceDiscoveryResults/SearchResults/SearchResults.scss
```diff
@@ -18,6 +18,18 @@
...
   border-bottom: 1px solid #F4F4F4;
+
+  @media screen and (max-width: 1120px) {
+    flex-direction: column;
+  }
+
+  @media screen and (max-width: 600px) {
+    padding-left: 21px;
+    padding-top: 18px;
+    padding-bottom: 18px;
```
richardxia, line 29: Is there a reason the padding is asymmetric between left and right? This will cause some things to be uncomfortably close to the right side of the screen.

davidagustin: Yeah, shouldn't be there. Was following the figma too close.

davidagustin:
Screen Shot 2020-07-12 at 9 15 51 PM

JacobDFrank:
I didn't anticipate what breakpoints should be used, use the one's you think would work best to best present the design.

However, it is not worth representing design perfectly below 320px and above 1280px.

app/pages/ServiceDiscoveryResults/SearchResults/SearchResults.scss
```diff
@@ -18,6 +18,19 @@
...
+    flex-direction: column;
+  }
+
+  @media screen and (max-width: 600px) {
+    padding: 18px 20px;
+    width: 375px;
+  }
+
+  @media screen and (max-width: 568px) {
+    width: 320px;
```
richardxia, line 32:
```suggestion
```

I think you can actually just get rid of these explicit widths, since you can just let the text flexibly expand to fit the size of the screen. This will allow the page to look nicer across a wider range of screen sizes.

app/pages/ServiceDiscoveryResults/SearchResults/SearchResults.scss
```diff
@@ -26,22 +39,44 @@
...
     font-size: 18px;
     margin: 32px 0;
+
+    @media screen and (max-width: 600px) {
+      font-size: 16px;
+      width: 335px;
+    }
+
+    @media screen and (max-width: 568px) {
+      width: 279px;
```
richardxia, line 67:
```suggestion
```

Similarly, you can remove the explicit widths here.

app/pages/ServiceDiscoveryResults/ServiceDiscoveryResults.scss
```diff
@@ -63,9 +68,13 @@
     margin-left: 0;
     cursor: pointer;
   }
+
+  @media screen and (max-width: 767px) {
```
richardxia, line 72:
```suggestion
  @media screen and (max-width: 600px) {
```

It's kind of strange to use this breakpoint. Why not use the same 600px breakpoint seen elsewhere on this page? Alternatively, you could make all of the 600px breakpoints be 767px, but it's just a bit confusing to have so many different breakpoints on the page at the same time.

app/pages/ServiceDiscoveryResults/SearchResults/SearchResults.scss
```diff
@@ -69,4 +104,15 @@
     white-space: nowrap;
     overflow: scroll;
   }
-}
\ No newline at end of file
+
+  @media screen and (max-width: 1120px) {
+    flex-direction: column;
+    margin-left: 0;
```
richardxia, line 110: For consistency, you should zero out both the left and the right margin. Otherwise, you end up with asymmetry like this:

Screen Shot 2020-07-14 at 12 06 04 AM

Even though that orange margin is invisible, it still affects the page layout, and it could have unexpected effects if we make other changes to the layout of the page.

app/pages/ServiceDiscoveryResults/SearchResults/SearchResults.scss
```diff
@@ -69,4 +104,15 @@
     white-space: nowrap;
     overflow: scroll;
   }
-}
\ No newline at end of file
+
+  @media screen and (max-width: 1120px) {
```
richardxia, line 108: Could you move this and the following media query higher up? Since they affect the `.sideLinks` selector, it's easier to read if you don't have the nested selectors in between, since these media queries affect the same element as the stuff at the top, not the nested selectors.

app/pages/ServiceDiscoveryResults/SearchResults/SearchResults.scss
```diff
@@ -69,4 +104,15 @@
...
+    flex-direction: column;
```
richardxia, line 109: You should probably have this selector reset the `width` back to `inherit`, since otherwise it'll continue to have the 250px width from the non-media-query selector. Without doing this, you end up with the phone numbers being truncated:

Screen Shot 2020-07-14 at 12 10 21 AM

PR #977, opened 2020-08-24
Homepage Redesign

AnnaKalkanis:
This PR looks bigger than it is at first glance, but the file count is mostly due to deleting old files/assets that are no longer in use. You can find the figma for the redesign [here](https://www.figma.com/file/sqT7cvXEPTNO4vqCLWCRdy/SFSG-Homepage-redesign---For-production?node-id=1%3A671)

**What I've done in this PR, by commit:**

1. Removed the featured categories (`CategoryList` component). These used to sit under the `Browse Directory`, but are removed in the redesign. (I confirmed the removal was intentional with Laura [here](https://sheltertech.slack.com/archives/C01A1JVUP4G/p1598221760007600))Screen Shot 2020-08-23 at 8 42 46 PM
2. Split the existing resource guides into 2 sections: `COVID-19 Resource Guides` and `Get step-by-step help`
3. Created new `ResourceList` and `ResourceCard` components to make this list view a little more re-usable. Replaced the existing `GuideList` with these.
4. Deleted the `GuideList` component & all the old image files.

Here's what it looks like at full-width:
Screen Shot 2020-08-23 at 8 48 03 PM

Responsive:
Screen Shot 2020-08-23 at 8 48 46 PM


JacobDFrank:
There's a drop shadow on the jobs icon, i think that was just an error in the design. If its an icon error you should be able to export the correct svg from figma or just remove the shadow from the css.

AnnaKalkanis:
nice catch! I've updated the jobs icon

PR #980, opened 2020-09-06
Add sitemap.xml

AnnaKalkanis:
This PR just adds a static sitemap.xml file, and a webpack plugin to copy it over to our build folder. Hopefully we can follow up by generating this sitemap file dynamically, but this should at least get us something.

Here it is in action!
Screen Shot 2020-09-06 at 3 02 59 PM


richardxia (approved):
LGTM! Thanks for doing this!

PR #994, opened 2020-11-21
Textellent

fayceltouili:


app/components/Texting/Texting.jsx
```diff
@@ -0,0 +1,223 @@
...
+  const errorView = () => (
+    <div className="responseIcons error">
+      <img src={images.icon('emergency')} alt="error" className="responseIcon" />
+      <h1 className="responseText error">Sorry, something went wrong, please try again later.</h1>
+    </div>
+  );
+
+  const view = () => {
+    if (error) {
+      return errorView();
```
AnnaKalkanis, line 193: I like how you've broken these out! I don't think you need to wrap them in functions, though. I think you should be able to get away with just assigning them directly to the variables like:

```
const loadingView = <Loader />;

...

const view = () => {
    ... 
    if(isLoading) {
        return loadingView;
    }
};
```

package.json
```diff
@@ -8,80 +8,80 @@
   },
   "browser": "app/init.jsx",
   "dependencies": {
-    "@babel/polyfill": "^7.4.4",
```
richardxia, line 11: @fayceltouili, you may want to discard all of your changes to both package.json and package-lock.json, since you've bumped the versions of many libraries, and that might be pretty dangerous, since there could be incompatibilities that we haven't tested for.

If you need to bump the versions of a few specific dependencies, that should be fine, but I wouldn't recommend bumping the versions of nearly every library here.

app/components/Texting/Texting.jsx
fayceltouili: Thank you @AnnaKalkanis 
I think we could just return Loader component like this:

 ```
 const view = () => {
  . . .
    if (isLoading) {
      return <Loader />;
    }
  . . .
  };
```

File: package.json (hunk 2 above)
Comment (fayceltouili) [reply package.json:11]: @richardxia, you've right I will add package.json and package-lock.json to .gitignore

File: app/components/Texting/Texting.jsx [hunk 3]
```diff
@@ -0,0 +1,223 @@
...
+    resourceId: id,
+    engagementType: 'Resource Info',
+    engagementInfo: {
+      Org_Name: resource.name,
+      Org_Address1: addresses.length ? addresses[0].address_1 : '',
+      Org_Address2: addresses.length ? addresses[0].address_2 : '',
+      City: addresses.length ? addresses[0].city : '',
+      State: addresses.length ? addresses[0].state_province : '',
+      Zip: addresses.length ? addresses[0].postal_code : '',
+      Org_Phone: phones ? phones[0].number : '',
```
richardxia, line 57: Could we use `snake_case` for all of these parameters? The rest of our APIs follow that naming convention, so I think it'd be better to stick with that.

I think you might be passing these directly through to the Textellent API, but if that's the case, you probably should apply the data transformation on our own backend API so that we decouple our frontend code from the Textellent API.

Also, for many of these fields, you can supply the data entirely from our backend API. You can even set the blank default values on some of these fields directly from our backend API so that the frontend code doesn't have to deal with it.

app/components/Texting/Texting.jsx
```diff
@@ -0,0 +1,223 @@
...
+  agreed: false,
+  isLoading: false,
+  isSent: false,
+  error: false,
+};
+
+// Text resource informations to the user phone
+
+const Texting = ({ toggle, resource }) => {
+  const [state, setState] = useState(initialState);
```
richardxia, line 22: It's generally not recommended to set multiple properties like this in a single `useState()` variable. See the [docs](https://reactjs.org/docs/hooks-faq.html#should-i-use-one-or-many-state-variables) for some more detailed information, but you will usually want a separate `useState()` variable for each state variable. In some cases it makes sense to combine values into the same state variable, but usually only if are related and often change together.

app/components/Texting/Texting.jsx
```diff
@@ -0,0 +1,223 @@
...
+        <div className="popup">
+          <div className="closeIconDiv" role="button" onClick={toggle} tabIndex={0}>
+            <img src={images.icon('close')} alt="close" className="closeIcon" />
+          </div>
+          { view() }
+        </div>
+      </div>, document.body,
+    )
+  );
+};
```
richardxia, line 216: This is quite a long function, and it might be a good idea to split it up into different pieces. One possible way of organizing this would be split the "logic" from the "view", where the state and the callbacks can be defined in one component and the actual view can be implemented as a pure function component. Another way of organizing the component would be to split up the views into separate smaller components representing the individual views.

app/components/Texting/Texting.jsx
```diff
@@ -0,0 +1,223 @@
+import React, { useState } from 'react';
+import ReactDOM from 'react-dom';
+import PropTypes from 'prop-types';
+import './Texting.scss';
```
richardxia, line 4: Do you think you could import this as a [CSS module](https://github.com/css-modules/css-modules) rather than a global CSs import? We use a library called CSS Modules, which allows us to write CSS files that are all locally scoped rather than globally scoped. This helps us avoid naming conflicts between different CSS files, since you can freely reuse names across CSS modules.

When imported into JavaScript, you can access each CSS class as a property on the imported object, and you are expected to pass them into the `className` property on each HTML element rather than just passing in a standalone string. See https://github.com/ShelterTechSF/askdarcel-web/blob/001e9a039787b9cdfa2d7142672ed0a925320d26/app/components/ui/Navigation/Navigation.jsx#L6 for an example of use using CSS modules. Also, you will need to update webpack.config.js to add this file to the list of files that should be treated as CSS modules: https://github.com/ShelterTechSF/askdarcel-web/blob/001e9a039787b9cdfa2d7142672ed0a925320d26/webpack.config.js#L134-L147

package.json
richardxia: Looks like these changes to package.json and package-lock.json are still present in this branch.

BTW, you shouldn't add either of these files to .gitignore; they are important to make sure we are all synchronized in terms of which versions of packages we are using.

You should be able to restore these back to what they were before with the following sequence of commands:

```sh
# Ensure latest commits from GitHub are fetched
git fetch

# Change package.json and package-lock.json back to what they were on the master branch
git restore --source=origin/master package.json package-lock.json

# Add and commit the changes
git add package.json package-lock.json
git commit -m "Restore original package.json and package-lock.json"
```

fayceltouili: Hi @richardxia I've fixed package.json and package-lock.json changes. Thank you!

app/components/Texting/Texting.scss
```diff
@@ -0,0 +1,178 @@
...
+  margin-bottom: 20px;
+  cursor: pointer;
+}
+
+.circle {
+    width: 150px;
+    height: 150px;
+    background: #00916A;
+    border-radius: 50%;
+    position: absolute;
```
richardxia, line 152: What is this `position: absolute` meant to do? It doesn't look like it has any effect on the positioning of this element, since when I disable it, it doesn't seem to change the page at all for me.

richardxia (commented):
Sorry for taking so long to review this! Thanks for working on this! The feature as a whole looks pretty good, and most of my comments are more about code organization and polish.

app/components/Texting/FormView/FormView.jsx
```diff
@@ -0,0 +1,55 @@
...
+
+  const buttonDisabled = () => !(agreed && /^\(?(\d{3})\)?[-. ]?(\d{3})[- . ]?(\d{4})$/.test(phone_number));
+
+  const onChange = name => event => {
+    const { type, value, checked } = event.target;
+    const newValue = type === 'checkbox' ? checked : value;
+    setState(prevState => ({ ...prevState, [name]: newValue }));
+  };
+
+  const config = useMemo(() => getConfig({ state, onChange }), [state]);
```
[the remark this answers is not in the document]
fayceltouili: I just wanted to generate form fields dynamically. Yes if you want I could remove `RenderInputs.jsx` and `config.js` and directly instantiate the form fields in this component

app/components/Texting/FormView/components/Privacy.jsx
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+import styles from '../../Texting.scss';
+
+const Privacy = () => (
+  <div className={styles.privacy}>
+    We will not share or sell your information with third party sites.
```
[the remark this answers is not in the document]
fayceltouili: I just copied the text exists in [Figma](https://www.figma.com/file/ChnhJ50Zq2IVZRSpIQ44xJ/Textellent?node-id=1%3A857).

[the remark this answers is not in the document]
richardxia: Thanks, I'll post my comment in Figma, then.

app/components/Texting/FormView/FormView.jsx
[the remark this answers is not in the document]
richardxia: Ah, is there a specific reason you were trying to generate the form fields dynamically? I think it’s tricky to get it right, not to mention it makes it difficult to read and difficult for static analysis tools, such as linters, to understand the code when you do this.

If you didn't have a specific reason for dynamically generating the form, then could we make the fields static? For this feature, it doesn't look like there's more than a single, fixed form, so it didn't seem to me that dynamic field generation was providing much utility here.

app/components/Texting/Texting.scss
```diff
@@ -0,0 +1,175 @@
...
+  border: 0;
+  cursor: pointer;
+  display: inline-block;
+  padding: 0;
+  position: absolute;
+  right: 20px;
+  top: 20px;
+}
+
+.title{
```
richardxia, line 45: Nit: Can you make sure there's a space before the `{`?

```suggestion
.title {
```

app/components/Texting/Texting.scss
```diff
@@ -0,0 +1,175 @@
...
+  margin-top: 1.5em;
+  font-style: normal;
+  font-weight: bold;
+  font-size: 1.8em;
+}
+
+.description {
+  font-size: 1em;
+  margin-top: 2em;
```
richardxia, line 54: This margin looks too large. In the actual Figma design, it appears to be 16px, which will be much smaller than 2em unless if your font size is 8px.

app/components/Texting/Texting.scss
richardxia, line 45: Separately, I think you need to set the font color, since the color that is showing up in the actual web app is a dark gray (`#30383A`), but in the Figma design, it should be an absolute black (`#000000`).

app/components/Texting/Texting.scss
```diff
@@ -0,0 +1,175 @@
...
+  position: absolute;
+  right: 20px;
+  top: 20px;
+}
+
+.title{
+  margin-top: 1.5em;
+  font-style: normal;
+  font-weight: bold;
+  font-size: 1.8em;
```
richardxia, line 49: In general, I don't think you want to set font sizes using `em` units, since it's a bit unintuitive how they work. It will set the font size based on the `em` size of the parent element's font size. This means that when you resize the parent element's font size, you may need to update the sizes here.

app/components/Texting/Texting.scss
```diff
@@ -0,0 +1,175 @@
...
+  }
+}
+
+.content {
+  position: relative;
+  width: 500px;
+  height: 660px;
+  background: #FFFFFF;
+  margin: 35px auto;
+  padding: 30px;
```
richardxia, line 20: This padding looks a little bit wrong. From the Figma design, it looks like there are 40px of padding on both the left and right sides and 24px of padding on the bottom.

app/components/Texting/Texting.scss
```diff
@@ -0,0 +1,175 @@
...
+
+.input {
+  background-color: $color-white;
+  height: 50px;
+  @media screen and (max-width: 900px ) {
+    height: 35px;
+  }
+}
+
+.label {
```
richardxia, line 65: In general, it looks like there is a little less spacing in your implementation than in the Figma design. Same goes for the dataRates submessage.

app/components/Texting/Texting.scss
```diff
@@ -0,0 +1,175 @@
...
+  top: 33%;
+  left: 35%;
+}
+
+.responseIcon {
+  width: 116.67px;
+  margin-top: 20%;
+}
+
+.successText {
```
richardxia, line 158: This font color also looks like it should be darker, an absolute black rather than a dark gray.

app/components/Texting/Texting.scss
```diff
@@ -0,0 +1,175 @@
...
+.dataRates {
+  font-size: 0.9em;
+  color: $color-grey5;
+}
+
+.checkBox {
+  margin-top: 1em;
+  font-size: 0.9em;
+  color: $color-grey5;
+  align-items: center;
```
richardxia, line 81: `align-items` will have no effect if this element is not a flexbox parent or a grid parent.

app/components/Texting/FormView/FormView.jsx
```diff
@@ -0,0 +1,96 @@
...
+    <div>
+      <Heading serviceName={serviceName} />
+      <label className={styles.label}>
+        First name (optional)
+        <input
+          type="text"
+          name="recipient_name"
+          className={styles.input}
+          value={recipient_name}
+          data-field="recipientName"
```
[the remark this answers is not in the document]
fayceltouili: `data-field` is a selector for Testcafe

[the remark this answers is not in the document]
richardxia: Ah, I see, thanks for pointing that out. However, wouldn't it be just as easy to select based on the `name` attribute, which wouldn't require defining a separate field just for testing purposes?

richardxia (approved):
Thanks for making the updates! Looks like you're failing one small lint check (line too long), but other than that, this looks pretty good to me. Congrats on shipping this!

PR #997, opened 2020-11-29
1722 - remove scroll to get rid of gray box in results

quanhuynh:
remove scroll in sideLinkText to get rid of the gray boxes
https://app.clubhouse.io/sheltertech/story/1722/on-some-environments-images-in-search-results-page-appear-with-gray-background

before:
Screen Shot 2020-11-29 at 3 31 30 PM

after:
Screen Shot 2020-11-29 at 3 31 44 PM

### DIFF
--- a/app/pages/ServiceDiscoveryResults/SearchResults/SearchResults.scss
+++ b/app/pages/ServiceDiscoveryResults/SearchResults/SearchResults.scss
@@ -105,5 +105,6 @@
   .sideLinkText {
     padding: 0 8px;
     white-space: nowrap;
+    overflow: auto;
   }
 }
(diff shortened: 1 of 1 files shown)


app/pages/ServiceDiscoveryResults/SearchResults/SearchResults.scss
```diff
@@ -105,6 +105,5 @@
   .sideLinkText {
     padding: 0 8px;
     white-space: nowrap;
-    overflow: scroll;
```
richardxia, line 108: You might want to set this to [`overflow: auto`](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow) instead. `overflow: auto` will add scrollbars only if the content is too wide to fit in the containing element. If someone does have a very long phone number, we probably want to allow people to scroll to see all of it, even if it doesn't look as great. Setting `overflow: auto` will at least make it look normal when the content does fit, though.

quanhuynh: got it. that makes sense

richardxia (approved):
LGTM!

PR #998, opened 2020-11-30
Rename CSS module files to *.module.scss

richardxia:
## Overview

This is kind of a large change, but hopefully it shouldn't affect any in-progress PRs besides @fayceltouili's https://github.com/ShelterTechSF/askdarcel-web/pull/994 and @Akhtam's https://github.com/ShelterTechSF/askdarcel-web/pull/989.

In this PR, I am renaming every .scss file that is a [CSS Module](https://github.com/css-modules/css-modules) to .module.scss. This allows us to eliminate the extra bookkeeping in webpack.config.js that lists exactly which .scss files are CSS modules and which ones are global stylesheets.

## Background

[CSS Modules](https://github.com/css-modules/css-modules) is a specification for treating a given CSS file as a modular file with locally-scoped CSS class names and how it interfaces with JavaScript. This is opposed to a normal (S)CSS file, where the class names are global, and where you must be cautious about not reusing a CSS class name between two different files, or else they could conflict and cause unintended effects. CSS Modules allows you to name your CSS classes without fear of name collisions, since the class names end up getting post-processed to include a hash in the name, allowing it to be unique to that file.

The very commonly used Webpack plugin [`css-loader`](https://github.com/webpack-contrib/css-loader) has CSS Modules built in to it, and it is automatically enabled for any file that ends in `.module.css` (or `.module.scss`), at least in newer versions of css-loader.

We have a mix of both (S)CSS Modules and global (S)CSS files, where we ideally want to migrate to mostly just using CSS Modules.

We've been using this convention in the rewrite of the https://github.com/ShelterTechSF/sheltertech.org codebase, and I feel like this has gone pretty well, so hopefully this isn't too difficult to adjust to.

## Developer-facing changes

From now on, I'd like to propose we use the name of a file to indicate whether it is a CSS module. All CSS module files should end in `.module.scss`. This makes it easier for us to opt-in to CSS modules one file at a time without having to explicitly list out each file in webpack.config.js. This also makes it easier for us to tell which files have been migrated to CSS Modules and which ones are still global.

@fayceltouili: In https://github.com/ShelterTechSF/askdarcel-web/pull/994, you will need to slightly update your code to match the new conventions being set in this PR. You will not need to touch webpack.config.js anymore, but in order to opt-in to CSS modules, you'll need to name your CSS file to match the `.module.scss` convention.
@Akhtam: In https://github.com/ShelterTechSF/askdarcel-web/pull/989, you will probably need to do something similar.

## Changes made in this PR

In addition to renaming all the existing CSS Module files and updating the Webpack configuration, I had to upgrade several of our libraries.

https://github.com/ShelterTechSF/askdarcel-web/commit/217b055107b663c842487a78738f8ef0e78204da upgrades Webpack to the latest version of the 4.x release, which was required in order to bump some of the other libraries. I held off from upgrading to Webpack 5, since that seemed to contain a substantial number of backwards-incompatible changes, and I didn't want to deal with making that work.

https://github.com/ShelterTechSF/askdarcel-web/commit/00cb24fa9c03baa658a823bbfc1ea89fbfb7de35 upgrades css-loader from 0.23.1 to 5.0.1, which is the latest release at the time of this writing. This is a lot of major version changes, and it includes jumping past the 1.0.0 stable release. Fortunately, there weren't too many changes to the configuration necessary, and the only other hiccup I ran into was solved by also upgrading style-loader.

https://github.com/ShelterTechSF/askdarcel-web/commit/e3a57ca0187c378c8989594aa4ff45e484394315 upgrades style-loader from 0.13.1 to 2.0.1, which is the latest release at the time of this writing. Since we only use the default configuration of style-loader, I didn't really have to change anything here.

https://github.com/ShelterTechSF/askdarcel-web/commit/b560d18237d169b2e50347a30c981c9f68160ba5 is the large change, where I actually renamed every CSS module file from `Foo.scss` to `Foo.module.scss`. This also contains the large change to webpack.config.js to collapse the three separate CSS-related loader rules into a single one, since css-loader natively supports the `.module.scss` naming convention for determining whether a file is a CSS module.

## How this was tested

I clicked through several pages in the local dev server while simultaneously navigating similar pages on the staging site. I manually inspected the pages to make sure that they looked the same on both my local dev server and staging. I'm fairly confident this is correct, since as I was working on this, I was dealing with either all CSS modules not working or all global CSS files not working, and it's very apparent if a large chunk of the CSS isn't working at all.


davidagustin (approved):

richardxia:
Thanks for taking a look!

PR #1001, opened 2020-12-13
1489: Add SF Families whitelabeling logic for logo and title

quanhuynh:
Adding logic to show SF Families logo and title when the domain is an SF Families domain.


drcaramelsyrup (approved):

app/components/App.jsx
```diff
@@ -134,6 +134,34 @@ class App extends Component {
...
+      );
+    } else {
+      return (
+        <>
+          <title>AskDarcel</title>
+          <meta property="og:url" content="https://askdarcel.org" />
+          <meta property="og:title" content="AskDarcel" />
+        </>
+      );
+    }
```
richardxia, line 162: I wonder if we could fold this all into a single, unconditional function that calls other functions to get the correct value for the properties. Since we already have a `getSiteTitle()` function, that could be used for both the `<title>` tag and the `<meta property="og:title">` tag. It would hopefully be easy to create a similar function for the site URL.

richardxia (approved):
LGTM, aside from one additional suggestion for DRYing up the code!

app/components/App.jsx
quanhuynh: that's a good idea!

PR #1002, opened 2020-12-13
Service Listing page not updating service website

iopkelvin:
The problem here is that: The service listing page did not update the service website when the service had a website inputted. It always kept the resource's website.
The Service page (within resource listing), however, did update the service website.

The error was found in service listing. Within ServiceListingPage.jsx, we have the "Table of Contacts info" which contains the service website. TableofContactInfo.jsx is a component, and there error is that:

The class ContactInfoTable has the constant website. This variable has the logic "service.website OR resource.website". I switched item.website to item.url, because service website is called url insteal of website in the API.

### DIFF
--- a/app/components/listing/TableOfContactInfo.jsx
+++ b/app/components/listing/TableOfContactInfo.jsx
@@ -5,7 +5,7 @@ class ContactInfoTable extends React.Component {
   render() {
     const { item } = this.props;
     // TODO May break for non services, need a better check for inheritance
-    const website = item.website || item.resource.website;
+    const website = item.url || item.resource.website;
     const email = item.email || item.resource.email;
     const phones = item.phones || item.resource.phones;
 


richardxia (approved):
Ah, good catch. Thanks for debugging and fixing this! It looks like the CI build flaked out, so I've restarted it, but if it passes, I can merge this in.

iopkelvin:
> Ah, good catch. Thanks for debugging and fixing this! It looks like the CI build flaked out, so I've restarted it, but if it passes, I can merge this in.

I don't understand why travis-ci keeps failing

richardxia:
It seems to be some intermittent issue that's failing more than 50% of the time. I'm not fully understanding it either, which is why I tried https://github.com/ShelterTechSF/askdarcel-web/pull/1003, but even when I rebased this branch on top of the changes I made to the master branch, it still seems to be pretty flaky. I can keep hitting replay until it succeeds.

richardxia:
Oh wait, actually, a couple things just clicked in my head just now! I don't think you can submit a PR from a fork of the repository, since for security reasons, Travis CI won't allow you to use the secret API keys that we have set up in Travis CI. We'll have to resubmit this PR as a branch from this actual repository, not from a fork. Do you think you can recreate this PR after pushing a branch directly to the ShelterTechSF copy of the repository?

iopkelvin:
> Oh wait, actually, a couple things just clicked in my head just now! I don't think you can submit a PR from a fork of the repository, since for security reasons, Travis CI won't allow you to use the secret API keys that we have set up in Travis CI. We'll have to resubmit this PR as a branch from this actual repository, not from a fork. Do you think you can recreate this PR after pushing a branch directly to the ShelterTechSF copy of the repository?

ok, that makes sense. Yes, I will re-do the setup. Should I not work with forks then?

richardxia:
> ok, that makes sense. Yes, I will re-do the setup. Should I not work with forks then?

Yeah, unfortunately, forks won't be able to access the necessary secrets for running the browser tests on Sauce Labs.

richardxia:
Replaced by https://github.com/ShelterTechSF/askdarcel-web/pull/1004

PR #1003, opened 2020-12-14
Upgrade testcafe

richardxia:
This is an attempt to try to make TestCafe and SauceLabs less flaky by upgrading some of the libraries. I also bumped up the version of the Chrome browser we are testing to the latest release (87).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -49,22 +49,19 @@
           "version": "4.0.0",
           "resolved": "https://registry.npmjs.org/arr-diff/-/arr-diff-4.0.0.tgz",
           "integrity": "sha1-1kYQdP6/7HHn4VI1dhoyml3HxSA=",
-          "dev": true,
-          "optional": true
+          "dev": true
         },
         "array-unique": {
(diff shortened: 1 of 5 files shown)


richardxia:
Looks like it failed because it didn't like the new configuration string I passed in to select the latest version of Chrome. Let me go back one version to see if that helps.

richardxia:
Picking an older browser version appeared to work, however, we still have a failure.

It turns out one of our tests was always failing, but because we never actually put an assertion on it, it would never actually fail. The latest version of TestCafe prints a warning if you have an `.expect()` statement without an assertion attached to it, like `.ok()` or `.eql()`, so when I "fixed" that, it actually started failing. Let me see if I can just quickly fix the test, since I think one of the selectors just changed.

richardxia:
Going to merge this in without review because it seems to be running Sauce Labs fairly reliably. I am not sure if this is because there was some temporary issue on either Sauce Labs' or Travis CI's ends earlier today or if my changes actually made things better, but I don't think it's harmful to merge.

PR #1004, opened 2020-12-14
Re-do: Service Listing page not updating service website

iopkelvin:
The problem here is that: The service listing page did not update the service website when the service had a website inputted. It always kept the resource's website.
The Service page (within resource listing), however, did update the service website.

The error was found in service listing. Within ServiceListingPage.jsx, we have the "Table of Contacts info" which contains the service website. TableofContactInfo.jsx is a component, and there error is that:

The class ContactInfoTable has the constant website. This variable has the logic "service.website OR resource.website". I switched item.website to item.url, because service website is called url insteal of website in the API.


richardxia (approved):
LGTM! Looks like this time, the CI passed!

PR #1010, opened 2021-01-18
Karla 372 login button

Raawr:


app/components/ui/Navigation/Navigation.jsx
```diff
@@ -39,6 +39,18 @@ class Navigation extends React.Component {
   render() {
     const { showSearch, toggleHamburgerMenu } = this.props;
     const { showSecondarySearch, query } = this.state;
+    /* const cookieVal = document.cookie = "username=rar"; */
+    const renderAuthButton = ()=>{
+      /* get cookie value by key 'username', should this be inside renderAuthButton? */
+      let cookieValue = document.cookie.replace(/(?:(?:^|.;\s)username\s*\=\s*([^;]*).*$)|^.*$/, "$1");
+      if(!cookieValue){
+      return <li>
```
richardxia, line 47: Nit: your indentation here doesn't match normal indentation standards, where you'd indent the body of `if` statements and whatnot. Our linter should be able to automatically fix this and many other kinds of code formatting issues if you run `npm run lint:fix`.

richardxia (commented):
Thanks for submitting this! I gave you a few pointers on a few things to look at. I agree with your code comment about moving the logic for checking if someone has been auth'd already somewhere else, and I think putting it in App.jsx would make the most sense.

PR #1018, opened 2021-02-13
Create top level header component

arashcodes:
Ticket 1846 - Creating the top-level header component for the admin interface


Ammar-alnatsheh:
I would suggest saving the style in the 'Admin.module.scss' and import it into your component since this method been used on the other pages I think it would be nice to keep the same style work. thx

arashcodes:
Thank you, @Ammar-alnatsheh. I moved the style to Admin.module.scss' per your comment.

PR #1020, opened 2021-02-21
Adding Google Tag Manager

ruochenhuang:
@richardxia Google Tag Manager script

### DIFF
--- a/app/index.html
+++ b/app/index.html
@@ -1,12 +1,23 @@
 <!DOCTYPE html>
 <html lang="en">
 <head>
+  <!-- Google Tag Manager -->
+<script>(function(w,d,s,l,i){w[l]=w[l]||[];w[l].push({'gtm.start':
+new Date().getTime(),event:'gtm.js'});var f=d.getElementsByTagName(s)[0],
+j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';j.async=true;j.src=
+'https://www.googletagmanager.com/gtm.js?id='+i+dl;f.parentNode.insertBefore(j,f);
+})(window,document,'script','dataLayer','GTM-5S2P6SD');</script>
+<!-- End Google Tag Manager -->
   <meta charset="utf-8">
   <meta name="viewport" content="width=device-width, initial-scale=1">
   <title><%= htmlWebpackPlugin.options.title %></title>
   <meta property="og:image" content="https://sfserviceguide.org<%= require('assets/img/sfsg-preview.png') %>" />
 </head>
 <body>
+  <!-- Google Tag Manager (noscript) -->
+<noscript><iframe src="https://www.googletagmanager.com/ns.html?id=GTM-5S2P6SD"
+height="0" width="0" style="display:none;visibility:hidden"></iframe></noscript>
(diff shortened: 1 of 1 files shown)


app/index.html
```diff
@@ -1,12 +1,23 @@
 <!DOCTYPE html>
 <html lang="en">
 <head>
+  <!-- Google Tag Manager -->
+<script>(function(w,d,s,l,i){w[l]=w[l]||[];w[l].push({'gtm.start':
+new Date().getTime(),event:'gtm.js'});var f=d.getElementsByTagName(s)[0],
+j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';j.async=true;j.src=
+'https://www.googletagmanager.com/gtm.js?id='+i+dl;f.parentNode.insertBefore(j,f);
+})(window,document,'script','dataLayer','GTM-5S2P6SD');</script>
+<!-- End Google Tag Manager -->
```
richardxia, line 10: Even though this is just a static file that's separate from the rest of the app, can we indent this to match the indentation of the rest of the file? Children of an element should be indented one level (two spaces) inward, and all sibling elements should be at the same indentation level.

Also, since this isn't a human readable snippet anyway, can we just put this entire `<script>` tag on one line? Putting line breaks arbitrarily like this makes it slightly more difficult to scan the file, since it's hard to see when the snippet starts and when the snippet ends.

app/index.html
```diff
@@ -1,12 +1,23 @@
...
   <meta charset="utf-8">
   <meta name="viewport" content="width=device-width, initial-scale=1">
   <title><%= htmlWebpackPlugin.options.title %></title>
   <meta property="og:image" content="https://sfserviceguide.org<%= require('assets/img/sfsg-preview.png') %>" />
 </head>
 <body>
+  <!-- Google Tag Manager (noscript) -->
+<noscript><iframe src="https://www.googletagmanager.com/ns.html?id=GTM-5S2P6SD"
+height="0" width="0" style="display:none;visibility:hidden"></iframe></noscript>
+<!-- End Google Tag Manager (noscript) -->
```
richardxia, line 20: Same comment about indentation and not inserting line breaks in odd places like this.

richardxia (changes requested):
Sorry for taking a while to look at this. And sorry to burden you with some style requests, but could you make the changes I requested?

Although we don't adhere to any specific style guide for HTML (due to most of our app being written in React/JSX, not plain HTML), I think this style guide by Google generally describes a good, readable style: https://google.github.io/styleguide/htmlcssguide.html

In particular, please see https://google.github.io/styleguide/htmlcssguide.html#General_Formatting (but feel free to ignore the kind of weirdo HTML5 "you don't actually need to close certain tags" thing.)

app/index.html
richardxia: Also, just to be absolutely clear, even though HTML comments (`<!-- this stuff -->`) aren't real elements, I would appreciate if they also follow the same indentation rules.

ruochenhuang:
Duplicate -- https://github.com/ShelterTechSF/askdarcel-web/pull/1032

PR #1024, opened 2021-03-15
Remove mobile navigation buttons from SFFamilies whitelabel

drcaramelsyrup:
This removes the top level navigation buttons from SFFamilies. Free text is not yet supported, and all the hamburger menu items are SFSG related.

image

Note that the original SFSG / askdarcel sites retain their elements:
image

### DIFF
--- a/app/components/ui/Navigation/Navigation.jsx
+++ b/app/components/ui/Navigation/Navigation.jsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { Fragment } from 'react';
 import PropTypes from 'prop-types';
 import { Link, withRouter } from 'react-router-dom';
 import qs from 'qs';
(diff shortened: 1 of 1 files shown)


app/components/ui/Navigation/Navigation.jsx
```diff
@@ -71,8 +71,10 @@ class Navigation extends React.Component {
             }
           </div>
           <div className={styles.mobileNavigation}>
-            <button type="button" className={styles.searchButton} onClick={this.toggleSecondarySearch} />
-            <button type="button" className={styles.hamburgerButton} onClick={toggleHamburgerMenu} />
+            {!isSFFamiliesSite()
+              && (<button type="button" className={styles.searchButton} onClick={this.toggleSecondarySearch} />)}
+            {!isSFFamiliesSite()
+              && (<button type="button" className={styles.hamburgerButton} onClick={toggleHamburgerMenu} />)}
```
richardxia, line 77: I think this would be clearer as a single `if` expression, rather than separate ones for each button, because otherwise it's not obvious upon first glance that the conditions are the same.

```suggestion
            {!isSFFamiliesSite() && (
              <>
                <button type="button" className={styles.searchButton} onClick={this.toggleSecondarySearch} />
                <button type="button" className={styles.hamburgerButton} onClick={toggleHamburgerMenu} />
              </>
            )}
```

Also, I noticed there was an existing `{!isSFFamiliesSite() && ` conditional expression just below these two changes, so could you combine them all into a single one? I think this just means putting the entire `styles.mobileNavigation` div element inside of the conditonal.

drcaramelsyrup: Thanks! I was mostly being super lazy and didn't want to remember what react JSX errors meant. Added a Fragment to wrap the conditional.

richardxia (approved):
Awesome, thanks!

PR #1026, opened 2021-03-16
Remove unused SCSS code.

richardxia:
I took a pass through the SCSS code and deleted anything that didn't look like it was being used. My primary methodology was "grep for the CSS selector name and see if there are any results". Aside from code hygiene, I did this because I was trying to debug some z-index issues on the Edit page, and I kept on getting search results for z-index properties on CSS classes that are never used.

I tested this by trying to visit every distinct page on the site locally and making sure that nothing looks obviously wrong. In addition, anything that was a Sass mixin that was actually being used by something would have automatically been caught by the Sass preprocessor.

There's a bunch of more stuff that I'd like to get rid of but can't, since they're actually still being used. This includes stuff like an entirely Sass mixin system for safely using flexbox across a bunch of old browsers from before 2010, which 1) is probably not necessary given that all major browsers, including Internet Explorer 9, have supported flexbox and 2) the commonly accepted approach to browser prefixing these days is to use a CSS postprocessor that operates on modern CSS and transpiles to older CSS.


drcaramelsyrup (approved):
Checked out SFSG and SFFamilies sites briefly, lgtm.

PR #1027, opened 2021-03-17
UserWay ADA compliance for SFFamilies domains

drcaramelsyrup:
One of the go-live requirements for SFFamilies is ADA compliance, which SFFamilies does via [UserWay](https://userway.org/) (a commercial solution for easy ADA compliance). They've shared an account ID with us, with the instructions to embed this javascript on our site:
```
<script>(function(d){var s = d.createElement("script");s.setAttribute("data-account", "4EhNmEIwPz");s.setAttribute("src", "https://cdn.userway.org/widget.js");(d.body || d.head).appendChild(s);})(document)</script><noscript>Please ensure Javascript is enabled for purposes of <a href="https://userway.org">website accessibility</a></noscript>
```

Fortunately, we can have a lot of control over the DOM and take an alternate approach. I took inspiration from our react-intercom widget and wrapped up the ugliness in a component that we can toggle on and off along with the rest of the label. The `componentDidMount` logic should add the appropriate script element once.

Like Intercom, the script adds a little widget / iframe on which it can render a lot of other things, but it also does some _very_ intense DOM and element class manipulation that allows it to change the look of the entire site.

Could be interesting for SFSG proper but this isn't something we'd want immediately.


app/components/ui/UserWay.jsx
```diff
@@ -0,0 +1,30 @@
...
+    userWayScript.setAttribute("data-account", appID);
+    document.body.appendChild(userWayScript);
+  }
+
+  render() {
+    return (
+      <noscript>Please ensure Javascript is enabled for purposes of <a href="https://userway.org">website accessibility</a></noscript>
+    );
+  }
+};
```
richardxia, line 24: Could I suggest that we rewrite this as a function-based component with a hook? Since this mostly just the installation of the script, this can basically be a component that consists of a single `useEffect()` hook along with the return of the `<noscript>` element. I think roughly, this component would look like this with hooks:

```jsx
import React, { useEffect } from 'react';

const UserWay = ({ appID }) => {
  useEffect(() => {
    if (logicCheckingIfAlreadyInstalled) return;

    // The body of your componentDidMount goes here
  });

  return (
      <noscript>Please ensure Javascript is enabled for purposes of <a href="https://userway.org">website accessibility</a></noscript>
  );
}
```

richardxia (commented):
I am definitely drinking the React Hooks Kool-Aid, but I think it results in less boilerplate in a lot of cases, and it makes the stateful effects easier to reason about, since `useEffect` unifies several of the componentDid/Will/Should* functions into a single concept.

This is a pretty interesting widget. We probably need to write more accessible code ourselves, since a lot of how we write our CSS and even our HTML may make things less accessible.

app/components/ui/UserWay.jsx
drcaramelsyrup: I haven't been in React land in a long time, but hooks do look and seem like what the cool kids do. Doing it!

app/components/ui/UserWay.jsx
```diff
@@ -0,0 +1,32 @@
+import React, { useEffect, useState } from 'react';
+import PropTypes from 'prop-types';
+
+const UserWay = ({ appID }) => {
+  const [installed, setInstalled] = useState(false);
+
+  useEffect(() => {
+    if (installed) return;
```
richardxia, line 8: This isn't the safest way of checking whether the script has been installed, since the state of the component gets blown away if the component is ever unmounted and then remounted. Since you're modifying the actual DOM, you should let that be the source of truth, so how about you set an `id` on the `<script>` element you're creating and inserting into the DOM and the check whether an element of that ID exists here?

app/components/ui/UserWay.jsx
richardxia: Thanks for switching this over to Hooks!

app/components/ui/UserWay.jsx
drcaramelsyrup: Cool, that works, updated!

richardxia (approved):
Cool, LGTM!

PR #1030, opened 2021-03-22
Log browser console errors experienced when running TestCafe tests.

richardxia:
I'm having a terrible time debugging or reproducing the [TestCafe errors](https://travis-ci.org/github/ShelterTechSF/askdarcel-web/builds/763814860) in `master`.

No matter what I do, I can't reproduce the errors locally, even when I check out the `master` branches of both repos and I reset my database. When I look at the video recording of the browser test failures in Sauce Labs, I can see that in the "Add resource phone number" always triggers a "Sorry! An error occurred." message when the test tries to hit the Save button, but I have no idea what's causing it. According to [the code](https://github.com/ShelterTechSF/askdarcel-web/blob/efc32e14f17db6b4b9e39bbe5c93684f530677e2/app/pages/OrganizationEditPage.jsx#L691), that error message appears if any of the multiple HTTP request promises fails. It looks like we do [log the promise error to the console](https://github.com/ShelterTechSF/askdarcel-web/blob/efc32e14f17db6b4b9e39bbe5c93684f530677e2/app/pages/OrganizationEditPage.jsx#L688), but that's not possible for us to view from the current Travis CI output or the TestCafe output.

This PR is an attempt at getting TestCafe to print out any console errors that it sees in the browser to the command line console (i.e. Travis CI) so that we have more useful debug information. This isn't the cleanest way of doing things, since it pollutes the console output for even tests that succeed with stuff that looks like this:

```
 Edit Resource
Captured console.log() messages:
[]
Captured console.warn() messages:
[]
Captured console.error() messages:
["React-Hot-Loader: misconfiguration detected, using production version in non-production environment.","React-Hot-Loader: Hot Module Replacement is not enabled.","InvalidValueError: setIcon: not a string; and not an instance of PinView; and no url property; and no path property","React-Hot-Loader: misconfiguration detected, using production version in non-production environment.","React-Hot-Loader: Hot Module Replacement is not enabled.","React-Hot-Loader: misconfiguration detected, using production version in non-production environment.","React-Hot-Loader: Hot Module Replacement is not enabled.","InvalidValueError: setIcon: not a string; and not an instance of PinView; and no url property; and no path property"]
 ✓ Edit resource name
```

However, I want to submit this as a PR so that I can at least see why the current "Add resource phone number" test fails and whether this gives us more useful information before I clean it up a bit more.

### DIFF
--- a/testcafe/editResource.js
+++ b/testcafe/editResource.js
@@ -7,7 +7,19 @@ const editResourcePage = new EditResourcePage();
 const servicePage = new ServicePage();
 
 fixture`Edit Resource`
-  .page(ResourcePage.url(1));
+  .page(ResourcePage.url(1))
+  .afterEach(async t => {
+    const { log, warn, error } = await t.getBrowserConsoleMessages();
+
+    console.log('Captured console.log() messages:');
+    console.log(JSON.stringify(log));
+
+    console.log('Captured console.warn() messages:');
+    console.log(JSON.stringify(warn));
+
+    console.log('Captured console.error() messages:');
+    console.log(JSON.stringify(error));
+  });
 
 
 async function testEditTextProperty(t, showPageSelector, editPageSelector, newValue) {


testcafe/editResource.js
```diff
@@ -7,7 +7,19 @@ const editResourcePage = new EditResourcePage();
 const servicePage = new ServicePage();
 
 fixture`Edit Resource`
-  .page(ResourcePage.url(1));
+  .page(ResourcePage.url(1))
+  .afterEach(async t => {
+    const { log, warn, error } = await t.getBrowserConsoleMessages();
+
+    console.log('Captured console.log() messages:');
```
drcaramelsyrup, line 14: Seems like (later) we could put these behind a conditional if `log`, `warn`, etc. is empty or null (since these appear to be arrays).

drcaramelsyrup (approved):

testcafe/editResource.js
richardxia: Yeah, although unfortunately, our app constantly prints stuff to `console.log()` and `console.error()` even during normal behavior. For instance, every page load on a development server causes the "React-Hot-Loader: misconfiguration detected, using production version in non-production environment." error message. It looks like the right way to do this kind of thing in TestCafe is with a custom [Reporter](https://devexpress.github.io/testcafe/documentation/guides/extend-testcafe/reporter-plugin.html), so I might port this code into a reporter once I actually get it working.

I'm still trying to debug the issue, since I still apparently have not logged enough information to get the real answer.

richardxia:
Sigh. Apparently logging the response body doesn't help at all because we don't actually return any information in the response body when an API call fails.

Trying two more things:

1. Printing the API URL endpoint as part of the rejected Promise message, so that even if the body is empty, we know which request failed
2. Printing the logs from the askdarcel-api Docker container if the TestCafe tests fail

richardxia:
I've stopped being able to reproduce the error, which is annoying for this PR, but it does seem like restarting the other failing PRs is getting them to pass now, so I'm at least happy with that.

PR #1031, opened 2021-03-23
Temporarily disabling textellent integration

jjfreund:


richardxia (approved):
LGTM, although longer term, the local variable should probably be renamed to something more like `shouldShowTextingAction` and the logic flipped around, since it's kind of hard to reason about `isProduction = true` when that's literally not describing reality.

Something like `const shouldShowTextingAction = !process.env.IS_PRODUCTION` would be easier to read and also easy to temporarily hardcode to a fixed value.

lgarofalo (approved):

richardxia:
Well, at least we now know that the tests correctly check that this functionality works.

PR #1033, opened 2021-03-28
Remove axios and CLI tool for approving change requests

richardxia:
I was about to upgrade [axios](https://github.com/axios/axios), since we've had a security alert for it for a few months, when I realized that we're only really using it in ~one~ two places, so instead I'm just getting rid of it.

One of the places is a command line script we wrote a long time ago for approving all of the open change requests, which is no longer useful now that we auto-approve all change requests upon creation. This is not used at all by the main app, and was only ever used by hand, so I just deleted it to remove a thing that depended on axios.

The other place we use it is a trivial query on the Home page to get the total number of resources, which is used in the placeholder text of the search bar to say "Search N resources in San Francisco". This was trivial to rewrite in terms of [fetch](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API), which has been supported in all browsers since 2017 (with Safari being the last holdout). Also, we're installing a polyfill for fetch for older browsers anyway.

Since we're already mostly using fetch instead of axios on the site, and because fetch is built in to all modern browsers, I think it's better to use fetch here, since it allows us to shed a dependency and not have to deal with updating it in the future.

I tested this locally by making sure the search bar placeholder text still displays properly after this change.


lexholden (approved):
Awesome, LGTM.

Don't even remember why I added some of this crap, what a weird way to automate our API interactions :blush:

richardxia:
> Awesome, LGTM.
> 
> Don't even remember why I added some of this crap, what a weird way to automate our API interactions 😊

Haha, I think it was a progression from 1) we should have an approval flow for any edits to resources, to 2) there's too many changes to be approved, let's write a script to bulk approve all open ones, to 3) let's just immediately change resources on edit. We still have the ChangeRequest model as a vestigial piece of technical debt, though...

PR #1035, opened 2021-04-01
Validate phone number textellent

Akhtam:
Added Normalize phone number for input;
when user types phone number this module will format phone number: (111) 111-1111

Added Validation for US only phone numbers, keep Submit button disabled until phone number is valid.

### DIFF
--- a/app/components/Texting/components/FormView/Form.module.scss
+++ b/app/components/Texting/components/FormView/Form.module.scss
@@ -33,6 +33,7 @@
   background-color: $color-white;
   border: 1px solid #DDDDDD;
   border-radius: 2px;
+  letter-spacing: 1px;
   &:active, &:hover, &:focus {
     box-shadow: 0px 2px 4px 0px rgba(0,0,0,0.1);
   }
--- a/app/components/Texting/components/FormView/Form.module.scss
+++ b/app/components/Texting/components/FormView/Form.module.scss
@@ -48,7 +48,12 @@
   border: none;
   background: $color-white;
   padding-left: 5px;
+  margin: 0; /* to avoid Safari for adding extra margin */
   box-shadow: none !important;
+  /* Remove Chrome's autocomplete background color for input */
+  &:-webkit-autofill {
+    -webkit-box-shadow: 0 0 0 100px white inset !important;
+  }
 }
 
 .input {
--- a/app/components/Texting/components/FormView/Form.module.scss
+++ b/app/components/Texting/components/FormView/Form.module.scss
@@ -27,10 +27,32 @@
   }
 }
 
+.phoneInputBox {
+  display: flex;
+  margin-top: 10px;
+  background-color: $color-white;
+  border: 1px solid #DDDDDD;
+  border-radius: 2px;
+  &:active, &:hover, &:focus {
+    box-shadow: 0px 2px 4px 0px rgba(0,0,0,0.1);
+  }
+}
+
+.phoneInputPrefix {
+  padding: 0.9375rem 0 0.9375rem 0.9375rem;
+  color: $color-grey6;
+  font-size: 1.0625rem;
+}
+
+.phoneInput {
+  border: none;
+  background: $color-white;
+  padding-left: 5px;
+  box-shadow: none !important;
+}
+
 .input {
   background-color: $color-white;
-  height: 50px;
-  padding: 15px 20px;
   margin-top: 10px;
   @media screen and (max-width: 900px ) {
     height: 35px;
--- a/app/components/Texting/components/FormView/FormView.jsx
+++ b/app/components/Texting/components/FormView/FormView.jsx
@@ -1,5 +1,9 @@
 import React, { useState } from 'react';
 import PropTypes from 'prop-types';
+import {
+  normalize,
+  validatePhoneNumber,
+} from '../../../../utils/validatePhoneNumber';
 import Heading from './Heading';
 import Privacy from './Privacy';
 import Buttons from './Buttons';
@@ -7,14 +11,16 @@ import styles from './Form.module.scss';
 
 const initialState = {
   recipientName: '',
-  phoneNumber: '',
   agreed: false,
 };
 
 const FormView = ({ service, handleSubmit, closeModal }) => {
   const [state, setState] = useState(initialState);
-  const { recipientName, phoneNumber, agreed } = state;
+  const { recipientName, agreed } = state;
+  const [phoneNumber, setPhoneNumber] = useState('');
+  const [isValidNumber, setIsValidNumber] = useState(false);
   const { serviceName, serviceId } = service;
+
   const onChange = evt => {
     const {
       type,
@@ -24,6 +30,13 @@ const FormView = ({ service, handleSubmit, closeModal }) => {
     } = evt.target;
     const newValue = type === 'checkbox' ? checked : value;
     setState(prevState => ({ ...prevState, [name]: newValue }));
+  };
+
+  const onPhoneNumberChange = evt => {
+    const normalized = normalize(evt.target.value);
+    const isValid = validatePhoneNumber(normalized);
+    setIsValidNumber(isValid);
+    setPhoneNumber(normalized);
   };
 
   const onSubmit = event => {
@@ -62,7 +75,8 @@ const FormView = ({ service, handleSubmit, closeModal }) => {
               name="phoneNumber"
               className={styles.phoneInput}
               value={phoneNumber}
-              onChange={onChange}
+              maxLength="14"
+              onChange={onPhoneNumberChange}
             />
           </div>
         </label>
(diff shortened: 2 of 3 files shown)


app/utils/validatePhoneNumber.js
```diff
@@ -0,0 +1,26 @@
+
+/**
+ *  @module normalize/validatePhoneNumber
```
richardxia, line 3: Forgive me if I'm wrong, but did you copy this file from somewhere else? This doc comment `@module` declaration doesn't match the actual file path, and we don't even use `@module` declarations elsewhere in the codebase. In addition, the style of this file is completely unlike the style of any of the other code in this codebase, which is why I wanted to make sure I ask.

If you did copy this from elsewhere, then 1) we really need to provide attribution for it, both for legal reasons and just for ethics, and 2) we should really check the license of the code to make sure we are compliant (most licenses at least ask you to provide attribution, but they may have stronger requirements). Also, although it's less of an issue in our codebase, since we decided to give AskDarcel one of the stricter copyleft licenses, GPLv3, but in general, you should be very careful about code that you copy from elsewhere, since the license of that code may [force you to relicense your own code that you're integrating it with](https://en.wikipedia.org/wiki/Viral_license).

If you didn't copy this, then disregard everything that I wrote above, and I apologize for falsely suspecting that this code may have been copied.

app/components/Texting/components/FormView/Form.module.scss
```diff
@@ -27,10 +27,38 @@
...
+  padding: 0.9375rem 0 0.9375rem 0.9375rem;
+  color: $color-grey6;
+  font-size: 1.0625rem;
+}
+
+.phoneInput {
+  border: none;
+  background: $color-white;
+  padding-left: 5px;
+  margin: 0; /* to avoid Safari for adding extra margin */
```
richardxia, line 52: This is a little bit surprising, since I thought our [reset stylesheet](https://github.com/ShelterTechSF/askdarcel-web/blob/7e4103e80cedfd96566c15f0800eac9913c86fc0/app/styles/utils/_reset.scss) would have taken care of cross-browser differences like this. Looking at it more closely, it seems like we created a custom reset stylesheet by mixing together multiple other ones, so it's not a vanilla reset stylesheet, and it has no reset styles for `input` elements.

Could you add a style to that file to reset all `input` element margins to 0? This could potentially have a large effect on the entire site, so we may want to look through the rest of the pages on the site to make sure we don't cause any major issues elsewhere. We may even want to do this as a completely separate PR before this PR merges in, just so we take care of it as a single, self-contained change.

app/components/Texting/components/FormView/Form.module.scss
```diff
@@ -27,10 +27,38 @@
...
+  box-shadow: none !important;
```
richardxia, line 53: Please avoid using `!important` when unnecessary, since it's a large hammer to wield, and it's difficult to override the effects of an `!important` later on.

I'm assuming you wrote this because you're running into an issue with [CSS specificity](https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity). More specifically, I'm assuming that the `:hover` pseudo-class style on the base `input` element is causing its `box-shadow` property to take priority over your non-pseudo-class `.phoneInput` selector here.

To achieve what you want in a safer way, all you need to do is escalate the specificity of your selector to be greater than that of the `input:hover` selector. Tag selectors have the lowest priority, followed by class and pseudo-class selectors, and then followed by ID selectors. If you have multiple selectors of the same level, then they take priority over ones with fewer selectors.

For example, `input` has less specificity than `.phoneInput`, since `input` is a tag selector and `.phoneInput` is a class selector. `input:hover` has higher specificity than `.phoneInput`, though, because `input:hover` has one level 2 selector (`:hover`) and one level 1 selector (`input`) while `.phoneInput` only has one level 2 selector (`.phoneInput`).

However, `.phoneInput:hover` would win over `input:hover` because `.phoneInput:hover` has _two_ level 2 selectors (`.phoneInput` and `:hover`).

If you just override the `box-shadow` style within a pseudo-class selector of the `.phoneInput`, then you'll be able to successfully take priority over the `input:hover` selector. Note that our [actual CSS style for setting the `box-shadow` on input elements targets three pseudo-classes](https://github.com/ShelterTechSF/askdarcel-web/blob/7e4103e80cedfd96566c15f0800eac9913c86fc0/app/styles/st-base/_forms.scss#L20-L22) — `:hover`, `:active`, and `:focus` — so you'll have to override all three.

app/components/Texting/components/FormView/Form.module.scss
```diff
@@ -27,10 +27,38 @@
...
+  /* Remove Chrome's autocomplete background color for input */
+  &:-webkit-autofill {
+    -webkit-box-shadow: 0 0 0 100px white inset !important;
+  }
+}
+
 .input {
   background-color: $color-white;
-  height: 50px;
-  padding: 15px 20px;
```
richardxia, line 33: What's the reason you decided to get rid of this?

app/components/Texting/components/FormView/FormView.jsx
```diff
@@ -1,20 +1,26 @@
...
-  phoneNumber: '',
   agreed: false,
 };
 
 const FormView = ({ service, handleSubmit, closeModal }) => {
   const [state, setState] = useState(initialState);
-  const { recipientName, phoneNumber, agreed } = state;
+  const { recipientName, agreed } = state;
+  const [phoneNumber, setPhoneNumber] = useState('');
+  const [isValidNumber, setIsValidNumber] = useState(false);
```
richardxia, line 21: I don't think you want to create a separate state hook to hold whether the number is valid. State is supposed to be the single source of truth of the state of a component, and but here, whether a phone number is valid is always derived from the number itself. Adding a separate state hook only increases fragility, since it allows for the following invalid states to be represented:

- `phoneNumber` holds a value that is valid, but `isValidNumber` is false
- `phoneNumber` holds a value that is invalid, but `isValidNumber` is true

Since whether the phone number is valid can be fully derived from the number itself, you should just derive it from the number instead of creating a separate store for it.

Could you just get rid of this state hook and just compute `isValidNumber` from `phoneNumber` as a normal variable?

app/components/Texting/components/FormView/FormView.jsx
```diff
@@ -55,13 +68,17 @@ const FormView = ({ service, handleSubmit, closeModal }) => {
...
-            type="text"
-            name="phoneNumber"
-            className={styles.input}
-            value={phoneNumber}
-            onChange={onChange}
-          />
+          <div className={styles.phoneInputBox}>
+            <span className={styles.phoneInputPrefix}>+1</span>
+            <input
+              type="text"
```
richardxia, line 74: I know that the previous version of this code didn't do this either, but could you change the type to [`tel`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/tel)? This is a special input element type that mostly looks like a normal text box, but on most mobile devices will cause a numeric keypad to pop open instead of a full text keyboard, which makes it easier to enter numbers.

Also, now that I'm looking at the docs for the `tel` input type, I'm seeing that it actually has a builtin `pattern` attribute for specifying a regex for validating the field, so maybe we should be using that instead of building our own custom JavaScript logic for it? Sorry for not remembering about that attribute beforehand 😅.

app/utils/validatePhoneNumber.js
```diff
@@ -0,0 +1,26 @@
...
+ * Validator for US only phone numbers.
+ */
+
+// eslint-disable-next-line import/no-unused-modules
+const PHONE_REGEX = /^\(?(\d{3})\)?[- ]?(\d{3})[- ]?(\d{4})$/;
+
+module.exports = {
```
richardxia, line 10: Please use ES6-style exports (i.e. the `export` keyword) instead of CommonJS exports (i.e. `module.exports`). Our app is not a Node.js application, so we don't need compatibility with that ecosystem, and even if we ever did need to, because we run our app through Webpack anyway, we could always ask Webpack to output a Node.js-compatible bundle, even if we use ES6 exports internally.

app/utils/validatePhoneNumber.js
```diff
@@ -0,0 +1,26 @@
+
+/**
+ *  @module normalize/validatePhoneNumber
+ * Validator for US only phone numbers.
+ */
+
+// eslint-disable-next-line import/no-unused-modules
```
richardxia, line 7: Please don't silence this lint rule. It's put there to prevent us from leaving around any code that is unused. If it goes off, it means that any publicly exported name isn't actually being used anywhere else in the app.

app/utils/validatePhoneNumber.js
Akhtam: ohh I tried to write some documentation but didn't realize I did it wrong but I did copy Regex's from stackoverflow.

app/components/Texting/components/FormView/Form.module.scss
Akhtam: I thought we don't need them since default input styling has padding and there wasn't big difference, If we have to keep it please let me know.

PR #1039, opened 2021-04-16
Upload Eviction Prevention Resource guides

lexholden:
https://app.clubhouse.io/sheltertech/story/1876/upload-eviction-prevention-resource-guides

Currently no design here, just a quick dirty working version to test this approach works.

Looking for feedback on:

- Should we have a "Resource Guides" tab listing all (that we can easily add too) in the header?
- Pretty sure url-loader doesn't mangle names for file resources, so just getting the /resource_guides/:id id directly from the webpack import path should be OK, need to test on a prod build though. Feels a bit janky, but minimizes developer effort to add new guides
- No design here at all so far. If we like this approach @derekfidler or anyone opinionated open to more robust design input here, or can just polish in line with other pages when it comes to that.

`sfserviceguide.org/resource_guides`:
image

`sfserviceguide.org/resource_guides/edc-guide-032021`:
image


app/pages/ResourceGuides/ResourceGuide.jsx
```diff
@@ -0,0 +1,38 @@
...
+import styles from './ResourceGuides.module.scss';
+import Footer from '../../components/ui/Footer/Footer';
+
+export const ResourceGuide = () => {
+  const { id } = useParams();
+  const [guide, setGuide] = useState(null);
+  const [loading, setLoading] = useState(true);
+  // const items = useMemo(() => Object.entries(ResourceGuidesLookup), []);
+
+  useEffect(() => {
```
richardxia, line 13: We had talked briefly on Slack about this, but I believe the `useEffect()` is unnecessary here, since everything that it's doing can just be done directly in the normal function body with no performance penalty.

app/routes.jsx
```diff
@@ -33,6 +34,8 @@ export default () => (
   <Switch>
     <Route exact path="/" component={HomePage} />
     <Route path="/about" component={About} />
+    <Route path="/resource_guides/:id" component={ResourceGuide} />
+    <Route path="/resource_guides" component={ResourceGuides} />
```
richardxia, line 38: Nit: Use dashes instead of underscores in URLs, since they're better for SEO purposes. Dashes are treated as word separators, while underscores cause the whole thing to be treated as a single word.

richardxia (commented):
I'm holding off from leaving a full review, since you're mainly looking for high level feedback on your direction.

> Should we have a "Resource Guides" tab listing all (that we can easily add too) in the header?

I'm indifferent to this, but I think it's probably a Product/Design team call, and most of them aren't on GitHub, so it's probably better to ask them on Clubhouse and/or Slack. Also, we should make sure that it doesn't cause the navigation header to be too cramped on mobile/tablet, since I'm not sure if our current breakpoints are set up to handle that many items in the navigation while also having the search bar up there.

> Pretty sure url-loader doesn't mangle names for file resources, so just getting the /resource_guides/:id id directly from the webpack import path should be OK, need to test on a prod build though. Feels a bit janky, but minimizes developer effort to add new guides

Since you've already assigned a "name" to each of these PDFs in that `ResourceGuidesData` data structure, maybe it'd be better to use the name as the identifier rather than the file name? That way, we'll have much more control over the IDs and not be dependent on how `file-loader` is configured. Also, `file-loader` is deprecated anyway, and we'll be forced to move onto the Webpack v5 assets system eventually, even if we're still running Webpack v4 today.

> No design here at all so far. If we like this approach @derekfidler or anyone opinionated open to more robust design input here, or can just polish in line with other pages when it comes to that.

If we don't get anyone from Design to comment, I can provide some lightweight feedback here. I think the overall simplicity is fine for both pages, since whole point is to just get the PDF there, but I think we'll probably want a bit more spacing between elements, since it looks a bit claustrophobic. Also, we'll probably want to apply the same max-width, centered, single column layout that we do on other pages, since from eyeballing your screenshots, I think your browser is at a width where our max width should have kicked in.

lexholden:
@richardxia since the objective is to get this out the door ASAP, have removed the nav link for now.

We can at least share the links directly, and create a followup ticket with a more considered design.

Content to merge this with a review unless there are objections.

app/components/ui/Navigation/Navigation.jsx
```diff
@@ -99,6 +99,11 @@ class Navigation extends React.Component {
                       About
                     </Link>
                   </li>
+                  {/* <li>
+                    <Link to="/resource-guides">
+                      Resource Guides
+                    </Link>
+                  </li> */}
```
richardxia, line 106: I think I'd prefer that we just remove this altogether until we actually want it. Having this here but commented out doesn't really provide any information to help us when we actually want to add this. Also, there's no equivalent change to the mobile version of the navigation menu, so it might even cause us to forget to add it to mobile later.

app/pages/ResourceGuides/ResourceGuides.module.scss
```diff
@@ -0,0 +1,20 @@
+.resourceWrapper {
+  margin: 0 auto;
+  padding: 26px;
```
richardxia, line 3: Do you think you could just add a `max-width: 1160px` here? I think that'll be enough to make the page not look weird on a wide browser window.

richardxia (approved):
Since you're failing a few CI link checks, I thought I'd make a few requests for changes. Even if we want this out ASAP, I still think it's better to try to uphold code quality on code we're merging, since we don't have the greatest track record for circling back on TODOs. Do you think you could make the changes as you address the lint failures?

This PR otherwise LGTM.

PR #1040, opened 2021-04-19
Typescript

lexholden:
WIP


webpack.config.js
```diff
@@ -71,23 +72,22 @@ module.exports = {
       CONFIG: config,
       NODE_ENV: JSON.stringify(process.env.NODE_ENV),
     }),
+    new ForkTsCheckerWebpackPlugin(),
```
richardxia, line 75: If we're using `@babel/preset-typescript`, then do we need the TypeScript type checking to be run as part of webpack? Although the project setup in sheltertech.org is different, due to it being built off the Gatsby framework, the way we have it set up there is that Gatsby itself is using `@babel/plugin-transform-typescript`, which just strips away the types, and we just have a separate [npm script to run `tsc` by itself](https://github.com/ShelterTechSF/sheltertech.org/blob/be5eb9fa689ea228c59f21e7f405a70e3b42d816/package.json#L14). You can run it with `npm run -- tsc --watch` to have it run a daemon that watches for file changes.

I think keeping it separate would make the Webpack setup simpler and have fewer moving parts, since the TypeScript side of it would be entirely standalone. Also, for people using IDEs or language servers, they don't even need Webpack running TypeScript since their editor will be doing it itself. I think `fork-ts-checker-webpack-plugin` makes a bit more sense to use if we were using `ts-loader`, since then we would already be depending on the TypeScript compiler within Webpack, but in the setup you have here, `@babel/plugin-transform-typescript` removes the need to have the TypeScript compiler in the loop at all.

richardxia (commented):
Thanks for getting the ball rolling on this, and I also appreciate the conversion from class-based components to function/hook-based components! I think most of these changes look good to me, and most of my bigger questions are about the tsconfig/webpack config.

Maybe one other thing we might want to include in this PR is an ESLint setup for the TypeScript files? If you wanted to see how we have that set up in the sheltertech.org project, here's the [.eslintrc.js file](https://github.com/ShelterTechSF/sheltertech.org/blob/be5eb9fa689ea228c59f21e7f405a70e3b42d816/.eslintrc.js).

app/init.tsx
```diff
@@ -17,11 +17,11 @@ if (process.env.NODE_ENV === 'production') {
   Sentry.init({ dsn: `https://${config.SENTRY_PUBLIC_KEY}@sentry.io/${config.SENTRY_PROJECT_ID}` });
 } else {
   // If Sentry is not enabled, use console instead.
-  Sentry.captureException = e => console.error(e);
-  Sentry.captureMessage = m => console.error(m);
+  (Sentry as any).captureException = (e: any) => console.error(e);
+  (Sentry as any).captureMessage = (m: any) => console.error(m);
```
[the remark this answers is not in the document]
lexholden: Sentry bundles its own types. The issue here is that these are marked `@readonly`, so TS complains at setting them.

Though pushing a better approach anyway, Sentry has a built in beforeSend hook I didn't know about.

app/pages/Covid/Covid.tsx
```diff
@@ -77,13 +77,11 @@ const CovidRoutes = () => {
   );
 };
 
-const CovidPages = () => (
+export const CovidPages = () => (
```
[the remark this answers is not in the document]
lexholden: My usual objection to it in large projects, it encourages renaming components when you import them, which is a pain to trace and debug.

But not militant about it, either way we should pick one convention in the whole codebase and work towards it. It's pretty inconsistent ATM.

tsconfig.json
```diff
@@ -0,0 +1,23 @@
...
+    "moduleResolution": "node",
+    "noEmit": true,
+    "noEmitOnError": true,
+    "resolveJsonModule": true,
+    "skipLibCheck": true,
+    "strict": true,
+    "strictPropertyInitialization": false,
+  },
+  "include": ["app"]
+}
```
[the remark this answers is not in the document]
lexholden: Good idea. I just set these to my own personal preference. Changing.

webpack.config.js
lexholden: @babel/preset-typescript also doesn't check types, just strips them away, so by itself we would rely only on the developer using a IDE to check types for them. Not convinced that helps much

ForkTsCheckerWebpackPlugin already runs in the background in a separate process, and is non blocking, so it makes sense to me to keep otherwise it would be perfectly easy to develop valid JS, push and merge a PR, that would not compile as typescript.

app/init.tsx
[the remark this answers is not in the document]
lexholden: Actually, without a DSN, the other approach (custom transport) doesn't work anyway, so sticking with this unpleasant hack for now.

app/pages/Covid/Covid.tsx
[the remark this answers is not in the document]
richardxia: > My usual objection to it in large projects, it encourages renaming components when you import them, which is a pain to trace and debug.

Fair point, although I think most editors have powerful enough static analyzers to resolve the imported name to the original definition, even if they mismatch. I say this as vim user who somewhat recently started using language servers.

But I'm also not militant and would prefer consistency over anything else. Maybe we can go with always using named exports, but also having the convention that the "main" export should have the same name as the file? This specific line is a bit annoying because in order to import it, you have to remember that the file name is `Covid` but that the named import is `CovidPages`.

app/pages/legal/TermsOfService.jsx
```diff
@@ -46,7 +44,7 @@ PLEASE READ THE FOLLOWING TERMS OF USE (“TERMS”) CAREFULLY BEFORE USING SF S
...
-    2. ShelterTech is not responsible for any special, indirect, incidental or consequential damages (regardless of whether they were foreseeable) that may arise from the use of, or the inability to use SF Service Guide, whether the materials contained in SF Service Guide are provided by sheltertech or a third party.
+    2. ShelterTechis not responsible for any special, indirect, incidental or consequential damages (regardless of whether they were foreseeable) that may arise from the use of, or the inability to use SF Service Guide, whether the materials contained in SF Service Guide are provided by sheltertech or a third party.
```
richardxia, line 47: Looks like you accidentally dropped a space here.

```suggestion
    2. ShelterTech is not responsible for any special, indirect, incidental or consequential damages (regardless of whether they were foreseeable) that may arise from the use of, or the inability to use SF Service Guide, whether the materials contained in SF Service Guide are provided by sheltertech or a third party.
```

webpack.config.js
richardxia: > @babel/preset-typescript also doesn't check types, just strips them away, so by itself we would rely only on the developer using a IDE to check types for them. Not convinced that helps much

Right, I meant to suggest that we directly run `tsc` in CI to type check the project so that it's still strictly enforced. I think I still prefer that over bundling it into the webpack process, since you get flexibility of running them separately or simultaneously. I may be overly cautious about throwing more stuff into webpack because I've been feeling a lot of pain of trying to do upgrades of libraries when there are a bunch of interdependencies that have to be bumped at the same time.

That said, I'm not adamant against adding this, since the actual integration itself is just adding a single line to the webpack plugins list, without requiring any extra configuration, so if you feel that this is important, then it's fine with me.

app/init.tsx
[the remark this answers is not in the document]
richardxia: Ah, makes sense, sounds good to me.

app/pages/Covid/Covid.tsx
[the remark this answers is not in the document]
lexholden: Yep 100% agree on naming. I wrote a simple custom eslint rule ages ago to enforce that `file-name-matches-class-name` that I use in most of my personal projects with service classes, can probably adapt to enforce it for functional react components easily.

For now, will rename the files I've converted to typescript and can adapt the others as we go along. `Covid` -> `CovidPages` - Though for consistency, would probably go for `CovidPage` or `CovidInfoPage`)

webpack.config.js
lexholden: Will leave for now then, if it causes bloat or notable slowdowns in contributors development, can easily remove at a later date.

lexholden:
@richardxia since there are a lot of TS bad practises here right now, and the js will fail many  of the rules in https://github.com/ShelterTechSF/sheltertech.org/blob/be5eb9fa689ea228c59f21e7f405a70e3b42d816/.eslintrc.js right now, I'd suggest treating that as a followup issue once more of this code has been converted to typescript.

richardxia:
> @richardxia since there are a lot of TS bad practises here right now, and the js will fail many of the rules in https://github.com/ShelterTechSF/sheltertech.org/blob/be5eb9fa689ea228c59f21e7f405a70e3b42d816/.eslintrc.js right now, I'd suggest treating that as a followup issue once more of this code has been converted to typescript.

Oh yeah, of course, I should have expected that. Yes, this sounds reasonable to me.

richardxia (approved):
This LGTM, as soon as the CI failures are fixed, which I think are just due to the package.json and lock files getting out of sync. Thanks again for working on this!

lexholden:
Good point, the last builds I'd seen fail were due to Travis limits, so had stopped checking them, assuming they were failing always.

I added them locally with yarn, not npm, since it tends to be faster and cache better. Pushing a fix with updated package-lock.

app/components/App.tsx
```diff
@@ -179,22 +171,21 @@ class App extends Component {
           <div className="container">
             <Routes />
           </div>
-          <PopUpMessage />
```
lexholden, line 182: This failed typechecks, because it expects a prop, and is actually hidden unless the prop is passed, so I see no harm in removing it.

app/components/App.tsx
```diff
@@ -94,45 +93,38 @@ const getLocationGoogle = () => new Promise((resolve, reject) => {
  * @returns A Promise of a location, which is either an object with `lat` and
  * `lng` properties or an error if location is unavaible or out of bounds.
  */
-const getLocation = () => new Promise((resolve, reject) => {
```
lexholden, line 97: This just seemed overly redundant. Tested and works fine.

app/components/App.tsx
```diff
@@ -59,7 +58,7 @@ const getLocationBrowser = () => new Promise((resolve, reject) => {
     console.log(msg);
     reject(msg);
   }
-}, { timeout: 10000 });
```
lexholden, line 62: Never seen this before, but is not in the promises API, so if it was ever supported it isn't now.

.eslintrc.js
```diff
@@ -35,8 +33,48 @@ module.exports = {
     'react/forbid-prop-types': 'off',
     'react/prefer-stateless-function': 'off',
     'react/prop-types': 'off',
+    'react/jsx-filename-extension': ['error', { extensions: ['.tsx', '.jsx'] }],
   },
   overrides: [
+    {
```
lexholden, line 39: Medium term, we should consolidate these and apply the same rules everywhere, but there are many issues that fail in JS files, so for now this was required to get linting to work.

app/components/App.tsx
richardxia: Hmm, I took a look at the component definition, and I think your statement is incorrect. It looks like it uses a Redux `connect()`, so the props are actually injected by Redux.

.eslintrc.js
richardxia: SGTM!

app/components/App.tsx
richardxia: Weird, maybe it was a bad refactor or merge conflict resolution in the past. Definitely glad that TypeScript is helping us catch these kinds of things!

app/components/App.tsx
lexholden: Good point, fixing.

richardxia (approved):
LGTM! Thanks for getting CI to all pass. Let's merge this ASAP, before anything else conflicts with it.

PR #1041, opened 2021-04-27
Add basic support for Github Actions to get around Travis CI limits and speed issues

lexholden:


.github/workflows/ci.yml
```diff
@@ -0,0 +1,68 @@
...
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+    strategy: { matrix: { node_version: [10.x] } }
```
richardxia, line 15: Is it worth it to have the version Matrix? We're not a library, so we only really need to support one version of Node.js. Might also make the file more legible to just hardcode the node version into that `node-version` field.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,68 @@
...
+          node-version: ${{ matrix.node-version }}
+      - name: Cache dependencies
+        uses: actions/cache@v2
+        with:
+          path: ~/.npm
+          key: npm-${{ hashFiles('package-lock.json') }}
+          restore-keys: npm-
+      - run: npm ci
+      - run: cp config.example.yml config.yml
+      - run: npm run lint
```
richardxia, line 30: Is it worth it to have a standalone step for linting? I have a feeling that we'll spend more time on installing a specific version of Node.js and installing the dependencies, even with caching.

richardxia (commented):
I added the Sauce Labs credentials to the repository so that you can access them from the GH Action. The username is under the secret named `SAUCE_USERNAME` and the password is under the secret named `SAUCE_ACCESS_KEY`. Looking at the Travis CI job, I think you just need to make sure that these secrets are exported into the environment for our scripts to pick them up.

There are a ton of other secrets defined in Travis, but I'm not sure how much is vestigial and how much we actually actually still depend on. I think we'll need the Docker Hub credentials, since that's what allows us to actually upload our `master` images. I don't think I have our Docker Hub credentials, so maybe @lgarofalo or @jjfreund know?

.github/workflows/ci.yml
lexholden: Yep good point. Will remove.

Though we are using a version of node that is now no longer even in maintenance mode as of May, seems worth a separate effort to upgrade to 16 or 17 (the current LTS). Will consider that separately.

.github/workflows/ci.yml
lexholden: Was planning to also add 'depcheck' here to run at this step to ensure we don't have a bunch of unused packages.

Personally, I prefer different status reports for different things, my issue with the current implementation being that if you have issues in linting and the e2es, you don't notice the latter until the former has been fixed.

But yeah I need to figure out how to bring down the `npm ci` step down more generally, possiblly caching the build of node-sass or using the javascript implementation of node-sass, not the c one - the performance difference will be negligible for such a small project.

If I can't get the install/cache steps down to sub 1m, then agreed seems worth merging at least unit and lint.

Finding time to work on this again this week.

.github/workflows/ci.yml
richardxia: Sounds good. BTW, I did a very lightweight test of upgrading to Node 16, and it looks like we'll probably need to bump the version of node-sass we're using, since it's so old that node-sass doesn't have a precompiled version of itself for Node 16.

PR #1044, opened 2021-05-05
New and edit address modal forms

richardxia:
Implements https://app.clubhouse.io/sheltertech/story/1445/add-new-address-and-edit-address-modals
Figma design: https://www.figma.com/file/tBRClGsxYN9m6ZvL2STB6t/WIP-Edit-Redesign---Unscheduled?node-id=2839%3A134

This replaces the old address forms from the Edit page with the new modal-based ones. See  for the designs that this is roughly following, though this PR is focused just on the edit address forms for Organizations, and not the related form for Services or any of the other changes.

This ended up being a pretty substantial rewrite of the edit address forms, since it switches away from the `editCollectionHOC` higher-order component that we had for automatically generating components for editing collections of objects to a new set of components.

A few of the more notable changes include:
- The new form consists of uncontrolled form components, rather than controlled components. This diverges from our usual style in the rest of the codebase, but I chose to do it this way to enable the "cancel" button in the modal, which should throw out changes. With controlled components, I would have needed to persist both the state for the form before and during editing, since I would need the old state around in order to roll back to. With uncontrolled components, I get the extra state for "free" because the native form elements retain their own state, and I only persist that state to the root React component when the "save" button is clicked.
- The "No Physical Location" checkbox no longer exists in the new designs, so I removed it and all code related to it, which happened to be quite a lot. This is all self-contained in https://github.com/ShelterTechSF/askdarcel-web/commit/730b5616ff8999b8c31ce6ca9370492a20b0c1d5

Big warning: This can't actually merge as is today, since we have a contradiction between the Figma designs and the askdarcel-api database schema. The new designs omit the Country field from addresses, but the askdarcel-api DB schema requires a country to be entered in. Should we lift that restriction and allow the country field to be null?

## Screenshots

Screen Shot 2021-05-04 at 11 19 40 PM
Screen Shot 2021-05-04 at 11 19 19 PM
Screen Shot 2021-05-04 at 11 19 51 PM

### DIFF
--- a/app/components/edit/EditAddress.jsx
+++ b/app/components/edit/EditAddress.jsx
@@ -213,17 +213,21 @@ const EditAddresses = ({ addresses, setAddresses }) => {
 
       <div className={s.addressListTitle}>Location</div>
       <div className={s.addressList}>
-        {addresses.map((address, i) => (
-          !address.isRemoved && (
+        {addresses
+          .map((address, arrayIndex) => [address, arrayIndex])
+          .filter(([address]) => !address.isRemoved)
+          // The displayIndex should skip over any elements that have been
+          // marked for removal, but the arrayIndex must be the actual index of
+          // the element in the addresses array
+          .map(([address, arrayIndex], displayIndexMinusOne) => (
             <AddressListItem
               key={address.id || JSON.stringify(address)}
-              displayIndex={i + 1}
+              displayIndex={displayIndexMinusOne + 1}
               address={address}
-              onEdit={() => setModalState({ type: 'edit', editingIndex: i })}
-              onRemove={() => removeAddress(i)}
+              onEdit={() => setModalState({ type: 'edit', editingIndex: arrayIndex })}
+              onRemove={() => removeAddress(arrayIndex)}
             />
-          )
-        ))}
+          ))}
       </div>
 
       <button
--- a/app/components/edit/EditAddress.jsx
+++ b/app/components/edit/EditAddress.jsx
@@ -1,6 +1,118 @@
-import React from 'react';
+import React, { useState } from 'react';
+import ReactModal from 'react-modal';
 import PropTypes from 'prop-types';
 import editCollectionHOC from './EditCollection';
+
+import s from './EditAddress.module.scss';
+
+// Subcomponents
+
+/** Modal containing the edit and add new address form.
+ *
+ * The form data is only synchronized with the outer component's state when the
+ * form is submitted. This allows the user to back out of any changes with the
+ * "Cancel" button. Therefore, we have chosen to implement the form inputs as
+ * uncontrolled components.
+ */
+const EditAddressModal = ({
+  isOpen, onRequestClose, defaultData, onSave,
+}) => {
+  const handleSubmit = e => {
+    e.preventDefault();
+    const formData = new FormData(e.currentTarget);
+    onSave(Object.fromEntries(formData.entries()));
+    onRequestClose();
+  };
+  return (
+    <ReactModal
+      className={s.modal}
+      overlayClassName={s.overlay}
+      isOpen={isOpen}
+      onRequestClose={onRequestClose}
+    >
+      <form className={s.modalContent} onSubmit={handleSubmit}>
+        <h1 className={s.title}>Add New Address</h1>
+        <div className={s.formBody}>
+          <label className={s.inputLabel}>
+            Address name (optional)
+            <input
+              className={s.inputTextbox}
+              type="text"
+              name="name"
+              placeholder="Name"
+              defaultValue={defaultData.name}
+            />
+          </label>
+          <div className={s.formControlGroup}>
+            <span className={s.formControlGroupLabel}>Street Address</span>
+            <input
+              className={s.inputTextbox}
+              type="text"
+              name="address_1"
+              placeholder="Address"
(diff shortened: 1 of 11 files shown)


app/components/edit/EditAddress.jsx
```diff
@@ -1,7 +1,119 @@
...
+              className={s.inputTextbox}
+              type="text"
+              name="name"
+              placeholder="Name"
+              defaultValue={defaultData.name}
+            />
+          </label>
+          <div className={s.formControlGroup}>
+            <span className={s.formControlGroupLabel}>Street Address</span>
+            <input
```
richardxia, line 55: Note to self: Make all of these field required, since these are uncontrolled components.

katerina-kossler (approved):
walked through this with Richard, some small issues to fix but otherwise looks good!

app/components/edit/EditAddress.jsx
```diff
@@ -1,129 +1,252 @@
...
+            />
+          </label>
+        </div>
+        <div className={s.actionRow}>
+          <button
+            className={s.cancelButton}
+            type="button"
+            onClick={onRequestClose}
+          >
+            ← Cancel
```
lexholden, line 111: `←` is probably better as an equivalent material icon here, I know the raw `←` character often renders weirdly off-center relative to the text on a the same line, its baseline is rarely perfectly centered.

app/components/edit/EditAddress.jsx
```diff
@@ -1,129 +1,252 @@
...
+            />
+          ))}
+      </div>
+
+      <button
+        className={s.newAddressButton}
+        type="button"
+        onClick={() => setModalState({ type: 'add' })}
+      >
+        Add location
```
lexholden, line 242: nit: "Add Location" - other buttons are title case

app/components/edit/EditAddress.jsx
```diff
@@ -1,129 +1,252 @@
...
+              className={s.inputTextbox}
+              type="text"
+              name="state_province"
+              placeholder="State"
+              required
+              defaultValue={defaultData.state_province}
+            />
+          </label>
+          <label className={s.inputLabel}>
+            Zip code
```
lexholden, line 94: Nit: Other labels are title case "Zip Code"

app/components/edit/EditAddress.jsx
```diff
@@ -1,129 +1,252 @@
...
+      className={s.modal}
+      overlayClassName={s.overlay}
+      isOpen={isOpen}
+      onRequestClose={onRequestClose}
+    >
+      <form className={s.modalContent} onSubmit={handleSubmit}>
+        <h1 className={s.title}>{title}</h1>
+        <div className={s.formBody}>
+          <label className={s.inputLabel}>
+            Address name (optional)
```
lexholden, line 44: Nit: Other labels are title case "Address Name (Optional)"

lexholden (approved):
A couple of nits, but broadly LGTM.

app/components/edit/EditAddress.jsx
richardxia: Yeah, good point. I swapped it out with a Remix icon, although it looks like our vertical alignment is still just a bit messed up in general due to a global `svg { vertical-align: middle }` CSS style that I'm not going to mess with right now. I would like to get rid of most of our global CSS at some point, since it's pretty dated and sometimes causes more issues with modern, idiomatic CSS.

Screen Shot 2021-05-23 at 1 15 23 PM

app/components/edit/EditAddress.jsx
```diff
@@ -1,129 +1,252 @@
...
+
+const AddressListItem = ({
+  displayIndex, address, onEdit, onRemove,
+}) => (
+  <div className={s.listItemContainer}>
+    <div className={s.listItemIndex}>
+      {`${displayIndex}.`}
+    </div>
+    <div className={s.listItemName}>{address.name}</div>
+    <div className={s.listItemAddress}>{compactAddressDisplay(address)}</div>
```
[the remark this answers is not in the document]
richardxia: I'm not sure we actually want to do that, since we actually want to remove all line breaks here to make it all fit in one line. I'm using `white-space: nowrap`, not `white-space: pre`, for the CSS of `listItemAddress`.

I'm going to leave this alone, but let me know if you feel strongly otherwise.

app/components/edit/EditAddress.jsx
richardxia: I had been following the mockup designs, which pretty consistently used sentence case, but you're right that this looks weird with some of the other stuff on the page being title case. I'll change this over, and if design prefers sentence case, we can take a pass through everything on the edit page to change that.

richardxia:
In addition to addressing some of @lexholden's comments, I added one more commit that wasn't previously reviewed, which is https://github.com/ShelterTechSF/askdarcel-web/pull/1044/commits/a00c7fd7399c4a771deca45d91d32d1d0d3655df. This hardcodes the value for "Country" to "USA", since it's currently a required field in the API, but we no longer have that field in the newest version of the address form.

I had talked to @katerina-kossler previously about updating the API app to make that field not required anymore, but in the interest of not blocking this PR on that task, I went ahead and made this change here, which we can revert once the API no longer requires a country.

Finally got Travis CI to pass (after paying Docker to not rate limit us), so I'm going to merge this in now.

PR #1048, opened 2021-05-07
Fix submit button bug feedback modal

Akhtam:
- Removed feedback(review) requirement if user chooses `Other` tag and make it optional, product team want it to be a lightweight feedback experience.
Screen Shot 2021-05-07 at 8 07 23 PM
Screen Shot 2021-05-07 at 8 07 33 PM

- We currently don't have a way of reaching back out so we had to remove sub-header message **"Team will usually replies within a day"**

- Removed unnecessary styling for mobile version

### DIFF
--- a/app/components/listing/feedback/FeedbackModal.jsx
+++ b/app/components/listing/feedback/FeedbackModal.jsx
@@ -71,11 +71,6 @@ const FeedbackModal = ({ service, resource, closeModal }) => {
       .catch(err => console.log(err));
   };
 
-  const isReviewRequired = (
-    tagOptions.some(({ tag, selected }) => tag === 'Other' && selected)
-    && vote === DOWNVOTE
-  );
-
   const STEPS = {
     tags: (
       <FeedbackTags tagOptions={tagOptions} onSelectTag={toggleSelectedTag} />
@@ -83,7 +78,6 @@ const FeedbackModal = ({ service, resource, closeModal }) => {
     review: (
       <Review
         reviewValue={review}
-        isReviewRequired={isReviewRequired}
         onReviewChange={handleReviewChange}
       />
     ),
@@ -115,7 +109,6 @@ const FeedbackModal = ({ service, resource, closeModal }) => {
             onPrevStep={handlePrevStep}
             onNextStep={handleNextStep}
             onSubmit={handleSubmit}
-            isReviewRequired={isReviewRequired}
             isSubmitted={isSubmitted}
           />
         </div>
--- a/app/components/listing/feedback/FeedbackModal.jsx
+++ b/app/components/listing/feedback/FeedbackModal.jsx
@@ -103,9 +103,6 @@ const FeedbackModal = ({ service, resource, closeModal }) => {
         <img src={images.icon('feedback-blue-header')} alt="feedback" />
         <span>Share your Feedback</span>
       </div>
-      <div className={styles.feedbackSubheader}>
-        The team usually replies within a day.
-      </div>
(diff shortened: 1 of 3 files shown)


app/components/listing/feedback/FeedbackModal.jsx
```diff
@@ -103,9 +97,6 @@ const FeedbackModal = ({ service, resource, closeModal }) => {
         <img src={images.icon('feedback-blue-header')} alt="feedback" />
         <span>Share your Feedback</span>
       </div>
-      <div className={styles.feedbackSubheader}>
-        The team usually replies within a day.
-      </div>
```
richardxia, line 108: This doesn't seem like it has anything to do with the change to make the feedback box optional. Was this accidentally deleted or was this intentional?

If this was intentional, then please at least make a mention of this change in your PR description. Currently, the PR title only says that this fixes a "bug" in the submit button, and the PR description makes no mention of other changes being made. If you don't explicitly mention these kinds of things, then it becomes difficult for a code reviewer to understand whether a change was intentional or accidental.

app/components/listing/feedback/FeedbackSteps.jsx
```diff
@@ -52,17 +52,15 @@ export const FeedbackTags = ({ tagOptions, onSelectTag }) => (
...
     <div className={styles.stepsPrompt}>
       Please provide your feedback below:
     </div>
     <textarea
       className={styles.feedbackTextarea}
       type="text"
-      placeholder={`Type your feedback here ${
-        !isReviewRequired ? '(optional)' : ''
-      }`}
+      placeholder="Type your feedback here (optional)"
```
richardxia, line 63: I'm a little bit confused by the screenshot you included in your PR description. The screenshot shows a placeholder text of "Type your feedback here (required)," but neither the old code nor the new code show that text. Was that screenshot taken from some intermediate version of the code with the incorrect text?

Just wanted to double check that there isn't some other place in the code that determines the placeholder text shown in that screenshot, because if there is, we should update that text not not include the word "required."

app/components/listing/feedback/FeedbackSteps.module.scss
```diff
@@ -105,7 +105,6 @@
   width: 340px;
   margin-top: 30px;
   @media screen and (max-width: 767px ) {
-    position: absolute;
     bottom: 40px;
```
richardxia, line 108: Why was this change made? It's not clear how this relates to the other changes in this PR. Even if it were an intentional change, it's not obvious to a reviewer what behavior this is intended to change.

Also, if you did intentionally make this change, then you should also remove the `bottom: 40px`, since it has no effect when the `position` property is not set to one of the non-default values (`absolute`, `relative`, `fixed`, or `sticky`).

richardxia (commented):
These changes mostly look good to me, assuming that they are all intentional. If you intentionally made changes to things other than removing the feedback requirement, then please make sure it is explicitly called out in your PR description, because otherwise it's not possible for a reviewer to be 100% confident about the intent of the changes.

app/components/listing/feedback/FeedbackModal.jsx
Akhtam: Thank you for bringing this up, just updated description.

app/components/listing/feedback/FeedbackSteps.jsx
Akhtam: Just fixed it, didn't realized I uploaded wrong screenshot

PR #1051, opened 2021-05-09
Upgrade to testcafe 1.14.0.

richardxia:
I'm hitting some issues trying to fix the TestCafe tests in https://github.com/ShelterTechSF/askdarcel-web/pull/1044, but for some reason TestCafe doesn't seem to be able to detect some of the newer components on the page. Upgrading TestCafe to 1.14.0 seems to fix the issue for me, so I decided to submit a separate PR just to upgrade that.

I have not actually tested this on the whole test suite, so I am just hoping that this passes on Travis CI.


katerina-kossler (approved):
lgtm!

PR #1053, opened 2021-06-07
[DONOTMERGE] Make external link prepend https if not provided

lexholden:
https://app.clubhouse.io/sheltertech/story/1933/external-website-urls-don-t-work-without-explicitly-writing-http

A solution as requested to the above. I also just fixed the underlying data on the bad organization page, which I'd argue is a more robust solution, in which case we should not merge this.

### DIFF
--- a/app/components/listing/ResourceInfos.jsx
+++ b/app/components/listing/ResourceInfos.jsx
@@ -1,4 +1,4 @@
-import React, { Fragment } from 'react';
+import React, { Fragment, useMemo } from 'react';
 import PropTypes from 'prop-types';
 import _ from 'lodash';
 
@@ -109,9 +109,13 @@ PhoneNumber.propTypes = {
 };
 
 
-const ExternalLink = ({ children, to }) => (
-  <a href={to} target="_blank" rel="noopener noreferrer">{children}</a>
-);
+const ExternalLink = ({ children, to }) => {
+  const url = useMemo(() => {
+    if (!/^https+:\/\//.test(to)) { return `https://${to}`; }
+    return to;
+  }, [to]);
+  return <a href={url} target="_blank" rel="noopener noreferrer">{children}</a>;
+};
 
 ExternalLink.propTypes = {
   to: PropTypes.string.isRequired,


app/components/listing/ResourceInfos.jsx
```diff
@@ -109,9 +109,13 @@ PhoneNumber.propTypes = {
...
 
 
-const ExternalLink = ({ children, to }) => (
-  <a href={to} target="_blank" rel="noopener noreferrer">{children}</a>
-);
+const ExternalLink = ({ children, to }) => {
+  const url = useMemo(() => {
+    if (!/^https+:\/\//.test(to)) { return `https://${to}`; }
+    return to;
+  }, [to]);
```
richardxia, line 116: `useMemo` seems like overkill here. Why not just perform this lightweight check on every render? The entire rendering of this `ExternalLink` component will be memoized anyway if its props don't change, and I can't think of a single scenario where `children` would change without `to` also changing.

lexholden: I remember you mentioning this on a previous PR as well, I'm not really sure that useMemo has a big overhead under the hood, it compares the dependency list by referential equality on each render and decides whether to recalculate, unless I'm missing something.

I agree that a basic regex check and string interpolation is also not a major overhead to do on each render, but neither is the overhead used by useMemo to save a render here and again, so not sure I understand the reluctance to use or think that it adds much complexity.

richardxia: > I remember you mentioning this on a previous PR as well, I'm not really sure that useMemo has a big overhead under the hood, it compares the dependency list by referential equality on each render and decides whether to recalculate, unless I'm missing something.

Right, but React components already do this for props, so if the props don't change, the component is not rerendered. React is optimized and designed around doing this for props, so it's already fast.

> I agree that a basic regex check and string interpolation is also not a major overhead to do on each render, but neither is the overhead used by useMemo to save a render here and again, so not sure I understand the reluctance to use it, or think that it adds much complexity.

It adds complexity in that `useMemo()` is a more complex, more powerful construct that novice React programmers may not be familiar with, and in that you have to make sure the second argument, the dependencies list, captures all variables that should invalidate the memoization, even after future possible changes to the code.

I think https://kentcdodds.com/blog/usememo-and-usecallback probably does a better job of explaining this stance than I am, but let me quote (with emphasis retained) something from it:

> <b>MOST OF THE TIME YOU SHOULD NOT BOTHER OPTIMIZING UNNECESSARY RERENDERS</b>. React is VERY fast and there are so many things I can think of for you to do with your time that would be better than optimizing things like this.

lexholden: Fair enough, did not know this, but very interesting.

Regardless, @fredciaramaglia and myself have started fixing the bad data here, so closing this unmerged. Not a fan of this approach.

PR #1057, opened 2021-06-28
update github actions to push images

lgarofalo:


lexholden (approved):
Awesome, LGTM.

.github/workflows/ci.yml
```diff
@@ -30,6 +31,7 @@ jobs:
       - run: npm run lint
 
   test_unit:
+    needs: lint
```
richardxia, line 34: Why does this depend on the `lint` job? The reason they were split up in the first place was so that they can run in parallel. If we no longer want them to run in parallel, we should just move the `npm run test` step into the `lint` job, right under the lint step, since we won't have to set up the repository twice sequentially.

.github/workflows/ci.yml
```diff
@@ -11,6 +11,7 @@ on:
 
 env:
   NODE_VERSION: 10.x
+  DOCKER_REPO: sheltertechsf
```
richardxia, line 14: Nit: I believe "repo" refers to the combination of the org and the specific repo name, which would be `sheltertechsf/askdarcel-web` in this case. Maybe this variable should be named `DOCKER_ORG` instead? Alternatively, you could make the value `sheltertechsf/askdarcel-web`.

richardxia (approved):

.github/workflows/ci.yml
lexholden: Very good point. I missed this. 100% agree, I think these should stay parallel.

lgarofalo: Will return to prior

.github/workflows/ci.yml
lgarofalo: Will do

lgarofalo:
This didn't actually fix the publish, working on that for another PR

PR #1061, opened 2021-07-12
Regenerate package-lock.json from scratch to fix development environment

richardxia:
The Internet seems to believe that "no such file or directory, rename"
errors we are running into are due to a race condition in NPM, and that
deleting and regenerating package-lock.json works around the issue. This
indeed appears to fix the issue for us.

See https://github.com/WordPress/gutenberg/issues/9474#issuecomment-769978287


davidagustin:
It works and I manually ran the build and verified it myself

davidagustin (approved):
Ran the build locally with the instructions set forth in the readme and the build works.

PR #1062, opened 2021-07-12
Check for null addresses on ServiceDiscoveryResults page.

richardxia:
@jjfreund let me know that the ServiceDiscoveryResults pages break if an organization is missing any addresses, which might be related to the multiple addresses project regarding how orgs with no addresses are represented. Previously there would be a dummy, blank address on an org, but now it's probably just returning null.

This adds an extra check to the search results page so that we handle the case where Algolia returns null or undefined for the addresses property.

Tested this locally by creating an org with no addresses, which initially broke the guided pathways results page before this one-line change.


jjfreund (approved):
lgtm

PR #1065, opened 2021-07-25
Updates to banner link text

Raawr:


jjfreund:
Looks reasonable other than the fact that de-linting is needed.

PR #1067, opened 2021-08-12
Karla banner links

jjfreund:


app/components/ui/Banner/Banner.jsx
```diff
@@ -6,15 +6,16 @@ export default function Banner() {
...
       <br />
       <strong>&nbsp;Need help applying for rent payment assistance?</strong>
-      &nbsp;Check out Code Tenderloin&apos;s workshops&nbsp;
-      <a href="https://docs.google.com/document/d/1t-RM30QQHA5UQdzIGHrD6kaFoBMrwmCBM7mEOBD8a9U/edit">HERE</a>
-      <br />
-      and HomeownershipSF&apos;s resources&nbsp;
-      <a href="https://www.homeownershipsf.org/application-assistance-for-renters/">HERE</a>
+      &nbsp;Check out&nbsp;
+      <a target="_blank" rel="noreferrer" href="https://docs.google.com/document/d/1t-RM30QQHA5UQdzIGHrD6kaFoBMrwmCBM7mEOBD8a9U/edit">Code Tenderloin&apos;s workshops</a>
+      &nbsp;and&nbsp;
```
richardxia, line 16: This is causing weird line breaks, since you're preventing the words "workshops", "and", and "Homeownership" to be forced on the same line. Altogether they are fairly long words, which causes them all to wrap together and make the page look weird.

We shouldn't be using `&nbsp;` anywhere here in the banner. NBSP stands for non-breaking space, which means that the browser cannot insert a line break between the words that it joins. To add explicit spacing between text and HTML elements, you should just insert a JavaScript space literal using `{" "}`, as in `<a>workshops</a>{" "}and{" "}<a>Homeownership</a>`.

You should only ever use `&nbsp;` if you want to prevent a line break from being inserted, which is actually fairly rare.

richardxia (approved):

PR #1069, opened 2021-08-19
Centralize White Label Data

davidagustin:


lexholden (approved):
LGTM. This is much cleaner!

davidagustin:
I can sure use a manual verify also to see if I covered all the areas

app/utils/whitelabel.ts
```diff
@@ -1,31 +1,75 @@
...
-  }
-  if (isSFFamiliesSite()) {
-    return 'SF Families';
-  }
-  return 'AskDarcel';
+// Use url to get key of config object
+// @ts-ignore
+const configKey = Object.keys(config).find((key) => config[key] === window.location.host) || 'default';
+// @ts-ignore
+let configurations: any;
```
lexholden, line 13:
```suggestion
const configurations: any = {};
```

Hmm good point actually just tried this locally, and it errors because this object is undefined, causing a `Uncaught TypeError: Cannot set property 'title' of undefined` error

lexholden: When I fix that locally, everything looks OK to me. Can probably do some more thorough QA on staging after this is merged

lexholden (changes requested):

app/utils/whitelabel.ts
richardxia: I would highly recommend that we don't give this a type of `any`, since that disables type checking on this object, which in turn disables type checking of the entire `whiteLabel` object being exported to other files.

I would suggest that we "transpose" the structure of this object so that the top-level properties are just the different site names (e.g. `SFFAMILIES_DOMAIN`, or in my other suggestion above, `SFFamilies`) and that all the other properties are nested under the site. This would allow us to define a type or interface that describes all of the properties that need to be set for each site, and it would allow the compiler to yell if we're missing any required properties.

The issue with the current structure is that if we added another site, it would be pretty easy to forget to add the new site key to each of the individual top-level properties.

An example of what I'm suggesting is the following:

```ts
interface WhiteLabelSite {
  title: string;
  siteUrl: string;
  appImages: {
    // Fill out each image here
  };
}

const configurations: Record<WhiteLabelSiteKey, WhiteLabelSite> = {};

configurations.SFFamilies = {
  title: 'SF Families',
  siteUrl: 'https://sffamilies.sfserviceguide.org/',
  appImages: { /* fill out images here */ }
}
```

This gives you type safety, preventing you from forgetting or typo'ing any properties, and it'll have better IDE hover information.

app/utils/whitelabel.ts
```diff
@@ -1,31 +1,75 @@
...
-  isSFServiceGuideSite, isSFFamiliesSite, getSiteTitle, getSiteUrl,
+// @ts-ignore
+const icons = require.context('../assets/img', true, /ic-.*\.(png|svg)$/i);
+const iconPathMap = {};
+
+// @ts-ignore
+icons.keys().forEach((key) => { iconPathMap[key.match(/ic-([^@]*)(?:@3x)?.(?:svg|png)/)[1]] = icons(key); });
+
+// @ts-ignore
+const icon = (name) => iconPathMap[name.toLowerCase().replace(/(\s+|\/)/g, '-')];
```
richardxia, line 37: This icon stuff doesn't look like it has anything to do with whitelabeling, since these are generic to all whitelabeled sites. Could you move this back to `assets/index.js`?

As an aside, I think we should probably get rid of this stuff at some point, since it's not really necessary in a modern frontend project, since Webpack should be able to take care of all the assets transformation stuff, so we really only need to just import the individual icons that we're using like any other Webpack image import.

richardxia (commented):
I think this is a good step, but I'd highly recommend that we don't basically turn off all of the TypeScript type checking on `whitelabel.ts`, since the combination of `@ts-ignore` and the `: any` type annotations basically results in almost every variable in that file being untyped.

I'd also suggest we fully eliminate the `isSFFamiliesSite` variable by replacing it with the actual feature that we are gating for each conditional it's used in.

app/utils/whitelabel.ts
```diff
@@ -1,31 +1,75 @@
...
-  if (isSFServiceGuideSite()) {
-    return 'SF Service Guide';
-  }
-  if (isSFFamiliesSite()) {
-    return 'SF Families';
-  }
-  return 'AskDarcel';
+// Use url to get key of config object
+// @ts-ignore
+const configKey = Object.keys(config).find((key) => config[key] === window.location.host) || 'default';
```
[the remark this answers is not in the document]
davidagustin: Great reviews! I become a better engineer everytime. Will push up changes

app/pages/HomePage/components/ResourceCard/ResourceCard.jsx
```diff
@@ -47,7 +49,7 @@ const ResourceCard = ({ resource }) => {
       className={styles.card}
       {...anchorTagProps}
     >
-      <img src={images.icon(icon)} alt={name} className={styles.icon} />
+      <img src={appImageIcon(icon)} alt={name} className={styles.icon} />
```
davidagustin, line 52: Another Icon was declared before

app/utils/whitelabel.ts
```diff
@@ -1,31 +1,89 @@
...
-const getSiteTitle = () => {
-  if (isSFServiceGuideSite()) {
-    return 'SF Service Guide';
-  }
-  if (isSFFamiliesSite()) {
-    return 'SF Families';
-  }
-  return 'AskDarcel';
+// Read only to force developer to modify configurations here, disallow changes at compile time
+const configurations: Partial<Record<Readonly<WhiteLabelSiteKey>, Readonly<WhiteLabelSite>>> = {};
```
davidagustin, line 30: I want to make sure that this is the central place to modify whitelabel data so thus I made this read only.

app/utils/whitelabel.ts
```diff
@@ -1,31 +1,89 @@
...
+    background: BackgroundImage,
+    logoLarge: AskDarcelImage,
+    logoSmall: AskDarcelImage,
+    algolia: SearchByAlgoliaImage,
+    mohcdSeal: SFSeal,
+    icon,
+  },
 };
+
+// Disallow changes at run time
```
davidagustin, line 85: Also I want to make sure none of the objects are modified at run time to also encourage development in this file

app/pages/ServiceListingPage.tsx
```diff
@@ -42,7 +44,7 @@ export const ServiceListingPage = () => {
   return (
     <div className="listing-container">
       <Helmet>
-        <title>{`${service.name} | ${getSiteTitle()}`}</title>
+        <title>{`${service.name} | ${configurationsTitle}`}</title>
```
davidagustin, line 47: Collision of variable name here

app/components/ui/Navigation/Navigation.jsx
```diff
@@ -86,7 +87,7 @@ class Navigation extends React.Component {
               )
             }
           </div>
-          {!isSFFamiliesSite()
+          {title !== 'SF Families'
```
jjfreund, line 90: showMobileNav or something similar would be the configurable property in this case -- true for everyone but sffamilies.

jjfreund (changes requested):

app/components/search/SearchEntry.jsx
```diff
@@ -47,7 +49,7 @@ class SearchEntry extends Component {
               <h4 className="entry-headline">{`${hitNumber}. ${hit.name}`}</h4>
               {hit.is_mohcd_funded && (
               <div className="mohcd-funded">
-                <img src={images.mohcdSeal} alt="MOHCD seal" />
+                <img src={mohcdSeal} alt="MOHCD seal" />
```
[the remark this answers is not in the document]
davidagustin: ok

app/components/ui/Navigation/Navigation.jsx
```diff
@@ -46,23 +49,21 @@ class Navigation extends React.Component {
...
-    let logoLink;
-    if (isSFFamiliesSite()) {
-      logoLink = (
-        <a className={styles.navLogoSFFamilies} href={getSiteUrl()}>
-          <img src={images.logoSmall} alt={getSiteTitle()} />
+
+    const logoLink = title === 'SF Families'
+      ? (
+        <a className={configurationsStyles?.navLogoSFFamilies} href={siteUrl}>
+          <img src={logoSmall} alt={title} />
```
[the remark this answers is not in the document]
davidagustin: We can refactor this on line 61 actually. Will see what I can do.

app/components/ui/Navigation/Navigation.jsx
davidagustin: We can add a new label, will wait for other reviews too.

app/App.tsx
```diff
@@ -152,11 +152,12 @@ class App extends Component {
...
-          && config.INTERCOM_APP_ID
-          && <Intercom appID={config.INTERCOM_APP_ID} />}
-        {isSFFamiliesSite()
-          && <UserWay appID={config.SFFAMILIES_USERWAY_APP_ID} />}
+        {
+          title === 'SF Families'
+            ? <UserWay appID={config.SFFAMILIES_USERWAY_APP_ID} />
+            : config.INTERCOM_APP_ID
+          && <Intercom appID={config.INTERCOM_APP_ID} />
+        }
```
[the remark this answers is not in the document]
davidagustin: I see, I'd have to implement the white label interface with added properties specifically for SF Families.

app/App.tsx
```diff
@@ -166,10 +167,10 @@ class App extends Component {
         />
         <div id={pageWrapId}>
           <Navigation
-            showSearch={!isSFFamiliesSite()}
+            showSearch={title !== 'SF Families'}
```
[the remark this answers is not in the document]
davidagustin: I'm trying to keep the whitelabel interface as consistent as possible while keeping them unaware of each other. I'll see what solutions I can find.

app/App.tsx
```diff
@@ -166,10 +167,10 @@ class App extends Component {
...
             toggleHamburgerMenu={this.toggleHamburgerMenu}
           />
-          {!isSFFamiliesSite() && <Banner />}
+          {title !== 'SF Families' && <Banner />}
```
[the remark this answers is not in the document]
davidagustin: Ok

app/App.tsx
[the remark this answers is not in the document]
richardxia: UserWay is not a chat widget; it's an accessibility widget. We may potentially want to have both widgets enabled simultaneously, so I think these really should be completely orthogonal properties.

app/components/Texting/Texting.jsx
```diff
@@ -1,14 +1,16 @@
...
-import { images } from 'assets';
+import configurations from '../../utils/whitelabel';
 import styles from './Texting.module.scss';
 import * as dataService from '../../utils/DataService';
 import FormView from './components/FormView';
 import Loader from '../ui/Loader';
 import SentView from './components/SentView';
 import ErrorView from './components/ErrorView';
 
+const { appImages: { icon } } = configurations;
```
richardxia, line 12: Why are you importing the `icon` function from the whitelabel instead of `assets`? None of the icons appear to be different for the different whitelabeled sites, so it doesn't need to be part of the whitelabel site. This comment applies to many of the other files in this PR diff, since a lot of them only import the `icon` function and nothing from the actual whitelabel configuration.

app/App.tsx
```diff
@@ -8,7 +8,7 @@ import Navigation from './components/ui/Navigation/Navigation';
 // import CategoryPage from './find/FindPage';
 // import ResourcesTable from './search/ResourcesTable';
 import { round } from './utils/index';
-import { getSiteTitle, getSiteUrl, isSFFamiliesSite } from './utils/whitelabel';
+import configurations from './utils/whitelabel';
```
richardxia, line 11: Could you name default imports the same as the file being imported, e.g. name this `whitelabel`, not `configurations? That will reduce cognitive load by not introducing a new name. If you don't think the name `whitelabel` is good, then we may want to rename that file.

app/utils/whitelabel.ts
```diff
@@ -1,31 +1,89 @@
...
+type WhiteLabelSiteKey = 'default' | 'SFServiceGuide' | 'SFFamilies';
 
-const isSFServiceGuideSite = () => !isSFFamiliesSite()
-  && window.location.host.indexOf(config.MOHCD_DOMAIN) > -1;
+interface WhiteLabelSite {
+  title: string;
+  siteUrl: string;
+  appImages: {
+    background: string;
+     // For appImages, logoLarge and logoSmall are usually different between each white label
```
richardxia, line 19: I'm actually a little bit confused by this comment, since all of the whitelabel configurations use the same image for both `logoLarge` and `logoSmall`. What did you mean by "usually different"?

app/utils/whitelabel.ts
richardxia: You don't need to mark `WhiteLabelSiteKey` as `Readonly`, since they're strings, and strings are immutable in JavaScript.

It also might be cleaner to put this at the bottom of the file so that you don't have the `Partial<>` anymore, and so that you can just mark the whole thing as readonly. For example, if you did:

```ts
const SFFamilies = {
  title: 'SF Families',
  // etc
} as const;
const SFServiceGuide = { /* stuff goes here */ } as const;
const default = { /* stuff goes here */ } as const;

const configurations: Readonly<Record<WhiteLabelSiteKey, WhiteLabelSite>> = {
  SFFamilies,
  SFServiceGuide,
  default,
} as const;
```

app/utils/whitelabel.ts
```diff
@@ -1,31 +1,89 @@
...
-    return 'SF Families';
-  }
-  return 'AskDarcel';
+// Read only to force developer to modify configurations here, disallow changes at compile time
+const configurations: Partial<Record<Readonly<WhiteLabelSiteKey>, Readonly<WhiteLabelSite>>> = {};
+
+// Include a domain in config.js
+function determineWhiteLabelSite(): WhiteLabelSiteKey {
+  if (config.SFFAMILIES_DOMAIN === window.location.host) return 'SFFamilies';
+  if (config.MOHCD_DOMAIN === window.location.host) return 'SFServiceGuide';
```
richardxia, line 35: Nit: Could you swap the order of the operands across the `===` operator, i.e. put the `window.location.host` on the left-hand side of the `===`? Even though it's logically equivalent, it's much more common in programming and in English to put the thing that is more likely to change on the left-hand side.

cypress/support/index.ts
```diff
@@ -18,3 +18,5 @@ import './commands';
 
 // Alternatively you can use CommonJS syntax:
 // require('./commands')
+
+Cypress.on('uncaught:exception', () => false);
```
richardxia, line 22: What is this for? If this is silencing uncaught exceptions, that this seems like it's a bad thing to do, since we don't want these to slip past CI.

richardxia (commented):
Thanks for the updates! I left some additional comments, but I agree with everything that @jjfreund wrote about eliminating all logic checking for the current site outside of the whitelabel.ts file and replacing them with distinct properties.

cypress/support/index.ts
davidagustin: ok, yeah this was a workaround for a previous error already solved

app/utils/whitelabel.ts
davidagustin: ok

davidagustin: ok for the WhiteLabelSiteKey
default is a reserved word so Id have to change the name

richardxia: Maybe just call it `"AskDarcel"`, since that's what it actually is?

app/utils/whitelabel.ts
```diff
@@ -1,31 +1,114 @@
...
+function determineWhiteLabelSite(): WhiteLabelSiteKey {
+  if (window.location.host === config.SFFAMILIES_DOMAIN) return 'SFFamilies';
+  if (window.location.host === config.MOHCD_DOMAIN) return 'SFServiceGuide';
+  return 'defaultWhiteLabel';
+}
+
+const configKey = determineWhiteLabelSite();
+
+// Specify what is viewed in each white label
+const SFFamilies = <const> {
```
davidagustin, line 43: using "as const" at the end of these objects gives an error in which there needs to be a semi colon before "as const" for example the linter would request something like: {}; as const

.eslintrc.js
```diff
@@ -24,6 +24,8 @@ module.exports = {
       js: 'never', jsx: 'never', ts: 'never', tsx: 'never',
     }],
     'import/no-extraneous-dependencies': 'error',
+    'import/no-named-as-default': 0,
+    'import/no-named-as-default-member': 0,
```
davidagustin, line 28: These were needed to get passed linter errors for the whitelabel.ts files

app/components/ui/Navigation/Navigation.jsx
```diff
@@ -46,27 +57,23 @@ class Navigation extends React.Component {
...
-
     return (
-      <nav className={isSFFamiliesSite() ? styles.siteNavSFFamilies : styles.siteNav}>
+      <nav className={siteNavStyle}>
         <div className={styles.primaryRow}>
-          <div className={styles.navLeft}>
-            {logoLink}
-            {showSearch
+          {
+            /^https?:\/\//.test(logoLinkDestination)
```
davidagustin, line 64: I was being creative here by using regex, hopefully this makes sense but it is a string matching the first portion that must be http:// or https:// to return true.

app/components/ui/Navigation/Navigation.jsx
```diff
@@ -85,8 +92,9 @@ class Navigation extends React.Component {
                 </form>
               )
             }
-          </div>
-          {!isSFFamiliesSite()
+        </div>
+        <div>
+          {showMobileNav
```
davidagustin, line 97: mobile nav whiteLabel as requested here

app/components/ui/Navigation/Navigation.jsx
```diff
@@ -46,27 +57,23 @@ class Navigation extends React.Component {
...
-      logoLink = (
-        <Link className={styles.navLogo} to="/">
-          <img src={images.logoSmall} alt={getSiteTitle()} />
-        </Link>
-      );
-    }
-
     return (
-      <nav className={isSFFamiliesSite() ? styles.siteNavSFFamilies : styles.siteNav}>
+      <nav className={siteNavStyle}>
```
davidagustin, line 61: White label request fulfilled here

app/components/ui/Navigation/Navigation.jsx
```diff
@@ -46,27 +57,23 @@ class Navigation extends React.Component {
...
         <div className={styles.primaryRow}>
-          <div className={styles.navLeft}>
-            {logoLink}
-            {showSearch
+          {
+            /^https?:\/\//.test(logoLinkDestination)
+              ? (
+                <a className={navLogoStyle} href={siteUrl}>
```
davidagustin, line 66: White label for the css of the navLogo here

app/utils/whitelabel.ts
```diff
@@ -1,31 +1,114 @@
...
+  showSearch: true,
+  siteNavStyle: styles.navLogo,
+  siteUrl: 'https://sfserviceguide.org',
+  title: 'SF Service Guide',
+  userWay: false,
 };
 
-export {
-  isSFServiceGuideSite, isSFFamiliesSite, getSiteTitle, getSiteUrl,
+const defaultWhiteLabel = <const> {
```
davidagustin, line 83: default is a reserved word so I had to change it do defaultWhiteLabel

app/utils/whitelabel.ts
```diff
@@ -1,31 +1,114 @@
...
+  logoLinkDestination: string;
+  navLogoStyle: string;
+  showBanner: boolean;
+  showMobileNav: boolean;
+  showSearch: boolean;
+  siteNavStyle: string;
+  siteUrl: string;
+  title: string;
+  userWay: boolean;
+}
```
davidagustin, line 31: @fredciaramaglia Hopefully the variable names makes sense. Let me know if there are any questions.

app/App.tsx
```diff
@@ -152,11 +160,8 @@ class App extends Component {
           <meta property="og:image:width" content="1200" />
           <meta property="og:image:height" content="630" />
         </Helmet>
-        {!isSFFamiliesSite()
-          && config.INTERCOM_APP_ID
-          && <Intercom appID={config.INTERCOM_APP_ID} />}
-        {isSFFamiliesSite()
-          && <UserWay appID={config.SFFAMILIES_USERWAY_APP_ID} />}
+        {userWay && <UserWay appID={config.SFFAMILIES_USERWAY_APP_ID} />}
+        {intercom && config.INTERCOM_APP_ID && <Intercom appID={config.INTERCOM_APP_ID} />}
```
davidagustin, line 164: Orthogonalized userway and intercom, hopefully this is the way that is wanted.

@fredciaramaglia @jjfreund @richardxia

app/assets/index.js
```diff
@@ -1,48 +1,8 @@
...
-  };
-  /* eslint-enable global-require */
-}
+icons.keys().forEach(key => { iconPathMap[key.match(/ic-([^@]*)(?:@3x)?.(?:svg|png)/)[1]] = icons(key); });
 
-const images = appImages;
+const icon = name => iconPathMap[name.toLowerCase().replace(/(\s+|\/)/g, '-')];
 
-export { images };
+export { icon };
```
davidagustin, line 8: Icon is the only asset imported now because the appImages are moved to the whiteLabel file

davidagustin (commented):
Requesting QA on staging @fredciaramaglia @jjfreund @richardxia . Hopefully I got the whitelabel options correct. Will need a thorough check.

app/utils/whitelabel.ts
davidagustin: More in the actual file itself, didn't highlight them all.

app/utils/whitelabel.ts
davidagustin: @richardxia

jjfreund (approved):
lgtm -- will merge and test. Further cleanup in subsequent PRs if need be.

lgarofalo:
Looks like a good refactor, I am going to merge and do verification/testing in staging, we won't push a prod release until we have gone through this well.  We will also be adding the linksf config on top of here before prod.

app/App.tsx
richardxia: Yep, the way they are now orthogonal looks great!

But I'm also now seeing something a bit weird about these lines, which is that `SFFAMILIES` has still leaked into the line via the `config. SFFAMILIES_USERWAY_APP_ID` configuration setting. This isn't good, because it means that configuration option is actually still tied to the whitelabeled site. At the configuration option level, I think we need to keep the `SFFAMILIES_` prefix because the ID is tied to them and possibly to some subscription they've paid for. However, I'm now thinking that we could maybe handle this within the whitelabel file itself by combining it with the existing `userWay` property.

What do you think of the following idea: In whitelabel.ts, replace the `userWay` property with `userWayAppID`, and have it be `null` either when the whitelabeled site should not have UserWay or when the configuration option is missing? This allows it to play both the role of only enabling it based on which whitelabel site as well as providing the actual value. I think we could do the same thing with Intercom, in which case you could then eliminate the extra check on `config.INTERCOM_APP_ID` here.

app/components/ui/Navigation/Navigation.jsx
richardxia: I think this is a reasonable thing to do, but it might be more readable if you extracted it out into its own function and gave it a name that was more descriptive, like `isExternalURL`.

app/components/ui/Navigation/Navigation.jsx
```diff
@@ -46,27 +57,23 @@ class Navigation extends React.Component {
...
+                  <img src={logoSmall} alt={title} />
+                </a>
+              )
+              : (
+                <Link className={navLogoStyle} to="/">
```
richardxia, line 71: For consistency, you should also set the `to` prop here to `logoLinkDestination`, which would better mirror what's going on in the other case.

app/utils/whitelabel.ts
```diff
@@ -1,31 +1,121 @@
...
+  if (window.location.host === config.MOHCD_DOMAIN) return 'SFServiceGuide';
+  return 'defaultWhiteLabel';
+}
+
+const configKey = determineWhiteLabelSite();
+
+/*
+  Specify what is viewed in each white label.
+  A '/' (which is a forward-slash) as a value for logoLinkDestination
+  denotes that the link is internal to the application.
```
richardxia, line 45: This comment would be better to place on the interface definition, right above the actual `logoLinkDestination`, since that interface should be the one place where we define what the interface is.

app/utils/whitelabel.ts
```diff
@@ -1,31 +1,121 @@
...
+  siteNavStyle: styles.navLogo,
+  siteUrl: 'https://askdarcel.org',
+  title: 'AskDarcel',
+  userWay: false,
+} as const;
+
+/*
+  whiteLabel made Readonly to force developer to modify whiteLabel object in this file.
+  Disallow changes at compile time.
+*/
```
richardxia, line 110: I'm not sure if this comment provides enough value to justify its existence. It mostly just describes what `Readonly` means in TypeScript without anything that is novel or specific to this codebase. I would argue that the primary purpose of `Readonly` is to do exactly what you've written here, so repeating that here in the form of a comment doesn't add much information, and possibly worse, it draws extra attention for something that is ordinary.

Do you think you could remove this comment?

app/utils/whitelabel.ts
richardxia: Similar to my other PR comment, I don't think this code comment is really saying anything that isn't already obvious from the code just below it. Could you remove it?

app/utils/whitelabel.ts
```diff
@@ -1,31 +1,121 @@
...
+const whiteLabel: Readonly<Record<WhiteLabelSiteKey, WhiteLabelSite>> = {
+  SFFamilies,
+  SFServiceGuide,
+  defaultWhiteLabel,
+} as const;
+
+// Disallow changes at run time
+Object.freeze(whiteLabel);
+Object.freeze(whiteLabel[configKey]?.appImages);
```
richardxia, line 119: You don't need the `?` anymore, since this should never be null or undefined.

Also, even previously, when `whiteLabel` was a `Partial<Record>`, you probably did not want this to use `?`, since that operator should only be used in the case where you want to short-circuit and return null/undefined early. Here, it should be illegal to encounter a null or undefined value, so you should not write code that handles that case, and instead it would be better assert that the value is not null.

Note that this comment doesn't apply to the current version of the code, since `whiteLabel` is no longer a `Partial<Record>`, but in the future if you have code where it should be impossible to have a null value but TypeScript doesn't know about it, you should assert the correct type, not work around a type that should not be possible.

richardxia (commented):
Sorry that I'm late to the review, but this LGTM and I don't see any blockers here.

I did leave some comments on a few things that I think would be good to fix sooner rather than later, so if you have time, I think it'd be great if you could fix them in a followup PR, @davidagustin. If you're busy, let me know, and I can submit a PR to fix it up as well.

app/components/ui/Navigation/Navigation.jsx
davidagustin: Acknowledged. Good Ideas, I agree with all of them. Implementing.

PR #1072, opened 2021-08-25
Convert RecurringSchedule to TypeScript.

richardxia:
Inspired by one of the discussions in https://github.com/ShelterTechSF/askdarcel-web/pull/1068, I've converted the RecurringSchedule set of classes to TypeScript.

Fortunately, this was relatively straightforward, since I had previously added JSDoc comments where I annotated the types of the parameters and return values, so I could simply just translate them to TypeScript.

I also scrapped the directory with three files and just merged it into one file, since it didn't seem to be worth it to have separate files.

I didn't change any of the method signatures, although in retrospect, my choice of using objects as arguments to all the constructors in order to emulate keyword arguments was probably not worth it, and it makes the TypeScript types more verbose.


richardxia:
Looks like the mocha tests started failing when I converted that file to TypeScript, since our mocha didn't understand how to read TypeScript files. I installed `ts-node` and configured mocha to use it, which gets the tests passing again.

lexholden (approved):
I had started doing this in my WIP PR for the org page, but your approach works better so LGTM.

Only major thing I chose to do slightly differently, was moving this to `models`, which feels more appropriate alongside Organization and Service and ChangeRequest etc, I had since split Schedule into its own file too, since it has a lot of utility functions alongside the types, all of which make sense to have their own unit tests eventually.

richardxia:
> Only major thing I chose to do slightly differently, was moving this to `models`, which feels more appropriate alongside Organization and Service and ChangeRequest etc, I had since split Schedule into its own file too, since it has a lot of utility functions alongside the types, all of which make sense to have their own unit tests eventually.

Ah yeah, that sounds like a good idea. I can make that change before merging this in. Thanks for the review!

PR #1074, opened 2021-09-13
Map white label subdomains to a staging name

lgarofalo:
Added one fix, the env configed domain is just the subdomain key, "sffamilies" etc for sffamilies.sfserviceguide.org

Also support same mapping for sffamilies-staging.sfserviceguide.org etc.

### DIFF
--- a/app/utils/whitelabel.ts
+++ b/app/utils/whitelabel.ts
@@ -32,8 +32,9 @@ interface WhiteLabelSite {
 
 // Include a domain in config.js
 function determineWhiteLabelSite(): WhiteLabelSiteKey {
-  if (window.location.host.split('.')[0] === String(config.SFFAMILIES_DOMAIN) || window.location.host.split('.')[0] === `${String(config.SFFAMILIES_DOMAIN)}-staging`) return 'SFFamilies';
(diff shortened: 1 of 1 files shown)


app/utils/whitelabel.ts
```diff
@@ -32,8 +32,8 @@ interface WhiteLabelSite {
 
 // Include a domain in config.js
 function determineWhiteLabelSite(): WhiteLabelSiteKey {
-  if (window.location.host === config.SFFAMILIES_DOMAIN) return 'SFFamilies';
-  if (window.location.host === config.MOHCD_DOMAIN) return 'SFServiceGuide';
+  if (window.location.host.split('.')[0] === String(config.SFFAMILIES_DOMAIN) || window.location.host.split('.')[0] === `${String(config.SFFAMILIES_DOMAIN)}-staging`) return 'SFFamilies';
+  if (window.location.host.split('.')[0] === String(config.MOHCD_DOMAIN) || window.location.host.split('.')[0] === `${String(config.MOHCD_DOMAIN)}-staging`) return 'SFServiceGuide';
```
richardxia, line 36: This makes the line very long and hence more difficult to read on GitHub or even in a text editor. Could you factor the check into a helper function?

richardxia (approved):
LGTM

app/utils/whitelabel.ts
lgarofalo: Yea it's not pretty, I made it a little shorter for now.

lgarofalo:
I think some of this was not correct, I'm stepping back to review.

PR #1075, opened 2021-09-13
Upgrade typescript packages

davidagustin:


davidagustin:
The intent of this PR was to make the developer environment less of a hassle for developers because of linter issues with typescript. The root of the problem was associated with babel. This PR is therefore closed.

PR #1077, opened 2021-09-15
Fix silent ESLint errors by upgrading babel-eslint

richardxia:
## Problem summary

We're getting silent errors with ESLint right now, due to our babel-eslint being too old to parse TypeScript files, which is an issue when we have plain JavaScript files import TypeScript files. An example of the silent errors can be seen in this [passing CI run](https://github.com/ShelterTechSF/askdarcel-web/runs/3553457316), under the "Run npm run lint" section:

```
Run npm run lint

> ShelterTech@1.0.0 lint /home/runner/work/askdarcel-web/askdarcel-web
> eslint --ext=.js,.jsx,.ts,.tsx .


Error while parsing /home/runner/work/askdarcel-web/askdarcel-web/app/utils/whitelabel.ts
Line 65, column 2: Line 65: Missing semicolon.

  63 |   title: 'SF Families',
  64 |   userWay: true,
> 65 | } as const;
     |  ^
  66 |
  67 | const SFServiceGuide: WhiteLabelSite = {
  68 |   appImages: {
`parseForESLint` from parser `/home/runner/work/askdarcel-web/askdarcel-web/node_modules/babel-eslint/lib/index.js` is invalid and will just be ignored
```

This does not cause the ESLint process to fail, so it still exits with non-zero status when no lint errors are detected. This still isn't good, since this could hide other issues due to the imports not being properly resolved.

## Background on our ESLint configuration
Our setup is a bit complicated, but we use two separate JavaScript parsers for ESLint: `@typescript-eslint/parser` for the TypeScript files and `babel-eslint` for the normal JavaScript files. The issue is that sometimes our JavaScript files import TypeScript files, and then `babel-eslint` will attempt to parse the file being imported, but fails due to not understanding newer TypeScript syntax.

The [babel-eslint](https://github.com/babel/babel-eslint) package was deprecated two years ago, and it was moved into [@babel/eslint-parser](https://www.npmjs.com/package/@babel/eslint-parser) to be under the `@babel` namespace.

## What this PR changes
This PR removes `babel-eslint` and switches to `@babel/eslint-parser`, since that's the new name of the package, which has actually been receiving updates over the past two years.

Since this upgrades the package by a couple of years, we picked up one more lint rule that we didn't have before which yells at you if you have unnecessary curly braces in JSX/TSX code. We only had one instance of that, so I fixed in this PR as well.

This should make https://github.com/ShelterTechSF/askdarcel-web/pull/1075 unnecessary, as the root cause of the silent ESLint issues was not the TypeScript parser but rather the Babel parser for TypeScript.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -64,7 +64,7 @@ module.exports = {
     {
       // Non-TypeScript, JavaScript files
       files: ['*.js', '*.jsx'],
-      parser: 'babel-eslint',
+      parser: '@babel/eslint-parser',
       rules: {
         'react/sort-comp': 'off',
         // Disable TypeScript-specific rules on regular JavaScript files.
--- a/app/components/listing/feedback/FeedbackModal.jsx
+++ b/app/components/listing/feedback/FeedbackModal.jsx
@@ -110,7 +110,7 @@ const FeedbackModal = ({ service, resource, closeModal }) => {
         <SubmitMessage closeModal={closeModal} />
       ) : (
         <div className={styles.stepsContainer}>
-          {<VoteButtons vote={vote} onVoteChange={handleVoteChange} />}
+          <VoteButtons vote={vote} onVoteChange={handleVoteChange} />
           {STEPS[step]}
           <NavigationButtons
             step={step}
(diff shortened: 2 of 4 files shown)


davidagustin (approved):
Approved and got my other PR closed.

PR #1084, opened 2021-10-13
Migrating the main app and core UI components to typescript, removing redux

lexholden:
- Migrate the core App component to TS, and simplfied it a little, moved helpers elsewhere, moved routes inside
- Entirely removed redux from few places that used it, including the search page and the edit page
  - Replaced redux usage with appContext to store LatLong and useAppContext hook
  - Remove Edit page reliance on redux to set Popup message, instead pass as prop directly
- Simplified some generic UI components in components/ui and moved to TS


app/components/listing/OrganizationCard.tsx
```diff
@@ -0,0 +1,25 @@
+import React, { useMemo } from 'react';
+import { Link } from 'react-router-dom';
+import StreetViewImage from 'components/maps/StreetViewImage';
+import { Organization } from '../../models';
+
+export const OrganizationCard = ({ org }: { org: Organization }) => {
+  const address = org.addresses[0];
```
richardxia, line 7: Wait, is this the reason this has been broken the whole time? That the old code had `org.address`, which is not a real field? I can see in production that every time I try to hover over the link, it causes a Sentry error to fire with "Cannot read properties of undefined (reading 'address_1')".

Thank you for finally fixing this, now that we actually have type checking.

app/init.tsx
```diff
@@ -21,27 +18,13 @@ if (process.env.NODE_ENV === 'production') {
...
-  ReactGA.pageview(loc.pathname);
-});
-
 const rootElement = document.getElementById('root')!;
 ReactModal.setAppElement(rootElement);
 
 ReactDOM.render((
-  <Provider store={store} key="provider">
-    <ConnectedRouter history={history}>
-      <ScrollToTop>
```
richardxia, line 40: Is `<ScrollToTop>` not necessary anymore?

app/components/maps/MapOfLocations.jsx
```diff
@@ -61,7 +61,7 @@ class MapOfLocations extends React.Component {
                 <AccordionItem
                   key={loc.address.id}
                   title={loc.address.address_1}
-                  headerRenderer={title => (
+                  headerRenderer={(
```
richardxia, line 64: Since you got rid of the kind of weird title argument to the `headerRenderer` render prop, maybe you could also just get rid of the `title` prop at all? I see no reason to have two props do the same thing, except one is allowed to be an arbitrary `JSX.Element` while the other is restricted to being a string.

app/components/ui/Datatable.tsx
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+
+interface DatatableProps<T = any> {
+  rows: T[];
+  rowRenderer: (row: T) => JSX.Element;
+}
+
+export const Datatable = ({ rows, rowRenderer }: DatatableProps) => (
```
richardxia, line 8: You should also define the type parameter on `Datatable` as well so that type inference at the call site can propagate into the type parameter.

Type parameters on arrow functions are pretty funky in JSX, since there's ambiguity in the parser as to whether to interpret it as an opening HTML tag or a type parameter, but you can work around that by using the `<T extends unknown>` [trick](https://stackoverflow.com/a/45576880), or you can just use a normal `function`-keyword function.

```suggestion
export const Datatable = <T extends unknown>({ rows, rowRenderer }: DatatableProps<T>) => (
```

By defining `Datatable` this way, TypeScript can warn you if you pass in a `rowRenderer` that doesn't actually match the `rows`. This is actually super neat, because you don't even need to explicitly tell TypeScript the concrete type at the call site, since it's just completely inferred via the type of `rows`.

Also, I'd get rid of the default value of the type parameter on `DatatableProps`, just to force users of the type to have to explicitly pass it in (or better yet, explicitly propagate the type from elsewhere).

app/components/ui/Navigation.tsx
```diff
@@ -0,0 +1,128 @@
...
+  siteNavStyle,
+  siteUrl,
+  title,
+} = whiteLabel;
+
+export const Navigation = ({ showSearch, toggleHamburgerMenu }: {
+  showSearch: boolean;
+  toggleHamburgerMenu: () => void;
+}) => {
+  const [query, setQuery] = useState('');
```
richardxia, line 23: I think we may want to do a bit of extra QA in staging for this, since I remember this being very complicated, since there's a very unintuitive sequence of how this is initialized, updated, and kept in sync across different components, especially due to part of this living in the browser history and URL location.

But from what I'm reading in your code changes, I think this looks about right, other than it looking like we previously didn't actually initialize `this.state.query`, meaning that it was probably `undefined`. I definitely prefer your way here, setting it to an explicit value, but I'd just want to be careful about it.

app/config.ts
```diff
@@ -1,22 +1,26 @@
...
-/* global CONFIG */
+declare global {
+  const CONFIG: any;
+}
+
 const config = {
   ALGOLIA_APPLICATION_ID: CONFIG.ALGOLIA_APPLICATION_ID,
   ALGOLIA_INDEX_PREFIX: CONFIG.ALGOLIA_INDEX_PREFIX,
   ALGOLIA_READ_ONLY_API_KEY: CONFIG.ALGOLIA_READ_ONLY_API_KEY,
+  GOOGLE_ANALYTICS_ID: (process.env.NODE_ENV === 'production' || window.location.host === 'www.askdarcel.org') ? 'UA-116318550-1' : 'UA-116318550-2', // TODO This should probably be in whitelabel
```
richardxia, line 10: Good call on moving this here, out of where it used to be hardcoded.

app/pages/SearchResultsPage.tsx
```diff
@@ -0,0 +1,55 @@
...
+        indexName={`${config.ALGOLIA_INDEX_PREFIX}_services_search`}
+        searchState={searchState}
+        onSearchStateChange={(nextSearchState: any) => {
+          const THRESHOLD = 700;
+          const newPush = Date.now();
+          setSearchState(nextSearchState);
+          setLastPush(newPush);
+          const newParams = new URLSearchParams(nextSearchState);
+
+          if (lastPush && newPush - lastPush <= THRESHOLD) {
```
richardxia, line 35: I think you don't need the `lastPush &&` part now, since you're always initializing it to some value.

app/init.tsx
lexholden: All it did was listen to route changes and perform window.scrollTo(0,0). Felt overkill to be a whole component, I just moved that behavior to a one liner in a useEffect inside <App />

I also added a todo, because I'm not sure we do need it anymore, but that's out of scope for this PR

app/components/listing/ListingTitleLink.tsx
```diff
@@ -0,0 +1,25 @@
...
+import { Tooltip } from 'react-tippy';
+import 'react-tippy/dist/tippy.css';
+import { OrganizationCard } from './OrganizationCard';
+import { ServiceCard } from './ServiceCard';
+import { Service, Organization } from '../../models';
+
+export const ListingTitleLink = ({ listing, type }: { listing: Service; type: 'service' } | { listing: Organization; type: 'org' }) => {
+  const isService = type === 'service';
+  const to = isService ? `/services/${listing.id}` : `/organizations/${listing.id}`;
+  const summaryCard = type === 'service' ? <ServiceCard service={listing as Service} /> : <OrganizationCard org={listing as Organization} />;
```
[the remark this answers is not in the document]
richardxia: Oh, got it. Yeah, I would support disabling that rule, or at least disabling it in this one special case, but I agree with treating it as out of scope for this PR.

PR #1086, opened 2021-10-21
Consolidated Map implementations to use the same library and components, in typescript

lexholden:
We used two different implementations for the Google Map rendered on the Search Results page and the Org page - here I consolidated them to both use the same google-map-react component (instead of using window.google.maps to manually render in the dom) and moved them to both TS.


app/components/listing/StreetViewImage.tsx
```diff
@@ -1,46 +1,28 @@
...
-import PropTypes from 'prop-types';
+import React, { useMemo } from 'react';
 import config from '../../config';
+import { Address } from '../../models';
 
-class StreetViewImage extends React.Component {
-  buildImageUrl() {
-    const { address } = this.props;
-
+export const StreetViewImage = ({ address, size = null }: { address: Address; size?: any }) => {
```
richardxia, line 5: Seems like `size` is supposed to be a string, based on looking at the old prop types and the fact that this is getting passed into the inline `style` prop.

app/components/search/SearchMap.tsx
```diff
@@ -0,0 +1,65 @@
...
+            return markers;
+          }, [] as ReactElement[]) }
+        </GoogleMap>
+      </div>
+    </div>
+  );
+};
+
+// eslint-disable-next-line react/no-unused-prop-types
+const SearchHitMarker = ({ hit, tag }: { lat: any; lng: any; hit: SearchHit; tag: string }) => (
```
richardxia, line 53: ~~I'm a little bit confused about this ESLint error that you're silencing. It looks like it's complaining about the fact that `lat` and `lng` are unused in this component. Isn't that a problem, since I don't understand how the custom markers get the correct coordinates if you're not passing them through.~~

I figured out that the library we're using just wants to see `lat` and `lng` props on the child components of the `<GoogleMap>` component. Maybe add a quick comment here describing that, since that would explain why we're silencing the ESLint error rather than just getting rid of the unused props?

app/models/Meta.ts
```diff
@@ -14,6 +17,14 @@ export interface Address {
   longitude: string;
 }
 
+export interface LocationDetails {
+  id: number;
+  address: Address;
+  name: string;
+  recurringSchedule: RecurringSchedule;
+  inherited: boolean | Schedule;
```
richardxia, line 25: Any specific reason you chose to make this a `boolean | Schedule` instead of a `Schedule | null`? One potential issue with `boolean | Schedule` is that it allows you to pass in `true`, which doesn't really make much sense as a value. If you're doing this for compatibility with how the code is written today, is it possible to at least use the type `false | Schedule`? The `false` part is a literal type, which means it'll only accept `false` and not `true`.

richardxia (approved):
Apologies for the months late review, but thanks for tackling this and cleaning up a lot of the messy old Maps code!

app/components/listing/StreetViewImage.tsx
schroerbrian: Richard makes a good point here. Null could be passed to the img width/height props. I'm not sure if it would matter, but perhaps a default CSS value would be better, like "auto" or something? Alternatively, if size is null, then the code could just not add it as a style attr to the img element

schroerbrian (approved):
Looking good. I think Richard made some good points that are worth considering before this is merged

app/models/SearchHits.ts
```diff
@@ -0,0 +1,40 @@
...
+ */
+export const transformHits = (hits: SearchHit[]) => hits.map(hit => {
+  switch (hit.type) {
+    case 'resource':
+      return {
+        ...hit,
+        recurringSchedule: hit.schedule?.length ? parseAlgoliaSchedule(hit.schedule) : null,
+      };
+    case 'service':
+      // eslint-disable-next-line no-case-declarations
```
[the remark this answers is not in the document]
lexholden: Hmm interesting, did not know this, thanks. Fixed.

app/models/Meta.ts
lexholden: Agreed. Actually this key used to be used, but no longer is, so I just removed it altogether.

PR #1087, opened 2021-11-09
1981 Add Link SF specific about page markup and conditional to displa…

schroerbrian:
…y based upon whitelabel title.


jjfreund (approved):
Discussed with Brian. Seems to me that we could avoid the conditional logic by having a property linkURL or in the config object and pointing to a different page but this should work for now!

PR #1088, opened 2021-12-01
2005 move about page JSX markup to individual files. Fix Link SF abou…

schroerbrian:
…t page dead link

Fixes Link SF TL Tech Lab dead links. (See screenshot below with updated href).

Also moves Link SF and default SF Service Guide about page JSX to their own files to clean up About page file.

Screen Shot 2021-11-30 at 5 41 40 PM
s.


jjfreund (approved):
I still think it would be preferable to somehow not have a reference to linksf in anywhere but the config object (perhaps instead referencing separate pages like discussed as config variables) but this should work for now.

PR #1090, opened 2022-01-06
Fixed techlab links

jjfreund:


cypress.json
```diff
@@ -1,3 +1,4 @@
 {
-  "baseUrl": "http://localhost:8080"
+  "baseUrl": "http://localhost:8080",
+  "defaultCommandTimeout": 10000
```
richardxia, line 3: Yeah, I don't think it's a timeout issue. The error message seems to indicate that the test is finding the text "Sunday", but it's also finding a bunch of other unrelated text. Confusingly, it looks like `contains.text` actually means "exactly matches" in Cypress/chai: https://github.com/cypress-io/cypress/issues/1969

You might be able to change the test to use a regex instead of an exact string match, like this: https://docs.cypress.io/guides/references/assertions#Text-Content

jjfreund: How would both of these statements ever work then (from the flaky test)?
            .should('contain.text', 'Sunday')
            .should('contain.text', '8:00 AM - 10:00 AM');

richardxia (approved):

PR #1091, opened 2022-01-09
Update Feedback

Akhtam:


app/components/listing/feedback/FeedbackModal.jsx
```diff
@@ -102,9 +100,6 @@ const FeedbackModal = ({ service, resource, closeModal }) => {
         <img src={icon('feedback-blue-header')} alt="feedback" />
         <span>Share your Feedback</span>
       </div>
-      <div className={styles.feedbackSubheader}>
```
Akhtam, line 105: As of May of 2021 We didn't have a way of reaching back out so we had to remove sub-header message "Team will usually replies within a day"

app/components/listing/feedback/FeedbackModal.jsx
```diff
@@ -39,19 +40,22 @@ const FeedbackModal = ({ service, resource, closeModal }) => {
     setReview(e.target.value);
   };
 
+  const isDownVote = vote === DOWNVOTE;
```
Akhtam, line 43: Since this logic been used in multiple places decided to assign it to variable

app/components/listing/feedback/FeedbackModal.jsx
```diff
@@ -39,19 +40,22 @@ const FeedbackModal = ({ service, resource, closeModal }) => {
...
 
   const handleSubmit = () => {
     setIsSubmitted('submitting');
-    const tags = tagOptions
-      .filter(({ selected }) => selected)
-      .map(({ tag }) => tag);
+
+    const tags = isDownVote
+      ? tagOptions.filter(({ selected }) => selected).map(({ tag }) => tag)
+      : [];
```
Akhtam, line 58: We wanna keep `review` and `tags` state consistent if user toggles between upvote and downvote but need to make sure we send empty tags before submit since tags associated with downvote only.

app/components/listing/feedback/FeedbackSteps.jsx
```diff
@@ -52,17 +52,15 @@ export const FeedbackTags = ({ tagOptions, onSelectTag }) => (
...
     <div className={styles.stepsPrompt}>
       Please provide your feedback below:
     </div>
     <textarea
       className={styles.feedbackTextarea}
       type="text"
-      placeholder={`Type your feedback here ${
-        !isReviewRequired ? '(optional)' : ''
-      }`}
+      placeholder="Type your feedback here(optional)"
```
richardxia, line 63: It'll look tidier and if you put a space between the word "here" and the parenthetical "(optional)":

```suggestion
      placeholder="Type your feedback here (optional)"
```

richardxia (approved):
Thanks for following up, this LGTM!

I think we're having a lot of issues with flaky CI runs, particularly the Cypress end-to-end tests. @jjfreund, @lexholden, @schroerbrian, we may need to prioritize getting CI to be more stable.

PR #1094, opened 2022-02-06
Remove deprecated node-sass in favor of sass, which works on M1 Macs

lexholden:
I get errors running/installing on my new Mac, and https://www.npmjs.com/package/node-sass is deprecated anyway, so replacing it.


richardxia (approved):
LGTM, but it looks like the Cypress tests are failing, and I don't think it's because of that flaky failure we've been seeing. It looks like every single test is failing on `502: Bad Gateway`.

PR #1095, opened 2022-02-06
Experiment replacing npm with yarn v3 for faster builds and modern tooling

lexholden:
This is an experiment, canvassing people's thoughts. Ultimately npm feels a little dated, and we're using a version of node that hasn't been supported for years.

- Replace usage of npm with yarn across the board
- Upgrade to yarn3, which recommends bundling zipped dependencies with the repo for faster builds, removes dependency on external package repos (slightly speeds up build times)
- Add dependencies that were actually missing but most people have installed globally
- Remove caching step from CI builds, no longer needed


schroerbrian:
Thanks for getting this going! I'm fairly agnostic about Yarn vs NPM, at least until I become more familiar with our code base and env. I'm okay with switching if that's the general feeling on our team. And it looks like Yarn has way more of a fanbase on Github. Would all of those binaries need to be added to the repo if we switched to Yarn? (And whatever we may choose, I think upgrading Node to the most recent LTS is a great idea.)

richardxia:
I have a _slight_ preference for npm over yarn, mostly because most distributions of Node.js also come with npm, but yarn usually has to be installed separately. I haven't used yarn much recently, especially since npm started catching up in terms of performance several years ago, so I'm less familiar with its newer features. In case you haven't used npm much recently, they have added a lot of features over time, I believe in response to features that Yarn had, so I believe the difference between the two is no so great these days.

However, I'm really not a fan of committing the dependencies to Git, since I feel like it significantly bloats up the repo without offering enough benefit. Because you've pushed up this branch with all the committed files, this means that now all git clones of askdarcel-web are a lot larger than they used to be. I did a few experiments to measure exactly how much extra space this is:

- First, from a fresh clone and checking out your branch, I ran `du -sh .yarn/cache`, which shows `131M`
- Second, doing a fresh clone of askdarcel-web and running `du -sh .git/` shows a total size of `140M`. Doing it on just `.git/` allows me to measure only the internal Git objects and not the checked out files
- Finally, I did another fresh clone, but stayed on `master` this time. I manually deleted the remote reference to this branch with `git branch -r -D origin/feature-yarn-3`, then forced a Git garbage collection with `git gc --prune=now`, and then measured the size of the `.git/` directory and got `16M`. This looks about right, even if it's larger than the delta above, since Git can do a little bit of compression but it won't be able to do much with already compressed (zipped) files. This shows that the repo used to be only about 16 MB before you pushed this branch up.

Based on my measurements, it looks like adding the Yarn cache to this repo increases the repo size by almost a factor of 10, and this is a project with 6 years of history, so this is pretty significant. Presumably future bumps to JavaScript packages will also increase the repo size, since each updated package will also be saved.

I have strong feelings _against_ committing the Yarn cache because if this merges into the master branch, it's stuck forever in the Git history, since we cannot delete files from history even if we delete them on the latest commits. The only way to delete files in the history is to rewrite history, which will change all the commit hashes and likely mess up everyone else's clones of the repo.

I think this feature you are referring to is called [Zero-Installs](https://yarnpkg.com/features/zero-installs) by the Yarn docs, but it looks like they don't even recommend it to all users, as shown in the warning box at the top:

> **Important**: Zero-install is an optional philosophy. It has some drawbacks, and while we believe this workflow to be a defining feature for professional-grade projects we don't have any plans to ignore or deprecate the typical yarn install workflow in any way, now or in the future.

While I can understand this being useful in large corporations, especially those that have internal mirrors or all third party packages or those that commit all third party source code directly in a monorepo, I don't think that kind of workflow makes sense for us. I also don't understand exactly how this works for dependencies that rely on non-JavaScript, dynamically linked libraries on the host system, such as libsass, since we don't have a uniform development environment like a larger company may be able to maintain and enforce.


---

> And it looks like Yarn has way more of a fanbase on Github.

I think it's hard to tell, since both organizations appeared to have created new repos for their main tools, so the number of forks or stars may not be easy to measure. NPM used to be at https://github.com/npm/npm, but they split it up into different repos, with https://github.com/npm/cli being the successor to the CLI tool. Yarn used to be at https://github.com/yarnpkg/yarn, but https://github.com/yarnpkg/berry is the modern repo.

FWIW, [GitHub acquired NPM a couple years ago](https://github.blog/2020-03-16-npm-is-joining-github/), so if anything, GitHub should have pretty good integration with NPM.

schroerbrian:
I think we may want to shelve this for the time being, or at least delete this remote branch. I just cloned askdarcel-web repo on my new comp and learned first-hand how large our repo is with the Yarn caching included 😄

richardxia:
> I think we may want to shelve this for the time being, or at least delete this remote branch. I just cloned askdarcel-web repo on my new comp and learned first-hand how large our repo is with the Yarn caching included 😄

Yeah, I think we can just close this for now. We can always restore the branch after deleting it, so there's no harm in deleting the branch.

lexholden:
Sure no problem, was a quick experiment I've seen reduce build times significantly in other larger projects, but impact on a smaller repo is negligible (5-10 seconds) so seems not worth the drawbacks.

Closing.

richardxia:
Thanks!

PR #1097, opened 2022-02-09
Reload service page in Cypress test rather than fetching to bypass po…

schroerbrian:
This adds an [intercept/wait](https://docs.cypress.io/api/commands/intercept) to some client side API requests during our e2e tests. The intercept/wait ensures that the request is returned before the subsequent tests execute. My hunch is that there was a race condition where sometimes the request would return prior to the test execution and sometimes after, causing the flakiness. 

I've also replaced some of the `cy.visit()` methods with `cy.reload(forceReload)` to ensure that an uncached copy of the page is loaded prior to the tests running. Lastly, I've added a step in the E2E job to upload a screenshot and video of the tests when they fail, per Richard's very good suggestion.

I've run the E2E job probably a couple dozen times and haven't gotten a failure for the main flakey test that we were seeing. However, there is another spec that has been failing occasionally on the organization page e2e tests—Cypress times out trying to find an element (see: https://github.com/ShelterTechSF/askdarcel-web/runs/4755264042?check_suite_focus=true). I added an intercept/wait there to make sure the resource API request returns first, but sometimes the wait times out too, because the request never occurs. I think this could actually be a problem with our code rather than the test; in the videos of this failed test, the API request is never actually logged. Thus it may not happen at all in rare circumstances. I'll look into this.

For now, I hope this PR makes the process a little more smooth for everyone. Suggestions/requests much appreciated, especially as I become re-familiarized with our conventions. Thanks!

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -66,7 +66,14 @@ jobs:
       - run: yarn cypress run
       - if: always()
         run: docker-compose -f docker-compose.test.yml logs -t # Output all docker logs from API and frontend to help debugging after the fact
-
+      - uses: actions/upload-artifact@v2
+        if: failure()
+        with:
+          name: E2E screenshot(s) and video(s)
+          path: |
+            /home/runner/work/askdarcel-web/askdarcel-web/cypress/videos/
+            /home/runner/work/askdarcel-web/askdarcel-web/cypress/screenshots/
+          retention-days: 3
 
   publish-latest:
     runs-on: ubuntu-latest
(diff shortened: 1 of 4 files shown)


lexholden (approved):

richardxia (commented):
I'm late to the party, but thanks for digging into this and getting the Cypress tests to be more reliable!

I did a little bit of searching around for for this class of problem, and I found this open ticket on the Cypress repo: https://github.com/cypress-io/cypress/issues/1773. There's no official solution to the problem, and I think the closest thing is the intercept/wait thing that you implemented here, but I did see someone post a link to https://www.npmjs.com/package/cypress-network-idle, which is a plugin that adds a new method for waiting until the network to be idle. If we get too tired of intercepting individual API calls to wait on, we could take a look at that plugin.

schroerbrian:
Awesome, this is super helpful info. I will look into both of these links. Thanks!

PR #1098, opened 2022-02-12
Migrated ActionSidebar components to TS

lexholden:
There is a lot of duplicated logic between the Mobile action bar and the regular web action sidebar. This cleans up and refactors the basic logic and fixes some bugs

- Fix CSS bug that was making the icons too big
- Refactored ActionSidebar and MobileActionBar into one file, consolidating shared logic, and moving to typescript FCs
- Move Feedback modal into listing pages and handle event from feedback action there

image


app/models/Organization.ts
```diff
@@ -74,3 +83,74 @@ export const getResourceLocations = (org: Organization): LocationDetails[] => {
...
+// };
+
+export const getResourceActions = (
+  resource: Organization,
+  service?: Service,
+  filterActions?: string[],
+) => {
+  const phoneNumber = _.get(resource, 'phones[0].number');
+  const latitude = _.get(resource, 'addresses[0].latitude');
+  const longitude = _.get(resource, 'addresses[0].longitude');
```
schroerbrian, line 109: I think we could replace the `.get` method with the ES6 optional chaining operator (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) to avoid having to import lodash in this file

schroerbrian (commented):
Looks great. I have a few small comments/questions but nothing major

app/models/Organization.ts
```diff
@@ -74,3 +83,74 @@ export const getResourceLocations = (org: Organization): LocationDetails[] => {
...
+  if (latitude && longitude) {
+    actions.push({
+      name: 'Directions',
+      icon: 'directions',
+      link: `http://google.com/maps/dir/?api=1&destination=${latitude},${longitude}`,
+    });
+  }
+
+  return filterActions
+    ? actions.filter(a => filterActions.indexOf(a.icon) > -1)
```
richardxia, line 154:
```suggestion
    ? actions.filter(a => filterActions.includes(a.icon))
```

I think `Array.prototype.includes()` is probably a clearer method to use here.

app/models/Organization.ts
```diff
@@ -74,3 +83,74 @@ export const getResourceLocations = (org: Organization): LocationDetails[] => {
...
+//         alert(`Issue verifying ${itemType}. Please try again.`);
+//       }
+//     });
+// };
+
+export const getResourceActions = (
+  resource: Organization,
+  service?: Service,
+  filterActions?: string[],
+) => {
```
richardxia, line 106: Maybe add the return type annotation here? Otherwise, someone reading this can't immediately see what it returns without having to read the body of the function.

app/pages/OrganizationListingPage.tsx
```diff
@@ -35,6 +38,18 @@ export const OrganizationListingPage = () => {
...
+  const actionBarProps: ActionBarProps = {
+    organization: org,
+    onClickAction: action => {
+      switch (action.icon) {
+        case 'feedback':
+          setFeedbackModalOpen(true);
+          break;
+        default:
+          break;
+      }
```
richardxia, line 50: This feels a little bit sketchy, since we have a separate mechanism for specifying the onOnclick handler for actions. What do you think about changing up the structure of the code so that we construct the list of actions in this component and create the feedback action with a `handler` method that calls the `setFeedbackModalOpen()` function, as opposed to constructing the list of actions in `Organization.ts`?

Aside from cleaning up the onClick handler code, I think moving the OrganizationActions here makes more sense than having them in models/Organization.ts, since the actions are really properties of the page, not of the organization itself. For instance, none of the actions directly "modify" the organization as stored in the DB, and half of the actions only make sense in the context of a web page, such as opening the feedback modal, activating the page printing method, or opening up the phone's telephone app via the `tel:` protocol link.

Here's a rough sketch of what I was thinking:

```js
export const OrganizationListingPage = () => {
  const [feedbackModalOpen, setFeedbackModalOpen] = useState(false);
  const allActions = getResourceActions(organization, setFeedbackModalOpen);  // The latter is attached to the feedback action
  const sidebarActions = allActions.filter(a => [/* types here */].include(a.icon));
  const mobileActions = allActions.filter(a => [/* types here */].include(a.icon));

  return (
    /* Actual JSX code where the actions are passed into the appropriate Action component */
  )
}
```

Although it adds more stuff to the page-level component, I think it would simplify the lower-level components.

richardxia (approved):
Nice job on cleaning this up and unifying the two, similar components. I have a suggestion for simplifying some of the onClick handler code, but aside from that, everything LGTM!

app/models/Organization.ts
```diff
@@ -74,3 +83,74 @@ export const getResourceLocations = (org: Organization): LocationDetails[] => {
     recurringSchedule: org.recurringSchedule,
   }));
 };
+
+// TODO Can we remove this?
```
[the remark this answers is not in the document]
lexholden: Since we still have the API to verify in the backend, I'd wondered whether removing this should be a standalone issue.

But agreed, removing, can re-add clean if required down the line.

app/models/Organization.ts
lexholden: +1, I'm all for getting rid of lodash everywhere, I think it does more harm than good.

app/pages/OrganizationListingPage.tsx
lexholden: Yeah very good point, this left a sour taste in my mouth but I was reluctant to change too much at once.

The only thing I disagree with in your approach, passing in the setFeedbackModalOpen felt a little unextensible, so I've simplified ActionBar to just take actions[] and onActionClick and emit the action clicked. I removed the `handler` key from action alltogether, so only links are handled by the component itself.

richardxia: Yeah, your new changes here LGTM, I wasn't a huge fan of how my suggestion was passing `setFeedbackModalOpen` into `getResourceActions()`, and your way looks like a better way of accomplishing a similar reduction. Thanks for updating it!

PR #1101, opened 2022-02-17
Move transformSchedule business logic to ts Schedule model

lexholden:


schroerbrian (approved):
Another helpful improvement. LGTM!

PR #1104, opened 2022-02-28
Multiple location search results

schroerbrian:
The main thrust of this PR is to consolidate the styles/layout of the Search Results page and the Service Pathway results page (accessed from the cards on the homepage). This PR also adds styling/layout on these pages to match the multiple location work designs.

I've abstracted the JSX/logic from these pages and moved it into the `components/search` directory. We can probably abstract these two pages more in the future, but I think we're in a good spot for now. 

I think this will be a good MVP and could potentially go to prod while I work on stage two of the multiple location work, which will include more geolocation work and the actual addition of secondary addresses in the UI.

**Note**: _Some of the new files are actually files that have been moved to new directories, though in some cases I added changes. Sorry for the confusion with that!_

Here's some screenshots of both the search results and the pathway results across screen sizes: 
(Figma designs: https://www.figma.com/file/eniqBBLxz4eHCqZnWebcWW/Multiple-Address-Mapping?node-id=263%3A1254)

Screen Shot 2022-03-25 at 4 00 16 PM

Screen Shot 2022-03-25 at 3 45 15 PM

Screen Shot 2022-03-25 at 4 01 26 PM


app/styles/utils/_mixins.scss
```diff
@@ -280,3 +283,12 @@ $prefix-for-spec:      true !default; // required for keyframe mixin
...
 }
+
+@mixin hide-scrollbar {
+  scrollbar-width: none;     /* Firefox */
+  -ms-overflow-style: none;  /* Internet Explorer 10+ */
+  &::-webkit-scrollbar {     /* WebKit */
+    width: 0;
+    height: 0;
+  }
+}
```
richardxia, line 294: I think we probably don't want to do this, since it could result in usability issues for non-mobile, non-macOS users with short enough screens or windows. Scrollbar behavior is unfortunately a very complex topic, since it depends on a multitude of factors including browser, operating system, and [whether a physical mouse is plugged in to a macOS machine](https://www.nytimes.com/2017/02/02/technology/personaltech/solving-the-case-of-the-disappearing-scroll-bars.html).

Instead, I think it's better to just leave these properties alone and instead just use `[overflow: auto](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow#values)`, which generally only shows the scroll bar when the content within the element is larger than the size of the element. This at least ensures that users understand content is overflowing while still hiding the scroll bar when the content fits within the element.

app/styles/utils/_layoututils.scss
```diff
@@ -14,6 +15,8 @@ $padding-xlarge: 26px;
 $padding-xxlarge: 36px;
 $max-container-width: 1280px;
 
+$header-height: 70px;
+$header-mobile-height: 50px;
```
richardxia, line 19: Although we probably don't want to do it in this PR, we may want to consider migrating to [CSS custom properties](https://developer.mozilla.org/en-US/docs/Web/CSS/--*) (a.k.a. CSS variables), which have the advantage of being able to observe media queries, since they are actually built-in to CSS itself. We currently make use of this in the sheltertech.org site, and it's actually very nice, since we can reuse the same variables between desktop and mobile views: https://github.com/ShelterTechSF/sheltertech.org/blob/714aefc95629b05c10dc207c007ad9c2f59d5804/src/stylesheets/global.css#L74-L86

richardxia (commented):
Nice job on consolidating these components!

I was wondering if there's any chance that you could break out a separate, smaller PR to _just_ rename and move around the files, with no other content changes? It's a bit challenging to review a PR of this size, and it doesn't help that there are enough changes to the moved files that GitHub no longer recognizes some of them as being the same file. Even for the files that GitHub can detect were originally the same file, creating a separate PR to just move them will also clean up the import path changes.

If that's too much trouble, then no worries, just let me know and I can try to review this PR as is.

schroerbrian:
> Nice job on consolidating these components!
> 
> I was wondering if there's any chance that you could break out a separate, smaller PR to _just_ rename and move around the files, with no other content changes? It's a bit challenging to review a PR of this size, and it doesn't help that there are enough changes to the moved files that GitHub no longer recognizes some of them as being the same file. Even for the files that GitHub can detect were originally the same file, creating a separate PR to just move them will also clean up the import path changes.
> 
> If that's too much trouble, then no worries, just let me know and I can try to review this PR as is.

Yep, that's a great idea. I will create a separate PR to handle the name/path changes and will add you as a reviewer.

app/styles/utils/_layoututils.scss
schroerbrian: That sounds great. I will keep that in mind next time I want to create/edit a CSS var

app/components/search/Pagination/ResultsPagination.scss
```diff
@@ -0,0 +1,91 @@
+// This cannot be a Sass module as it styles elements
+// included by Pagination, a 3rd party component
```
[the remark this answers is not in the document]
schroerbrian: That's awesome. Thanks for the tip. I've gone ahead and converted the file to a module. I've also nested the global Pagination styles within the component's parent class to be on the safe side.

app/styles/utils/_mixins.scss
schroerbrian: Good point. Accessibility and usability are key especially for a service like ours. Removed.

app/pages/ServiceDiscoveryForm/ServiceDiscoveryForm.jsx
```diff
@@ -22,7 +22,6 @@ const ServiceDiscoveryForm = ({ match }) => {
       eligibilities={eligibilities}
       subcategories={subcategories}
       steps={category.steps}
-      subcategorySubheading={category.subcategorySubheading}
```
richardxia, line 25: I think you accidentally reverted some of the changes you made in https://github.com/ShelterTechSF/askdarcel-web/pull/1106.

app/pages/HomePage/HomePage.tsx
```diff
@@ -11,14 +11,14 @@ import ResourceList from './components/ResourceList/ResourceList';
 
 const covidResources = [
   { name: 'Food', icon: 'food', categorySlug: 'food-resources' },
-  { name: 'Shelters', icon: 'bed', categorySlug: 'shelter-resources' },
-  { name: 'Showers, Hygiene and other Services', icon: 'shower', categorySlug: 'hygiene-resources' },
+  { name: 'Shelter', icon: 'bed', categorySlug: 'shelter-resources' },
+  { name: 'Showers and Restrooms', icon: 'shower', categorySlug: 'hygiene-resources' },
```
richardxia, line 15: And I think these changes accidentally revert https://github.com/ShelterTechSF/askdarcel-web/pull/1105.

schroerbrian:
> I haven't reviewed the whole PR yet, but I figured I might as well post the comments I have so far, in case you have time to work on it before the next time I have a chance to continue reviewing.
> 
> Thanks again for cleaning up the diff by merging in the file path renaming stuff first, it helps a lot!

Sounds great. Feel free to make comments as you go since I'll likely have more time to address them than you will to make them!

app/components/search/SearchAndServiceResultsPage.module.scss
```diff
@@ -0,0 +1,84 @@
...
+
+.listIcon {
+  background-position: 33px 33px;
+  background-image: url(../../assets/img/ic-map.png);
+}
+
+.mapIcon {
+  background-position: 32px 33px;
+  background-image: url(../../assets/img/ic-list.png);
+}
```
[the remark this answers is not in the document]
schroerbrian: That was probably due to botched image editing on my part :D. I went ahead and combined the active/non-active icon states into single SVG files and now it looks a lot cleaner 
image
image

app/pages/ServiceDiscoveryResults/ServiceDiscoveryResults.jsx
```diff
@@ -88,14 +88,35 @@ const InnerServiceDiscoveryResults = ({
...
+            aria-label="list icon"
+            onClick={() => setExpandList(true)}
+          />
+          <span
+            className={`${styles.mapIcon} ${!expandList ? styles.activeView : ''}`}
+            role="button"
+            tabIndex={0}
+            aria-label="map icon"
+            onClick={() => setExpandList(false)}
+          />
```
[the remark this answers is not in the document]
schroerbrian: Good call on the button. I added a button parent class. I'm still using spans as the child element – I'm giving them a background-image prop and changing the background-position value depending on active state, since a single icon image file contains both the active and non-active image. If you think it'd be better to have the active/non-active images be separate files and to use an `img` tag w/dynamic src attribute, let me know and I'll make that change instead!

I like the idea of a segmented control component. If I get some downtime, I will work on that. It'd be useful. It would help to remove repeated code between the ServiceDiscoveryResults and the SearchResultsPage, for one thing

app/pages/ServiceDiscoveryResults/ServiceDiscoveryResults.jsx
```diff
@@ -105,50 +126,21 @@ const InnerServiceDiscoveryResults = ({
...
+            eligibilities={eligibilities}
+            subcategories={subcategories}
+            subcategoryNames={subcategoryNames}
+          />
 
-          </div>
           <div className={styles.results}>
-            <SearchResults />
+            <SearchResults
+              props
```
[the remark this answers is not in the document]
schroerbrian: I'm trying to remember why I did that. I think eslint was giving me some kind of odd error message and I added both the spread operator and the props to help me circumvent it 🤔 . Anyhow, I removed the props and spread operator and now it's working with no eslint issues, so it's all good.

app/components/search/Sidebar/Sidebar.tsx
```diff
@@ -0,0 +1,96 @@
...
+        <ClearAllFilters />
+        <div className={styles.filterGroup}>
+          <div className={styles.filterTitle}>Availability</div>
+          <OpenNowFilter attribute="open_times" />
+        </div>
+
+        <div className={styles.filterGroup}>
+          <div className={styles.filterTitle}>Eligibilities</div>
+          {eligibilityRefinementJsx}
+        </div>
```
richardxia, line 84: It looks like this causes the `Eligibilities` title to show up even if you don't have any eligibility filtering options available. I think you need to guard the whole filter group based on whether `eligibilityRefinementJsx` is null. Same goes with the Categories section below.

app/pages/SearchResultsPage/SearchResultsPage.tsx
```diff
@@ -1,54 +1,108 @@
...
 };
+
+/** Stateless inner component that just handles presentation. */
+const InnerSearchReslts = ({
+  history, userLocation, lastPush, setLastPush, expandList, setExpandList, searchState,
+}: {
+  history: any;
+  userLocation: GeoCoordinates;
+  lastPush: number;
+  setLastPush: Function;
```
richardxia, line 50: Let's specify the full function type signature for function types. `Function` doesn't tell us much, since it doesn't say anything about the number of arguments, the types of the arguments, or the return type.

```suggestion
  setLastPush: (time: number) => void;
```

Same comment goes for `setExpandList`.

app/components/search/SearchAndServiceResultsPage.module.scss
```diff
@@ -0,0 +1,95 @@
...
+
+  @media screen and (max-width: $break-tablet-l) {
+    flex-direction: column;
+  }
+}
+
+.results {
+  width: 100%;
+  height: 100%;
+}
```
richardxia, line 95: Does this do anything? I see that this style already existed before this PR, but it doesn't seem to affect anything when I disable these styles.

app/components/search/SearchMap/SearchMap.tsx
```diff
@@ -25,19 +28,35 @@ export const SearchMap = ({ hits, hitsPerPage, page }: {
...
+                  tag += alphabeticalIndex;
+                }
+
+                return tag;
+              };
+
               markers.push(<SearchHitMarker
-                key={hit.id}
+                // eslint-disable-next-line react/no-array-index-key
+                key={`${hit.id}.${i}`}
```
richardxia, line 56: You probably don't want to disable this ESLint warning, since it could cause weird cache invalidation issues in React in the case where addresses get changed around (maybe removed due to additional filters?). I think what you actually want to do is embed something that is unique to the address itself as the key, perhaps the lat/lng?

app/components/ui/Navigation.module.scss
```diff
@@ -3,7 +3,7 @@
 @mixin siteNavShared {
   position: sticky;
   top: 0;
-  z-index: 1;
+  z-index: 3;
```
richardxia, line 6: I don't have a great suggestion for this right now, but I usually view it as a code smell when we have to bump z-indexes. Normally I try to keep z-indexes localized as much as possible, as in within a [single stacking context](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Positioning/Understanding_z_index/The_stacking_context), since otherwise it's really hard to figure out which z-index values are being assigned to components within the same stacking context but in very different files.

One approach I've tried in the past is to have a single Sass file define constants for all the global z-indexes, and then to have each component use their respective constant. E.g.

```scss
$z-index-site-nav: 3;
$z-index-filters-menu: 4;
```

This isn't a perfect solution, since there's no such thing as a global z-index (z-indexes are always scope to their stacking context), but it's at least better than having hardcoded numbers split across multiple files, since we'll have a single place where we can see the ordering of layers.

The real fix would be to restructure our HTML and CSS so that we don't have to reason about floating elements' z-indexes across multiple files, but that's going to be a bigger project.

app/components/search/SearchResults/SearchResults.jsx
```diff
@@ -24,23 +26,54 @@ const transformHits = hits => hits.map(hit => {
...
+    if (centerCoords) {
+      googleMapObject.setCenter(centerCoords);
+    }
+  }, [centerCoords]);
+
   return (
-    <div className={styles.searchResultsContainer}>
-      <div>
-        { hits.map((hit, index) => <SearchResult hit={hit} index={index} key={hit.id} />) }
+    <div className={styles.searchMapContainer}>
```
richardxia, line 44: Maybe rename this to `searchResultsAndMapContainer`? When I was reading the CSS file, I thought this only contained the map, but it actually contains both the map and the text results.

app/components/search/SearchResults/SearchResults.module.scss
```diff
@@ -1,85 +1,136 @@
...
+@import '~styles/utils/_helpers.scss';
 
-.searchResultsContainer {
+.searchMapContainer {
   position: relative;
+  display: flex;
+  flex-direction: row;
+  @media screen and (max-width: $break-tablet-s) {
+    flex-flow: column-reverse;
+  }
```
richardxia, line 9: I don't know how comfortable you are with CSS grid, but I think it would be a better fit for the layout we're trying to accomplish here than flexbox. For single-dimensional layouts like this, we still benefit from using CSS grid because it allows the parent element be responsible for positioning and sizes of its children, and then the children don't have to worry about it.

If you're not familiar with CSS Grid, I've found the [CSS Tricks page on grid](https://css-tricks.com/snippets/css/complete-guide-grid/) to be a super useful resource.

I think you could make it look something like this:

```css
.searchMapContainer {
  display: grid;

  /* Large layout */
  grid-template-columns: 6fr 4fr;  /* Replaces the width: 60% on .searchResultsContainer */
  grid-areas: "results map";
  grid-column-gap: 15px;  /* Replaces the margin-right: 15px on .searchResultsContainer */

  /* Phone layout */
  @media screen and (max-width: $break-tablet-s) {
    grid-template-columns: auto;
    grid-template-rows: auto auto;
    grid-areas:
      "map"
      "results";
  }
}

.searchResultsContainer {
  grid-area: "results";
}

.searchMap {
  grid-area: "map";
}
```

app/components/search/SearchResults/SearchResults.module.scss
```diff
@@ -1,85 +1,136 @@
...
+}
+
+.searchResultsContainer {
   display: flex;
   flex-direction: column;
-  width: 100%;
+  flex-grow: 1;
```
richardxia, line 15: I'm not sure this does anything if you're also setting `width: 60%` on this.

app/components/search/SearchResults/SearchResults.module.scss
```diff
@@ -1,85 +1,136 @@
...
+    font-size: 18px;
+    padding-bottom: 24px;
+  }
+
+  &_address {
+    border-bottom: 1px solid $color-grey3;
+    padding: 24px 0;
+  }
+
+  &_addressLast {
```
richardxia, line 76: I know that this isn't complete yet and that you're working on it for the next PR, but could this be replaced with [`:last-child()`](https://developer.mozilla.org/en-US/docs/Web/CSS/:last-child)?

richardxia (approved):
Sorry for the late review, but this looks pretty great! I've finally gone through and reviewed I think almost all the files. Apologies in advance if I commented on some stuff that was already in the codebase before but that GitHub wasn't able to detect that the code was moved. I think anything that was already in the existing code is fine to leave as is, since it's obviously not causing any problems now.

Thanks again for tackling all this! It's a huge improvement to consolidate the two results pages!

app/pages/ServiceDiscoveryResults/ServiceDiscoveryResults.jsx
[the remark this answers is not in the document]
richardxia: Awesome, yeah, I think it's fine to continue using the image sprite approach. We may want to add back in some of the ARIA stuff, though, since then otherwise it may not be clear to screen readers or search engine crawlers what these buttons actually do.

app/components/search/SearchMap/SearchMap.tsx
```diff
@@ -25,19 +28,35 @@ export const SearchMap = ({ hits, hitsPerPage, page }: {
...
+              const appendAddressLabel = () => {
+                // Appends letter to index number if there are multiple addresses per service
+                let tag = `${page * hitsPerPage + index + 1}`;
+                if (i > 0) {
+                  const alphabeticalIndex = (i + 9).toString(36).toUpperCase();
+                  tag += alphabeticalIndex;
+                }
+
+                return tag;
+              };
```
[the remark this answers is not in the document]
schroerbrian: Good call; I've changed it. Removes a few LOCs as well

app/components/search/SearchAndServiceResultsPage.module.scss
```diff
@@ -0,0 +1,95 @@
...
+  padding: 15px 25px;
+  height: auto;
+}
+
+.mapListToggleBtn {
+  background: 0;
+  padding: 0;
+  margin: 0;
+  width: auto;
+  height: auto;
```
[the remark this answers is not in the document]
schroerbrian: I agree that it'd be nice to remove the defaults at some point. I'm keeping a list of general improvements that we can make. I've added the explanatory comment for now, thanks!

app/components/search/Sidebar/Sidebar.tsx
```diff
@@ -0,0 +1,96 @@
...
+  isSearchResultsPage, eligibilities = [], subcategories = [], subcategoryNames = [],
+}: {
+  isSearchResultsPage: boolean;
+  eligibilities?: Array<object>;
+  subcategories?: Array<object>;
+  subcategoryNames?: Array<string>;
+}) => {
+  const [filterActive, setFilterActive] = useState(false);
+  const transformItems = (items: any): any => items.sort((a:{label: string},
+    b:{label: string}) => a.label.localeCompare(b.label));
```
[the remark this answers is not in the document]
schroerbrian: That is a good point. I have updated this in line with your suggestion.

app/components/search/Sidebar/Sidebar.tsx
```diff
@@ -0,0 +1,96 @@
...
+
+const Sidebar = ({
+  isSearchResultsPage, eligibilities = [], subcategories = [], subcategoryNames = [],
+}: {
+  isSearchResultsPage: boolean;
+  eligibilities?: Array<object>;
+  subcategories?: Array<object>;
+  subcategoryNames?: Array<string>;
+}) => {
+  const [filterActive, setFilterActive] = useState(false);
```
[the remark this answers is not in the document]
schroerbrian: Changed to filterMenuVisible for improved clarity.

app/components/search/Sidebar/Sidebar.module.scss
```diff
@@ -0,0 +1,151 @@
...
+
+  to {
+    transform: translateY(70px);
+  }
+}
+
+.sidebar {
+  flex-shrink: 0;
+  position: sticky;
+  height: calc(100vh - 70px);
```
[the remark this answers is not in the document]
schroerbrian: Yep

app/components/ui/Navigation.module.scss
schroerbrian: It's a tricky issue to solve for. At previous companies I've worked out, we used general vars which at least act as enums of sorts to limit all manner of z-index values. I like the specific var names you suggest and have created a couple of them in our layouts.scss file, which I think is a small improvement for now, short of a larger restructure of HTML/CSS

app/components/search/SearchResults/SearchResults.jsx
schroerbrian: Thanks for the call out! Changed it.

app/components/search/SearchResults/SearchResults.module.scss
schroerbrian: That's a great bit of magic. I've only scratched the surface of grid thus far. Your sketch worked more or less out of the box and removed several lines of CSS

app/components/search/SearchResults/SearchResults.module.scss
schroerbrian: I've added a to-do for now and will address in the next stage.

app/components/search/Sidebar/Sidebar.module.scss
```diff
@@ -0,0 +1,151 @@
...
+      z-index: 4;
+      position: fixed;
+      height: 100vh;
+      top: 0;
+      overflow-y: auto;
+      box-shadow: 0 0 20px rgb(0 0 0 / 55%);
+      animation-duration: 200ms;
+      animation-name: slideRight;
+    }
+  }
```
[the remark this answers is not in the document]
schroerbrian: Yep that makes total sense. I had tried doing so previously, but moved on to something else and put it aside. I've gone and updated it now though and it works well and feels more logical.

app/components/search/Sidebar/Sidebar.module.scss
```diff
@@ -0,0 +1,151 @@
...
+  top: $header-height;
+  border-right: 1px solid $color-grey3;
+  padding: 20px 25px 50px 30px;
+  overflow-y: auto;
+
+  @media screen and (max-width: $break-tablet-l) {
+    height: auto;
+    width: 60%;
+    padding: 0 30px;
+    z-index: 2;
```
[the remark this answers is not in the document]
schroerbrian: I've removed the CSS rule in the .showFilters state below, but kept it here since the filters bar is sticky and scrolls down with the page so it needs to overlay the search results like so:
image
(I will double check with the designers if they wanted it to scroll down but that was my assumption)

app/components/search/Sidebar/Sidebar.module.scss
```diff
@@ -0,0 +1,151 @@
...
+      }
+
+      .closeIcon {
+        display: inline-block;
+      }
+
+      span {
+        display: none;
+      }
+    }
```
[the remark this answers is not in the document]
schroerbrian: Yep, done!

app/components/search/Sidebar/Sidebar.module.scss
```diff
@@ -0,0 +1,151 @@
...
+
+.filterGroup {
+  margin-top: 12px;
+}
+
+.filterTitle {
+  font-size: 16px;
+  font-weight: bold;
+  padding: 16px 0;
+}
```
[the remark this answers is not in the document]
schroerbrian: I've fixed up the filters UI and margin/padding. Looks better and is cleaner. Example below on tablet view. Thanks!

image

app/pages/SearchResultsPage/SearchResultsPage.tsx
schroerbrian: Fixed. Thanks!

app/components/search/SearchAndServiceResultsPage.module.scss
schroerbrian: Doesn't seem to. I commented it out and tested and there were no visible changes.

app/components/search/SearchMap/SearchMap.tsx
schroerbrian: I might be being paranoid, but I changed it to the `id + lat/lng + address_1 + address_2` in case we get services with multiple addresses within the same building

schroerbrian:
> Sorry for the late review, but this looks pretty great! I've finally gone through and reviewed I think almost all the files. Apologies in advance if I commented on some stuff that was already in the codebase before but that GitHub wasn't able to detect that the code was moved. I think anything that was already in the existing code is fine to leave as is, since it's obviously not causing any problems now.
> 
> Thanks again for tackling all this! It's a huge improvement to consolidate the two results pages!

Thanks! Your comments were super helpful. I really appreciate it. I think I've addressed/resolved all of them.

app/components/search/SearchMap/SearchMap.tsx
```diff
@@ -2,17 +2,20 @@ import React, { ReactElement } from 'react';
...
-  hits: SearchHit[];
-  hitsPerPage: number;
-  page: number;
+export const SearchMap = ({
+  hits, hitsPerPage, page, setMapObject,
+}: {
+    hits: SearchHit[];
+    hitsPerPage: number;
+    page: number;
+    setMapObject: Function;
```
richardxia, line 18: Could we change this to a more specific function signature, even if we don't know exactly what a map object is? This at least makes it clearer that there's a single argument and that the return type is void.

```suggestion
    setMapObject: (map: any) => void;
```

app/components/search/Sidebar/Sidebar.tsx
```diff
@@ -0,0 +1,98 @@
...
+import filtersIcon from 'assets/img/filters-icon.png';
+import styles from './Sidebar.module.scss';
+
+const Sidebar = ({
+  isSearchResultsPage, eligibilities = [], subcategories = [], subcategoryNames = [],
+}: {
+  isSearchResultsPage: boolean;
+  eligibilities?: Array<object>;
+  subcategories?: Array<object>;
+  subcategoryNames?: Array<string>;
```
richardxia, line 18: Nit: For simple array types, let's use the shorter syntax: `ItemType[]`.

```suggestion
  eligibilities?: object[];
  subcategories?: object[];
  subcategoryNames?: string[];
```

richardxia (approved):
Just a few small comments, but this looks pretty good to me! Thanks again for working on this!

app/components/search/Sidebar/Sidebar.tsx
schroerbrian: Got it. Good call.

app/components/search/SearchMap/SearchMap.tsx
schroerbrian: Yep, missed that one. thanks!

PR #1116, opened 2022-04-14
Replace SearchBox widget inside the InstantSearch component so that A…

schroerbrian:
…lgolia search accepts query parameter.

Discovered a bad bug where the query was not being passed to Algolia when searching via the search box in the nav. I had removed this component from the SearchPage originally, having thought it wasn't doing anything. My staging algolia index seems messed up for text searches in particular, so this went undiscovered for a while


richardxia (approved):
Nice catch, LGTM!

PR #1117, opened 2022-04-14
Multiple location results by distance

schroerbrian:
~~****** Keeping this as a draft until the current work on Master/Staging is deployed to Prod ******~~

This PR does 3 things:

1. Adds search by radius radio buttons in the filters, which allow users to bound their search by 4 blocks (400 meters), 1 mile, and 3 miles
2. Adds a no results found UI
3. Fixes the map flashing bug where a the map flashes from the default mid point in SF our app sets to the user's actual location

Screen Shot 2022-04-14 at 5 07 44 PM

Screen Shot 2022-04-14 at 5 01 30 PM


app/App.tsx
```diff
@@ -58,11 +58,16 @@ export const App = () => {
   const [hamburgerOpen, setHamburgerOpen] = useState(false);
   const [popUpMessage, setPopUpMessage] = useState<PopupMessageProp>({ message: '', visible: false, type: 'success' });
   const [userLocation, setUserLocation] = useState<GeoCoordinates>(COORDS_MID_SAN_FRANCISCO);
+  const [userLocationPromiseReturned, setUserLocationPromiseReturned] = useState(false);
```
schroerbrian, line 61: I'm unsure if this is the best/most React-y way to do this – I'm using this AppContext boolean property to determine whether we can initialize the map (and thus set the center to the user's actual coords or SF default coords) to avoid the existing map flash bug that will become more apparent as our site becomes more "map-centric". Thoughts or suggestions are welcome

app/pages/SearchResultsPage/SearchResultsPage.tsx
```diff
@@ -25,24 +25,42 @@ export const SearchResultsPage = () => {
   const [lastPush, setLastPush] = useState(Date.now());
   const { search } = useLocation();
   const [expandList, setExpandList] = useState(false);
-  const searchState = useMemo(() => qs.parse(search.slice(1)), [search]);
+
+  // Todo: I'm not sure if this is the greatest solution; querying the searchState
+  // for the configure.aroundRadius property causes a TS type warning unless I
+  // define the interfaces below
+  interface ConfigureState {
```
schroerbrian, line 32: I added these interfaces to fix an TS warning on line 43

app/pages/SearchResultsPage/SearchResultsPage.tsx
```diff
@@ -82,17 +102,16 @@ const InnerSearchReslts = ({
       }}
       createURL={(state: any) => `search?${qs.stringify(state)}`}
     >
-      {userLocation ? (
```
schroerbrian, line 85: If a user refuses to share their location, we set userLocation coords to a default SF location. Thus, we will always have a userLocation and this check is moot

app/components/search/SearchResults/SearchResults.module.scss
```diff
@@ -32,6 +32,15 @@
...
 }
 
+.noResultsFound {
+  display: none;
+  &.showNoResultsMsg {
+    display: block;
+    margin: 35px auto 65px;
+    text-align: center;
+  }
+}
```
richardxia, line 42: Since these are meant to be attached to the same element, it might be easier to reason about if you separate the displayedness of the component from the other properties of element, even if those other properties are not visible when the element is hidden. What if you named and organized the CSS rules like this?

```suggestion
.noResultsMessage {
  margin: 35px auto 65px;
  text-align: center;

  &.hidden {
    display: none;
  }
}
```

and updated the JSX code's logic accordingly, such that the `hidden` class gets applied by default, and is only removed when there are no results?

richardxia (approved):
I left a few comments, but this mostly LGTM!

app/pages/SearchResultsPage/SearchResultsPage.tsx
```diff
@@ -25,24 +25,42 @@ export const SearchResultsPage = () => {
...
+    [key: string]: any;
+  }
+
+  interface SearchState {
+    configure?: ConfigureState;
+    [key: string]: any;
+  }
+
+  const searchState: SearchState = useMemo(() => qs.parse(search.slice(1)), [search]);
+  const [searchRadius, setSearchRadius] = useState(searchState?.configure?.aroundRadius || 'all');
```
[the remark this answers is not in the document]
schroerbrian: Ah nice. I hadn't known about that operator. Thanks!

That is a good catch. I added code to set any new text queries to the existing search state when a user enters a new query. I think it would probably be better to use one source of truth across components (perhaps using an appContext prop?) rather than doing this; I spent some time trying to rig that up but couldn't quite get it right without bugs. Thus, in the interest of time I went with this quicker solution.... I might be revisiting this again as part of my next chunk of work, as we will need to add another search box to the page per the designs. I could try to refactor it then. Not sure. Let me know if you have thoughts. Thank you

app/components/search/Sidebar/Sidebar.tsx
```diff
@@ -10,8 +10,11 @@ import filtersIcon from 'assets/img/filters-icon.png';
 import styles from './Sidebar.module.scss';
 
 const Sidebar = ({
-  isSearchResultsPage, eligibilities = [], subcategories = [], subcategoryNames = [],
+  setSearchRadius, searchRadius, isSearchResultsPage,
+  eligibilities = [], subcategories = [], subcategoryNames = [],
 }: {
+  setSearchRadius: (radius: any) => void;
```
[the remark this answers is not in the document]
schroerbrian: yep, good catch

app/components/search/SearchResults/SearchResults.module.scss
schroerbrian: Much better, thanks

app/pages/SearchResultsPage/SearchResultsPage.tsx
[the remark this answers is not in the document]
richardxia: Yeah, I think we can revisit it later. I think I had made an attempt at that in the past, and I can to the conclusion that we have to do something special on the initial page load, since we should first try to read from the URL, but after the initial page load, we should treat our internal app state as the source of truth. I think there was also some weird interaction with react-router as well, so we probably need to keep that in mind.

schroerbrian:
We've deployed the latest to Prod, so I'm going to merge this in now.

PR #1119, opened 2022-05-02
Stop using legacy eligibility field.

richardxia:
Eligibility used to be a text field on Service. It was later replaced
with a many-to-many relationship between Services and a dedicated
Eligibilities table, but a couple places in the askdarcel-web app still
read from the legacy field.

This removes the last instances of this older field.


schroerbrian (approved):
LGTM!

PR #1127, opened 2022-05-10
2093 Fix broken resource/organization links on search results page.

schroerbrian:
This bug was introduced by my consolidation of the code for the search results and "service pathway results" pages. I missed that orgs can be returned by search queries as well.
https://app.shortcut.com/sheltertech/story/2093/cannot-open-organizations-found-by-top-level-search


richardxia (approved):
LGTM! It'd be really nice to get more of the code migrated to TypeScript so that we can get the type checker to yell at us about these kinds of things!

PR #1128, opened 2022-05-16
Google translate prototype

schroerbrian:
This will only be viewable in Staging. If you have any suggestions for a better way to check that we're on the staging environment, I'm all ears! Perhaps someday we'll want to add a dedicated Staging (or QA eventually) entry in our whitelabel.ts file. 

I tried to keep logic, script tags, JSX strictly within the Translate component to avoid dirtying up the rest of our repo with what is essentially a prototype that has a not very programmatic way of checking that the app is running in Staging. 

Screen Shot 2022-05-16 at 2 24 40 PM
Localhost hostname is not visible/ not in DOM

Screen Shot 2022-05-16 at 2 24 48 PM
Staging hostname (using /etc/hosts) widget is visible


app/components/ui/Translate.tsx
```diff
@@ -0,0 +1,39 @@
...
+// Note: At least while testing, this is a <li> element as it goes within the
+// nav <ul> site links element.
+
+const Translate = () => {
+  const isStaging = window.location.hostname === 'staging.sfserviceguide.org';
+  if (isStaging) {
+    return (
+      <li className="googleTranslateContainer">
+        {/* todo: When we go live with Google Translate OR we have a dedicated test environment,
+                  the scripts within this Helmet snippet should be moved to index.html  */}
```
richardxia, line 20: It might still need to be gated based on the specific whitelabeled site.

app/components/ui/Translate.tsx
```diff
@@ -0,0 +1,39 @@
+import React from 'react';
+import { Helmet } from 'react-helmet-async';
+
+// This is a prototype component for testing out the Google Translate widget. The logic below
+// dictates that this component only be displayed on staging; i.e. the widget will be displayed IF
+// location.hostname === 'staging.sfserviceguide.org'. To test locally point
+// staging.sfserviceguide.org to your local host IP in your /etc/hosts file; i.e.,
+// 127.0.0.1       staging.sfserviceguide.org
+// Then, navigate to http://staging.sfserviceguide.org:8080 to see the widget.
```
richardxia, line 9: Might be easier to test this by just modifying the code to be `if (true)`. Browsers might complain if you try to visit a site with http that you've visited in the past with https, since a common attack is to downgrade a secure connection to a non-secure one and to man-in-the-middle the user.

richardxia (approved):
A few small comments, but this is a pretty straightforward component, so this LGTM!

app/components/ui/Translate.tsx
schroerbrian: > A few small comments, but this is a pretty straightforward component, so this LGTM!

Good points, per usual. I've clarified/edited the comments a bit. Will merge in now.

PR #1132, opened 2022-06-07
Fix broken new organization page. Improve a few UI issues.

schroerbrian:
I'm not sure if we should fix it this way or perhaps change the `organizations/new` path altogether so it won't be matched by the `organization/:id` path?

I also fixed an issue where the buttons overlay the address modal on smaller screens, and I added disabled styling to the submit button since creating an org can take some time and lead to a confusing experience for the user.

button overlay bug:
Screen Shot 2022-06-06 at 6 06 47 PM


richardxia (approved):
Nice catch! Thanks for adding the extra comment to make sure we don't break this again in the future!

PR #1133, opened 2022-06-09
Hide Add Organization button per request from content team.

schroerbrian:
Hides "Add Org" button per Amanda's request. Fortunately it only exists in one component.
image
Screen Shot 2022-06-08 at 7 08 53 PM


richardxia (approved):
LGTM!

PR #1138, opened 2022-07-15
Add patient handouts

schroerbrian:
Borrows from the EditCollection/EditCollectionHOC components. This PR uses functional components and TS. Since the existing EditCollection component uses logic for editing a collection child of a Resource, it is slightly different than this component, which is intended for a Service. Perhaps we can find a way to abstract this component in the future to also handle Resource child collections


app/components/edit/EditPatientHandout.tsx
```diff
@@ -0,0 +1,54 @@
...
+  id: string;
+  language: string;
+  link: string;
+  [key: string]: string;
+}
+
+const EditPatientHandout = ({ index, item, handleItemChange }: {
+  index: number;
+  item: patientHandout;
+  handleItemChange: (handoutIndex: number, handout: patientHandout) => null;
```
richardxia, line 13:
```suggestion
  handleItemChange: (handoutIndex: number, handout: patientHandout) => void;
```

You'll want `void`, not `null`, for functions whose return values you're not supposed to use. A return value of `null` means that the function must return `null`, whereas a return value of `void` is special and means that the function can return anything it wants (including the default return value of `undefined`), but the caller is not allowed to make use of the returned value at all.

app/components/edit/EditPatientHandout.tsx
```diff
@@ -0,0 +1,54 @@
...
+}) => {
+  const handleFieldChange = (
+    e: React.ChangeEvent<HTMLInputElement>,
+    handout: patientHandout,
+  ) => {
+    const { value } = e.target;
+    const { field } = e.target.dataset;
+    if (field) {
+      if (handout[field] || value !== item[field]) {
```
richardxia, line 22: I'm getting a little bit confused between the usage of `handout` vs `item`. As far as I can tell, `handout` will always be the same as `item`, where `item` is a prop on the `EditPatientHandout` component whereas `handout` is just a parameter directly on this handler function.

I think the code might actually be a lot simpler if you just got rid of the `handout` parameter and closed over the `item` prop. Not only does this avoid confusion between the two variables that always hold the same value, it'll even make your JSX code below a bit cleaner because you won't have to construct an arrow function for `onChange` prop. That is, you could just write `<input onChange={handleFieldChange}>`.

app/components/edit/EditPatientHandout.tsx
```diff
@@ -0,0 +1,54 @@
...
+        const newHandout = {
+          ...handout,
+        };
+        newHandout[field] = value;
```
richardxia, line 26:
```suggestion
        const newHandout = {
          ...handout,
          [field]: value,
        };
```

Nit: This might be personal preference, but I think it's a bit clearer to just initially define the object with all its final property values than to define it as a clone and then immediately mutate it.

app/components/edit/EditPatientHandout.tsx
```diff
@@ -0,0 +1,54 @@
...
+  item: patientHandout;
+  handleItemChange: (handoutIndex: number, handout: patientHandout) => null;
+}) => {
+  const handleFieldChange = (
+    e: React.ChangeEvent<HTMLInputElement>,
+    handout: patientHandout,
+  ) => {
+    const { value } = e.target;
+    const { field } = e.target.dataset;
+    if (field) {
```
richardxia, line 21: When would this ever be false? It might be better to throw an exception in that case instead of silently do nothing.

app/components/edit/EditServiceChildCollection.tsx
```diff
@@ -0,0 +1,120 @@
...
+  initialCollectionData,
+  handleCollectionChange,
+  ResourceObjectItem,
+  label,
+  blankTemplateObj,
+  buttonText,
+  propertyKeyName,
+}: {
+  initialCollectionData: any[];
+  handleCollectionChange: (field: any, value: any) => null;
```
richardxia, line 30: This would be a perfect application of generics. If you define the component with a type parameter, say `T`, then you can guarantee that both this collection type (`T[]`) and the handle change function (`(field: string, value: T) => void`) work on objects of that type as well.

Note that there's a bit of an issue with defining generic arrow functions in a TSX file because of a conflict between the TypeScript and JSX syntaxes, so you may have to apply [a workaround](https://stackoverflow.com/a/45576880) to actually type it out.

app/components/edit/EditServiceChildCollection.tsx
```diff
@@ -0,0 +1,120 @@
...
+    if (firstUpdate.current) {
+      firstUpdate.current = false;
+      return;
+    }
+
+    if (updateParentService.current) {
+      // We only want to update the parent service when an item is deleted or modified;
+      // we don't want to call update when an empty template has been added
+      handleCollectionChange(propertyKeyName, resourceCollection);
+    }
```
richardxia, line 51: It's a little bit strange for me to see something like this in a `useEffect()` call. It's a bit hard for me to put it clearly in words, but I think this is working against the fundamental model of React of state and state updates. Most React components encapsulate state via the `useState()` hook (or this.state for class-based components), and they encapsulate state updates with function calls (`setState()` for many kinds of updates, and arbitrary function callbacks for more complex state updates that you may want to encapsulate).

What you have here, instead, is almost an inversion of the model, where you are listening in to specific changes to the state rather than to the individual state update calls. I think this is why you ended up having to define some ex-React persistent state via refs.

Have you considered a different approach where you directly trigger calls to `handleCollectionChange()` in the individual handler methods you've defined here? For example, if you don't want to update the parent collection when adding a new item, you can simply omit the `handleCollectionChange()` call in that case, but in the other two cases explicitly call it. I think this would make the component a bit easier to reason about, since instead of having to reason about it at two layers — the immediate handler functions and then the effect handler —, everything collapses into the immediate handler functions.

app/components/edit/EditPatientHandout.tsx
schroerbrian: Thanks for the reminder, I made a mistake there.

app/components/edit/EditPatientHandout.tsx
schroerbrian: I find it deeply confusing, haha. I copied this over from the `EditPhones.jsx` file: https://github.com/ShelterTechSF/askdarcel-web/blob/2f8c327936d9f729bf7d0e5febb7dcf0846a466f/app/components/edit/EditPhones.jsx#L25

When doing this work, I thought there may be a reason for the two variables, perhaps having to do with a pre-existing vs new handout/phone #, but upon closer examination I can't find any examples where there's a difference between the 2 vars, either here or the EditPhones method for that matter.

app/components/edit/EditPatientHandout.tsx
schroerbrian: Thank you, agreed. I wasn't aware I could use the square bracket syntax to define a key name.

app/components/edit/EditPatientHandout.tsx
schroerbrian: Without this check, the TS compiler yells at me that `Type 'undefined' cannot be used as an index type.ts(2538)` I tried various things to fix this but to no avail. Any suggestions?

app/components/edit/EditServiceChildCollection.tsx
schroerbrian: I added a generic here. The only part where I'm a little unsure is using the `ResourceItem` interface defined on line 21 to extend `T` generic. I needed this as the `createItemComponents` method assumes that all members (T) of the collection will contain either an `id` or `keyId` prop. The `keyId` prop is actually added in the `addItem` method. It's needed for new items created on the client that don't have IDs assigned to them by the DB. Anyhow, let me know if you see any red flags with my implementation here. Thank you!

app/components/edit/EditServiceChildCollection.tsx
schroerbrian: I can't quite remember why I did it this way. I think I did initially try to call the `handleCollectionChange` function within the individual handler functions. But I may have been passing the `resourceCollection` state var into the `handleCollectionChange` function rather than the updated `newCollection`. The `resourceCollection` state var wasn't being updated before the `handleCollectionChange` executing, thus I created this whole roundabout way of doing this. 

Anyhow, invoking the `handleCollectionChange` within the individual handlers and passing in the `newCollection` var works as desired. It also removes quite a bit of code, in addition to following the principles you've identified above. Thanks for calling this out!

app/components/edit/EditServiceChildCollection.tsx
```diff
@@ -0,0 +1,104 @@
...
+      ...resourceCollection.slice(0, index),
+      ...resourceCollection.slice(index + 1),
+    ];
+
+    setResourceCollection(newCollection);
+    handleCollectionChange(propertyKeyName, newCollection);
+  };
+
+  const createItemComponents = (itemComponentCollection: T[]) => (
+    itemComponentCollection.map((item, index) => (
```
richardxia, line 75: Does this need to be a function? You only ever call this a single time per render, so I think you can just inline it. For example:

```tsx
const itemComponents = resourceCollection.map((item, index) => (
  // Everything else that you have here in the function
));

return (
   <>
      <label htmlFor="edit-item">{label}</label>
      <div className="edit--section--list--item--sublist">
        {itemComponents}
      </div>
```

richardxia (commented):
Sorry for the super long delay in reviewing this, but thanks for refactoring this to use a generic. The changes look good to me so far, but I think there's some accompanying backend code that's not here. I tried running this locally against the latest master branch of askdarcel-api, but I get the following error response when trying to add a new handout and saving it:

```
{"error":"Unknown attribute in request: \"patientHandouts\""}
```

I don't see any open PRs on the askdarcel-api repo that look like they add anything related to handouts. Is that still under development, or do I need to check out a specific branch to get it?

app/components/edit/EditServiceChildCollection.tsx
schroerbrian: Oh yes, good catch. No need for that. Thank you.

app/components/edit/EditPatientHandout.tsx
schroerbrian: Great, I have added error throwing if field is falsy. Thanks!

schroerbrian:
> 

Ahh, I'm sorry for not making that clear :(. Actually, as of Friday, there is an API repo [PR](https://github.com/ShelterTechSF/askdarcel-api/pull/640) up that includes the Handouts (called `documents`) property on GETs to the Service endpoint. I have just pulled that down and was able to fetch a Service with Handouts locally. You will need to run `docker-compose run --rm api rake db:setup db:populate` to get it running locally. Note that the PR is incomplete as the `documents` model does not include a `description` property, which we will use to label the handout language. I've made a comment: https://github.com/ShelterTechSF/askdarcel-api/pull/640/files Also, there is no endpoint that accepts creates or updates, yet, however, so you will still see the error you mention above.

Joe/Will are still working on the create/update endpoints for both Patient Handouts (`documents`) and Clinician Actions (`instructions`) . I added a Todo detailing that this f/e code will be slightly altered in the near future. I now know I will also need to include update requests for the handouts, which as we discussed in Slack, will need to go thru the change request interface 😨. I intend to follow the pattern established for [notes](https://github.com/ShelterTechSF/askdarcel-web/blob/a5f9572d21b491dc76beb37d3599233cfcb03ef7/app/pages/OrganizationEditPage.jsx#L955)   Until then, we will see that error that you describe above.

In the future – I'm not sure – does it make sense to not open PRs until necessary backend work is done or to open them with mock data and then open subsequent PRs when the backend work is done? One disadvantage of the latter is that it could backload a bunch of PR work at the end of a project. I suppose it is a case by case thing. 

Thanks Richard for your time and your great reviews!

richardxia:
> Ahh, I'm sorry for not making that clear :(. Actually, as of Friday, there is an API repo [PR](https://github.com/ShelterTechSF/askdarcel-api/pull/640) up that includes the Handouts (called `documents`) property on GETs to the Service endpoint. I have just pulled that down and was able to fetch a Service with Handouts locally. You will need to run `docker-compose run --rm api rake db:setup db:populate` to get it running locally.

Thanks, I see that the PR is merged now, but when I try to rebuild the DB on the latest `master` branch, I'm hitting an error:

```
Created database 'askdarcel_development'
rake aborted!
ActiveRecord::StatementInvalid: PG::UndefinedTable: ERROR:  relation "documents_services" does not exist
LINE 8:  WHERE a.attrelid = '"documents_services"'::regclass
```

I'm not really sure why I'm hitting it but that CI still passed, since it looks like we're running `rake db:setup db:populate` in CI as well.


> In the future – I'm not sure – does it make sense to not open PRs until necessary backend work is done or to open them with mock data and then open subsequent PRs when the backend work is done? One disadvantage of the latter is that it could backload a bunch of PR work at the end of a project. I suppose it is a case by case thing.

I think it makes sense to still open the PRs here before the backend work is done, but maybe you could mention in the PR description whether a backend change is needed, whether there's a PR opened for the backend change, and also whether there are any steps you took locally to modify or mock out the data to actually develop and test the components locally.

Thanks again for all the info and for being patient!

richardxia:
> Thanks, I see that the PR is merged now, but when I try to rebuild the DB on the latest `master` branch, I'm hitting an error:

I managed to resolve this. It looks like I had some weird state locally due to having a DB table that was renamed, but where we just deleted the original migration instead of creating a migration to rename it. After wiping my local DB and ensuring I didn't have a modified schema.rb, everything is running fine for me now.

richardxia:
I just wanted to double check that we're still waiting on some changes to be made to the API side, right? I checked this branch out, and it looks like there's still a bit of a mismatch between the API resource's name and what client side is doing here. I tried changing the `patientHandouts` string here to `documents`, but the API still responds with:

```
NoMethodError (undefined method `each' for #<String:0x00007fe2b823c6c8>):
```

Let me know when this is ready for another round of review.

schroerbrian:
> I just wanted to double check that we're still waiting on some changes to be made to the API side, right? I checked this branch out, and it looks like there's still a bit of a mismatch between the API resource's name and what client side is doing here. I tried changing the `patientHandouts` string here to `documents`, but the API still responds with:
> 
> ```
> NoMethodError (undefined method `each' for #<String:0x00007fe2b823c6c8>):
> ```
> 
> Let me know when this is ready for another round of review.

The API changes are up as of last night (Tuesday, 27th). I have updated the frontend code to integrate with the `documents` and `instructions` endpoints. The new endpoints are not using the change request pattern; however, they seem to work quite nicely. 

I have pushed 3 new commits. The commits set up `postDocuments` and `postInstructions` functions in the `OrganizationEditPage` that handle POSTs, PUTs, and DELETEs for their corresponding objects. I also edited the logic for deleting collection member objects in the `removeItem` method in the  `EditServiceChildCollection` component, as the initial design wasn't working with the PDF/Documents. Lastly, I have made a few small var naming improvements. Thanks for your patience with this back and forth!!

richardxia (approved):
For some reason, when I check this out locally along with the latest `master` branch from askdarcel-api, the edit page appears to hang when I hit "Save Changes". However, the handout/document that I add does seem to get saved to the DB. By "hang," I mean that I can see two asynchronous HTTP requests to the `change_requests` endpoint that are stuck in the "Pending" state, and that never seem to get a response.

Actually, it might be a performance issue related to the Resource that I'm editing; I have a copy of the staging DB locally, and Resource 202 (Compass Family Services) seems to hang up even if I make a small change to something other than the new handouts/clinician actions fields.

Anyway, I'm going to hit Approve on this PR since the latest changes look good to me, but I wanted you to be aware of the edit page acting weird, at least for my on this one particular resource.

PR #1140, opened 2022-07-19
Ucsf flow updates

schroerbrian:
**Update: This has been rebased on top of the Node.js upgrade changes. I've tested it, and everything looks good.**

A PR that changes some UX related features. This includes: 

- Switching to radio buttons on the UCSF home page to allow only one checked resource at a time
- Changes to the eligibility selection logic including logic for checking "see all"
- General code refactors


t-petersen (approved):

schroerbrian:
@tpetersen0308 - can you add your review using your Sheltertech account? thanks!

tpetersen0308 (approved):
Sorry for the delay!

app/pages/UcsfClientEligibilityPage/UcsfClientEligibilityPage.tsx
```diff
@@ -1,56 +1,30 @@
...
+    name: string;
+  }
+
+  interface clientEligibilityGroup {
+    label: string;
+    eligibilities: clientEligibility[];
+  }
+
+  const resourceEligibilityGroup = rawEligibilityData[resourceSlug];
+  const [eligibilityGroupList, setEligibilityGroupList] = useState(resourceEligibilityGroup);
```
richardxia, line 27: What you're doing here is a little bit dangerous, since it has a risk of caching the state longer than you want. Specifically, if `resourceSlug` changes, the state won't be updated to contain the correct set of eligibilities. Out of sheer luck, as currently structured, the app can't run into that case here, since currently the `ClientEligibilities` component is guaranteed to be unmounted before being remounted with a new prop value. However, I think we should still fix this, since it could lead to some very weird bugs in the future if we change how this component is instantiated by its parent.

What you'll want to do is add a `useEffect()` call that triggers when the `resourceSlug` prop changes and that resets this eligibilityGroupList state. See https://bobbyhadz.com/blog/react-update-state-when-props-change for a simple example of this.

app/pages/UcsfClientEligibilityPage/UcsfClientEligibilityPage.tsx
```diff
@@ -66,30 +40,63 @@ const ClientEligibilities = () => {
...
+    if (targetEligibility.name === 'See all') {
+      updatedEligibilities = massToggleGroupEligibilities(
+        eligibilityGroup.eligibilities,
+        targetToggleState,
+      );
+    } else {
+      // If a user untoggles a specific eligiblity, this code also untoggles "See all"
+      // Currently, there is no logic to toggle "See all" if a user toggles all individual
+      // eligibilities given that this is somewhat of an implicit understanding; however, this
+      // can be added in the future
```
richardxia, line 57: If we get around to refactoring this in the future, one design we could consider is to not represent "See all" in the React state at all and instead model it as a function of the states of the other checkboxes. It's an application of the concept of "[Making illegal states unrepresentable](https://v5.chriskrycho.com/journal/making-illegal-states-unrepresentable-in-ts/)", where it would be impossible to represent the state of having the "See all" checkbox unchecked at the same time as when all the other checkboxes in the group are checked, and I think it would clean up this toggle handler function, since you wouldn't have to special case these edge cases for ensuring the state is consistent.

app/pages/UcsfClientEligibilityPage/UcsfClientEligibilityPage.tsx
```diff
@@ -66,30 +40,63 @@ const ClientEligibilities = () => {
...
+  );
+
   return (
     <div className={styles.eligibilitiesBox}>
       <div className={styles.eligibilitiesBox_title}>Client Identity</div>
       <ol className={styles.eligibilitiesLabels}>
 
         {/* Todo: This list rendering logic will be refactored when the API is setup */}
-        {eligibilityGroupList.map((eligibilityGroup, index) => (
+        {eligibilityGroupList.map((eligibilityGroup: clientEligibilityGroup, index: number) => (
```
richardxia, line 99: It's been a little while since I've written some significant TypeScript, but I feel like you should be able to elide these type annotations, since they should be inferable. Maybe the issue is that the initial declaration of `eligibilityGroupList` from the `useState()` call doesn't have enough info to allow TypeScript to infer its type?

app/pages/UcsfClientEligibilityPage/UcsfClientEligibilityPage.tsx
```diff
@@ -118,62 +125,45 @@ const ClientEligibilities = () => {
...
-    // Todo: We will need to alter the Service Discovery Results file to honor requests
-    // that can include more than one resource (e.g. substance abuse AND mental health)
-    // before successfully navigating to the suggested-resources page
-    alert('Navigate to target resources'); // eslint-disable-line no-alert
-    // history.push('/suggested-resources', {selectedResources: selectedResources});
+  interface LocationState {
+    selectedResourceSlug: string;
+  }
+
+  const state = history.location.state as LocationState;
```
richardxia, line 132: You shouldn't need to assert the type on this line if you pass in the state type as a type parameter to `useHistory()`:

```ts
const history = useHistory<LocationState>();
const state = history.location.state;
```

app/pages/UcsfClientEligibilityPage/ucsfEligibilities.ts
```diff
@@ -0,0 +1,101 @@
+// We are hardcoding the eligibility data as it currently exists on the design specs.
+// When content has entered UCSF specific eligibilities, will we fetch that data from the API
+
+export interface ucsfEligibilityDictionary {
```
richardxia, line 4:
```suggestion
export interface UcsfEligibilityDictionary {
```

Nit: Please use a capital initial letter for interfaces, since they are kind of like defining a class at the type level.

app/pages/UcsfClientEligibilityPage/ucsfEligibilities.ts
```diff
@@ -0,0 +1,101 @@
...
+  [key: string]: object;
```
richardxia, line 5: I think it would be valuable to give this a more precise type, since we know exactly the shape of the data we expect here, and adding a more precise type will allow TypeScript to enforce better type safety everywhere else in the app. Based on the data below, I think the following would be the right type definition:

```ts
interface Eligibility {
  name: string;
  checked: boolean;
}

interface EligibilityGroup {
  label: string;
  eligibilities: Eligibility[];
}

export interface UcsfEligibilityDictionary {
  [key: string]: EligibilityGroup[];
}
```

app/pages/UcsfServiceTypePage/UcsfServiceTypePage.module.scss
```diff
@@ -0,0 +1,80 @@
+@import '~styles/utils/_helpers.scss';
+
+.eligibilityPage {
```
richardxia, line 3: Nit: Having the word "eligibility" in all of these class names is a little bit confusing, since this is not the eligibilities page.

That said, it might be better to do a followup refactoring where we actually share common CSS/React components for the stuff that is the same between the different UCSF pages.

richardxia (approved):
Sorry to take a while to get to this, but overall, it looks pretty great. I think the major thing that I'd like to see fixed before we merge this is the potential issue with state that is derived from props, since that could lead to confusing issues in the future.

app/components/ui/inline/Button/Button.module.scss
```diff
@@ -7,6 +7,10 @@
   min-width: 120px;
   height: 56px;
   padding: 0 18px;
+
+  &:disabled {
+    cursor: not-allowed;
+  }
```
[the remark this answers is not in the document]
schroerbrian: Sure, great point. I have added an opacity on disabled state.

app/pages/UcsfClientEligibilityPage/UcsfClientEligibilityPage.module.scss
```diff
@@ -64,6 +64,7 @@
 
 .eligibilitiesBtns {
   text-align: right;
+  margin-bottom: 20px;
```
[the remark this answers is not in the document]
schroerbrian: Good call out and thanks for the link! I've moved spacing management to the parent ".grid" component using `padding`. (There is only one row within that component so I cannot use gap). The `eligibilitiesBtns` children were also setting a margin-left, so I moved that to use `gap` under the auspice of the `eligibilitiesBtns` class. BTW, is there a reason you prefer padding over margin? I suppose because the padding whitespace could technically be considered part of the element's box rather than as a space outside of the element?

app/pages/UcsfClientEligibilityPage/UcsfClientEligibilityPage.tsx
```diff
@@ -1,56 +1,30 @@
 import React, { useState } from 'react';
 import { useHistory } from 'react-router-dom';
 
-import { Checkbox } from 'components/ui/inline/Checkbox';
-import { Button } from 'components/ui/inline/Button';
-import { Section } from 'components/ucsf/Section';
-import { Layout } from 'components/ucsf/Layout';
+import { Checkbox } from 'components/ui/inline/Checkbox/Checkbox';
```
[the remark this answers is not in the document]
schroerbrian: I think I removed the re-exports while I was changing some import statements to play nicely with some our Node 18 and associated package updates – I can't remember if it was a linting issue or what that inspired this. In  making these changes I decided to also delete the `index.ts` files so as to remove the extra boilerplate. But I do see what you mean about the `index.ts` as a one time boilerplate vs once per import statement. That makes sense to me, and I'll follow that convention in the future.

app/pages/UcsfClientEligibilityPage/UcsfClientEligibilityPage.tsx
schroerbrian: Awesome thanks for explaining this subtle issue and providing the link. I have changed it per your suggestion.

app/pages/UcsfClientEligibilityPage/UcsfClientEligibilityPage.tsx
schroerbrian: Great, thanks for sharing that link. I bookmarked it to read during some down time and can refactor in the future.

app/pages/UcsfClientEligibilityPage/UcsfClientEligibilityPage.tsx
schroerbrian: Creating a more specific `UcsfEligibilityDictionary` interface, as suggested below, fixed this. Thanks!

app/pages/UcsfClientEligibilityPage/ucsfEligibilities.ts
schroerbrian: Great, this also fixes an issue with type inference that you called out above. Thanks!

schroerbrian:
> Sorry to take a while to get to this, but overall, it looks pretty great. I think the major thing that I'd like to see fixed before we merge this is the potential issue with state that is derived from props, since that could lead to confusing issues in the future.

Thanks as always for the careful review!

app/pages/UcsfClientEligibilityPage/UcsfClientEligibilityPage.module.scss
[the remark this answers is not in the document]
richardxia: > BTW, is there a reason you prefer padding over margin? I suppose because the padding whitespace could technically be considered part of the element's box rather than as a space outside of the element?

Yeah, that's basically it, but beyond it feeling "philosophically" right, there's a more pragmatic reason to favor padding over margin. When a component sets margin, it might make it more difficult for a parent component to make use of it, since the parent might want to position it in a way such that the child is closer to an element than the child's margin allows it to. It's always easy for a parent to add margin to a child (you can use [`>`](https://developer.mozilla.org/en-US/docs/Web/CSS/Child_combinator) to more narrowly select a specific child), but it's more difficult for a parent to undo the margin of a child. If you buy into the school of thought that the parent ought to control the layout of its internals, then a child element setting margin will interfere with that control. Padding isn't the only way that a parent element can set the spacing of its children; you can also use flexbox, `gap`, grid, or even using `first/last/nth-child` selectors to add margin to a specific child, but the point is that the parent should be the one in control.

app/pages/UcsfClientEligibilityPage/UcsfClientEligibilityPage.tsx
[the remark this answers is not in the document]
richardxia: Hmm, I see. I can try missing with the ESLint/TypeScript settings to see if we can avoid that.

That said, I'm not opposed to removing the `index.ts` files; I just wanted to make sure we have an agreed upon convention and that we make sure all of our files match that convention.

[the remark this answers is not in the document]
schroerbrian: I wish I could remember the exact issue that the linter was complaining about. I think it may have been something about using `default`

app/pages/UcsfClientEligibilityPage/UcsfClientEligibilityPage.module.scss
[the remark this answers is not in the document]
schroerbrian: Gotcha, thanks for the explanation!

PR #1141, opened 2022-07-20
Add QR code icon to search and service results pages. Move search and…

schroerbrian:
… service results header markup to component.

I started out adding the QR code functionality for UCSF and then ended up moving all the Service Results and Search Results page header UI into a standalone component since it is repeated across both pages. I will discuss with the product team and designers whether we want to include the QR code on SF Service Guide as well

The only changes to the header markup is adding the QR Code button to it. Everything else was just copied over.

There are also some small refactors on the Button component.

Do we include package-lock changes? I know that is standard, but I also know that can always be a little funky. Is there any policy we have regarding that? Thanks!

Screen Shot 2022-07-20 at 3 43 59 PM

Screen Shot 2022-07-20 at 3 44 05 PM


app/components/search/Header/Header.module.scss
```diff
@@ -0,0 +1,78 @@
+@import "~styles/utils/_helpers.scss";
```
schroerbrian, line 1: With the exception of the qr* classes, this styles in this file have been migrated  from: `app/components/search/SearchAndServiceResultsPage.module.scss` below

app/components/search/Header/Header.tsx
```diff
@@ -0,0 +1,49 @@
+import React, { useState } from 'react';
```
schroerbrian, line 1: This component was extracted from the searchResultsPage and serviceDiscoveryResults page below. The only changes involve the QR code modal

schroerbrian:
Closing. Updated PR is here: https://github.com/ShelterTechSF/askdarcel-web/pull/1147

PR #1146, opened 2022-07-29
Ucsf flow updates node upgrade fixes

tpetersen0308:
Fixes issues resulting from recent node upgrade.


schroerbrian:
Going to close this since it was a demonstration/reference PR that is no longer needed

PR #1147, opened 2022-08-01
Ucsf qr code copy

schroerbrian:


app/components/search/Header/Header.tsx
```diff
@@ -0,0 +1,47 @@
+import React, { useState } from 'react';
```
schroerbrian, line 1: This component was extracted from the searchResultsPage and serviceDiscoveryResults page below. The only new changes involve the addition of the QR code modal

tpetersen0308 (approved):
LGTM!

PR #1149, opened 2022-08-05
Service discovery form custom steps

schroerbrian:
This add the Long Term Housing tile to our homepage pathways. The LTH pathway can have two steps beyond what our current setup allows for. One of the possible steps actually takes the user to the Shelter discovery form. I added some code in the ServiceDiscoveryForm that is specific to the LTH tile to account for these extra possibilities. I was hoping to create something more general, but given time constraints for our demo, I didn't end up doing it. I'm very open to suggestions for improving this code in general. That said, HSH/MOHCD may want us to get this tile up sooner rather than later. Even if we can add some TODOs that'd be great.

This document has the possible decision paths when navigating the LTH pathway: https://docs.google.com/document/d/17NKgCHDP2KHH0eI7rZ2-cLdlSjT_Y7U7O3ftnW_Rzz8/edit


app/pages/ServiceDiscoveryForm/ServiceDiscoveryForm.tsx
```diff
@@ -4,63 +4,110 @@ import qs from 'qs';
...
+  const disableNextBtn = selectedRadioItem === null && ['housingStatus', 'subcategoriesRadio'].includes(stepName);
   // TODO: Should goBack go back to the previous step?
-  const history = useHistory();
-  const goBack = () => history.goBack();
-  const goToNextStep = () => setCurrentStep(currentStep + 1);
+  const goBack = () => {
+    history.push('/');
+  };
+
+  if (stepName === 'housingStatus') {
```
schroerbrian, line 82: I tried to figure out a more abstract solution than adding Long Term Housing specific conditions in the service discovery form, but I didn't feel like I had the time given our demo deadline.  I think that if we reorganize how we structure our category/steps data in the constants.ts file it might allow for more flexibility here, though that would of course include a larger refactor of this module.

app/pages/ServiceDiscoveryForm/ServiceDiscoveryForm.tsx
```diff
@@ -114,25 +165,48 @@ const Content = ({
     );
   };
 
+  const handleRadioSelect = (targetCategoryId: number) => {
+    setSelectedSubcategories({ [targetCategoryId]: true });
+    setSelectedRadioItem(targetCategoryId);
+  };
+
   switch (steps[currentStep]) {
+    case 'housingStatus':
```
schroerbrian, line 174: See my above comment regarding LTH specific conditions in this file

app/pages/ServiceDiscoveryForm/ServiceDiscoveryForm.tsx
```diff
@@ -114,25 +165,48 @@ const Content = ({
...
-          selectedOptions={selectedSubcategories}
-          toggleOption={handleSubcategoryClick}
+          refinements={subcategories}
+          selectedRefinements={selectedSubcategories}
+          toggleRefinement={handleSubcategoryClick}
+        />
+      );
+    case 'subcategoriesRadio':
+      return (
+        <RadioFormStep
```
schroerbrian, line 205: I think at least a few of our current pathways should be using a radio form instead of checkboxes as we only want users to select on refinement at a time in some cases.

richardxia (commented):
Do I need to run some commands/create some fake data in order to check this out locally? When I check out this branch and try to visit the Long-Term Housing page, I don't see any options to select.

app/pages/ServiceDiscoveryForm/ServiceDiscoveryForm.tsx
```diff
@@ -4,63 +4,110 @@ import qs from 'qs';
...
   id: number;
 }
 
-interface SelectedResources {
+interface SelectedRefinements {
   [key: number]: boolean;
 }
 
 /** Wrapper component that handles state management, URL parsing, and external API requests. */
 export const ServiceDiscoveryForm = () => {
```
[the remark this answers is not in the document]
schroerbrian: My bad – I saw the `step` state vars in the inner component and lazily figured the wrapper component wasn't really being used as such anymore. I only moved state vars/API requests down to lessen the number of args being passed down. But now I've moved them back to the wrapper component.

schroerbrian:
> Do I need to run some commands/create some fake data in order to check this out locally? When I check out this branch and try to visit the Long-Term Housing page, I don't see any options to select.

@jjfreund will get a PR up to add the SQL to our dev environment, but for now you will have to pull down the latest copy of Staging (as of Aug 18th ~5pm there are essential data updates that fix bugs from previous data deploys)

app/pages/ServiceDiscoveryForm/ServiceDiscoveryForm.tsx
```diff
@@ -4,63 +4,117 @@ import qs from 'qs';
...
+    steps,
+    slug: categorySlug,
+  } = category;
 
-  // TODO: Should goBack go back to the previous step?
+  const [currentStep, setCurrentStep] = useState(0);
   const history = useHistory();
-  const goBack = () => history.goBack();
-  const goToNextStep = () => setCurrentStep(currentStep + 1);
+  let goToNextStep;
```
richardxia, line 85: Apologies if I made this comment in some older PR, but the code might be a bit easier to follow if moved the `if (stepName === 'housingStatus')` conditional into the definition of `goToNextStep` rather than conditionally assigning two different functions. E.g.:

```js
const goToNextStep = () => {
  if (stepName === 'housingStatus') {
    // ...
  } else {
    // ...
  }
}
```

app/pages/ServiceDiscoveryForm/ServiceDiscoveryForm.tsx
```diff
@@ -4,63 +4,117 @@ import qs from 'qs';
...
-  const eligibilities: CategoryRefinement[] = useEligibilitiesForCategory(category.id) || [];
-  const subcategories: CategoryRefinement[] = useSubcategoriesForCategory(category.id) || [];
+  // The activeCategoryId is updated if the user proceeds to a further step that has child
+  // subcategories to be displayed. When it is set, the target subcategory refinements are
+  // fetched and rendered in place of the previous refinements
+  const [activeCategoryId, setActiveCategoryId] = useState<string | null>(category.id);
+
+  // Add the selectedRadioItem to state so that when a user clicks the next button
+  // the selectedRadioItem can be passed to the goToNextStep function
+  const [selectedRadioItem, setSelectedRadioItem] = useState<number | null>(null);
```
richardxia, line 45: Not sure if it should go here or somewhere else, but I think a comment somewhere stating that the radio item values are going to be the subcategory ID would be useful, since it's not completely obvious from the code alone.

app/pages/ServiceDiscoveryForm/ServiceDiscoveryForm.tsx
```diff
@@ -4,63 +4,117 @@ import qs from 'qs';
...
+        return;
+      }
+
+      if (selectedRadioItem === 1100045) {
+        // User has selected first option in Long Term Housing step 1. Set category ID to Shelter
+        // ID and redirect user to the Shelter resources form.
+        setActiveCategoryId('1000010');
+        history.replace('/shelter-resources/form');
+      } else {
+        const targetCategoryId = (selectedRadioItem);
```
richardxia, line 107:
```suggestion
        const targetCategoryId = selectedRadioItem;
```

Nit: No parentheses needed here.

app/pages/ServiceDiscoveryForm/ServiceDiscoveryForm.tsx
```diff
@@ -185,29 +266,35 @@ const Header = ({ onGoBack }: {
...
   <div className={styles.footer}>
     <div className={styles.progressBarContainer}>
       {
         /*
-         * Add 1 to current step because it is 0-indexed.
-         * Subtract 1 from numSteps because we shouldn't include the RESULT step.
-         */
+        * Add 1 to current step because it is 0-indexed.
+        * Subtract 1 from numSteps because we shouldn't include the RESULT step.
+        */
```
richardxia, line 283: Revert accidental whitespace changes?

app/pages/ServiceDiscoveryForm/ServiceDiscoveryForm.module.scss
```diff
@@ -111,16 +138,38 @@
...
+
     .actionBack {
         background-color: transparent;
         color: $color-brand;
     }
+
     .actionSubmit {
         background-color: $color-brand;
+        @media screen and (max-width: $break-tablet-s) {
+            margin-bottom: 8px;
```
richardxia, line 167: This might be a bit too small, since it could lead to users accidentally hitting one button when they meant to hit a different one. From a quick Google search, I found one [blog post](https://uxmovement.com/mobile/optimal-size-and-spacing-for-mobile-buttons/) that recommends between 12–48px spacing, depending on the size of the buttons. I'm not sure if their recommendations for circular buttons translates completely over to our rectangular ones, but I think we could at least bump up the spacing a bit.

richardxia (approved):
Sorry for the delayed review, but this LGTM! I managed to dump the staging DB and load it locally, so I was able to check out these new pages locally. I think these are reasonable steps to take in the short term, before we have a plan for the fully generalized branching question form.

One minor thing that I noticed, which I don't think there's a good fix for, is that when you select the first option, "
I am experiencing homelessness and I need immediate help finding shelter," when you get invisibly moved into the existing shelter resources form, the progress bar disappears, since you're no longer in the long-term housing form. I don't think there's a good way to fix this right now, but I just wanted to point it out.

app/pages/ServiceDiscoveryForm/ServiceDiscoveryForm.module.scss
schroerbrian: Thanks for the link. This is great research. I moved the spacing to the parent element `.actionGroup` using `gap` per your great suggestion in a previous PR. I then set the value to 14px. Reading this article makes me think we may even want to bump up the height of these buttons. But I didn't want to change that unilaterally.

app/pages/ServiceDiscoveryForm/ServiceDiscoveryForm.tsx
schroerbrian: Yes, good suggestion, thanks! I had been considering renaming those vars to `selectedSubcategoryId` but thought that could be a little confusing in relation to the `selectedSubcategories` state var, and that it may not always represent a subcategoryId if the radio component is ever used for eligibilities. 

However, I've gone ahead and updated the state var name to `selectedSubcategoryId` anyhow, since I think it is clearer. When we do a larger refactor of this process, we can consider updating this further

app/pages/ServiceDiscoveryForm/ServiceDiscoveryForm.tsx
schroerbrian: Oh yes, excellent suggestion. Thank you!

schroerbrian:
> Sorry for the delayed review, but this LGTM! I managed to dump the staging DB and load it locally, so I was able to check out these new pages locally. I think these are reasonable steps to take in the short term, before we have a plan for the fully generalized branching question form.
> 
> One minor thing that I noticed, which I don't think there's a good fix for, is that when you select the first option, " I am experiencing homelessness and I need immediate help finding shelter," when you get invisibly moved into the existing shelter resources form, the progress bar disappears, since you're no longer in the long-term housing form. I don't think there's a good way to fix this right now, but I just wanted to point it out.

Yes this is a great point. I cant think of an easy way to do this either. I made a note to include this in the larger refactor when we get around to more generalized branching. I was also trying to rig up a way to make the back button go one step back but ran into some weird states when nav'ing back from the Shelter options back to the LTH options. That would be nice to include as well though when we have more time. 

Thanks again for all the helpful suggestions!

PR #1150, opened 2022-08-10
Show banner with monkey pox information.

schroerbrian:
Using these designs: https://www.figma.com/file/YB6UoGIXIms5j2lFQei8hv/Banner%2FBulletin-Board?node-id=614%3A4179

image


app/components/ui/Banner.tsx
```diff
@@ -1,23 +1,17 @@
...
-    {' '}
-    Check out
-    {' '}
-    <a target="_blank" rel="noreferrer" href="https://docs.google.com/document/d/1t-RM30QQHA5UQdzIGHrD6kaFoBMrwmCBM7mEOBD8a9U/edit">Code Tenderloin&apos;s workshops</a>
-    {' '}
-    and
-    {' '}
-    <a target="_blank" rel="noreferrer" href="https://www.homeownershipsf.org/application-assistance-for-renters/">HomeownershipSF&apos;s resources</a>
-    .
+    <div className={styles.bannerContent}>
```
richardxia, line 7: What's the purpose of this inner div? The CSS style doesn't exist, either. If it's not needed, can we remove this layer and have the rest of the content directly under the `bannerContainer`?

richardxia (approved):
A few comments about the CSS/layout:

In the Figma design, the left edge of the alert icon is supposed to line up with the main content on the page, as shown by the vertical red line in the design:
Screen Shot 2022-08-10 at 12 10 41 PM

When I view the page, the banner content hangs noticeably to the left of the main content, though it may be dependent on the browser width. Could we make the banner content line up with the rest of the site content?

I've also noticed that the icon is slightly lower vertically than where it is in the Figma designs. This is because vertically aligning inline content [is actually a pretty tricky topic](https://christopheraue.net/design/vertical-align), and in this scenario, your image and the text next to it are part of the same inline flow.

One way to work around this is to create a grid that wraps the banner content and that defines two columns: one for just the icon, and one for the rest of the (inline) content. You can then `align-items: center` the whole grid, and it'll look a bit better.

---

These are both non-blocking concerns, so I think it's fine to merge and clean up the CSS afterwards.

app/components/ui/Banner.tsx
schroerbrian: I ended up using it to align the content with the homepage tiles per Dongyeon's designs. Currently we set our tile layout by setting a percentage width and a max-width. I just copied those properties onto the banner for now, rather than create my own kind of grid layout that would align with the homepage. I think I/we could go back and fix it up better, but I didn't want to sink too much more time into it.

schroerbrian:
> A few comments about the CSS/layout:
> 
> In the Figma design, the left edge of the alert icon is supposed to line up with the main content on the page, as shown by the vertical red line in the design: Screen Shot 2022-08-10 at 12 10 41 PM
> 
> When I view the page, the banner content hangs noticeably to the left of the main content, though it may be dependent on the browser width. Could we make the banner content line up with the rest of the site content?
> 
> I've also noticed that the icon is slightly lower vertically than where it is in the Figma designs. This is because vertically aligning inline content [is actually a pretty tricky topic](https://christopheraue.net/design/vertical-align), and in this scenario, your image and the text next to it are part of the same inline flow.
> 
> One way to work around this is to create a grid that wraps the banner content and that defines two columns: one for just the icon, and one for the rest of the (inline) content. You can then `align-items: center` the whole grid, and it'll look a bit better.
> 
> These are both non-blocking concerns, so I think it's fine to merge and clean up the CSS afterwards.

Thanks for the helpful feedback. I made the banner align with the home page content (see my comment above). 

I think Dongyeon designs for screens that are fairly large. I ended up shortening "Information" to "Info" but had I not the text would have wrapped around to another line on my 14 inch laptop using the style constraints that our homepage is using. I know he did have to dash off this design quickly as he is techincally on vacation. 

I think we can go back and fix up the CSS and styling when we work on implementing a banner that can be dynamically shown/hidden by the content team. I'm going to go ahead and merge this work in now though, just so that I can do an early deploy tomorrow. Thanks again!

richardxia:
> I think we can go back and fix up the CSS and styling when we work on implementing a banner that can be dynamically shown/hidden by the content team. I'm going to go ahead and merge this work in now though, just so that I can do an early deploy tomorrow. Thanks again!

Yep, sounds good, and thanks for addressing the alignment with a quick fix. I'm definitely down to help out with some of styling, since I've been itching to do a lot of cleanup of our CSS in this repo!

PR #1153, opened 2022-08-14
Remove old COVID pages.

richardxia:
These weren't even reachable outside of the mobile site's hamburger menu
navigation, and these are pretty out of date and no longer relevant to
current events.

If you aren't familiar with these pages, they're available at `/covid`, and they're something we put up very quickly after the pandemic started, when we realized that most of the information in our database would be out of date due to the sudden lockdowns. Here's a screenshot of the page:

Screen Shot 2022-08-13 at 6 06 49 PM


schroerbrian (approved):
Looks great. Thank you for cleaning this up.

richardxia:
Thanks for the review!

PR #1155, opened 2022-08-26
Clinician actions modal

schroerbrian:
UPDATE: I added some more work to this PR. I created a parent Modal component to standardize the modal styles that we're using on the UCSF service results page – the QR Code modal, Clinician Actions modal, and the PDF Handouts modal. I began adding it to the Text Me modal, but that is slightly more complicated. I will continue working on that while this PR is up and will hopefully get it in before review is done. If not, I think it can be its own self-contained PR soon. 

I also altered some styles on the QRCode modal to reflect changes that the designers made (to standardize it against the other modals) after it was dev complete.

Designs: https://www.figma.com/file/HUTUyVuJf8lctj5fAFIY32/UCSF?node-id=5470%3A7112

Screen Shot 2022-08-26 at 12 12 37 PM

Screen Shot 2022-08-26 at 12 12 20 PM

image


app/components/search/SearchResults/SearchResults.jsx
```diff
@@ -194,9 +161,13 @@ const SearchResult = ({ hit, index, setCenterCoords }) => {
           <Link to={`/organizations/${resourceId}`}>{hit.service_of}</Link>
         </div>
         <div className={styles.address}>{renderAddressMetadata(hit)}</div>
-        <ReactMarkdown className={`rendered-markdown ${styles.description}`} source={hit.long_description} />
+        <ReactMarkdown className={`rendered-markdown ${styles.description}`} source={hit.long_description} linkTarget="_blank" />
```
schroerbrian, line 164: Adding this linkTarget attr on our markdown editor component will instruct links to open in new tabs. This was a product request from Tammy, which I think makes a lot of sense.

app/components/search/SearchResults/SearchResults.jsx
```diff
@@ -102,58 +114,6 @@ const SearchResult = ({ hit, index, setCenterCoords }) => {
     return <span>No address found</span>;
   };
 
-  // Todo: to be included with next stage of multiple location work
```
schroerbrian, line 105: It remains to be seen if we will go back to our multiple location work or not. The codebase has already changed quite a bit since I wrote this comment, so it is probably of less relevance. And if I or anyone else wants to consult it, it will be in our history.

app/components/listing/TableOfContactInfo.tsx
```diff
@@ -34,7 +34,7 @@ export const TableOfContactInfo = ({ service }: { service: Service }) => {
             <tr>
               <th>Website</th>
               <td>
-                <a href={website}>{website}</a>
+                <a target="_blank" rel="noopener noreferrer" href={website}>{website}</a>
```
schroerbrian, line 37: As a bonus to this work, I added target="_blank" attrs to anchor/link elements on our services that go to external sites per a product request from Tammy. I think it makes a lot of sense to open a new tab when clicking to a new site

app/components/search/SearchResults/SearchResults.jsx
```diff
@@ -194,7 +161,7 @@ const SearchResult = ({ hit, index, setCenterCoords }) => {
           <Link to={`/organizations/${resourceId}`}>{hit.service_of}</Link>
         </div>
         <div className={styles.address}>{renderAddressMetadata(hit)}</div>
-        <ReactMarkdown className={`rendered-markdown ${styles.description}`} source={hit.long_description} />
+        <ReactMarkdown className={`rendered-markdown ${styles.description}`} source={hit.long_description} linkTarget="_blank" />
```
schroerbrian, line 164: Adding the linkTarget="_blank" attr to the ReactMarkdown component adds target="_blank" to all links in the markdown; this is for the service description

app/components/ui/QrCodeModal/QrCodeModal.module.scss
```diff
@@ -34,8 +34,6 @@
   position: absolute;
   right: 30px;
   top: 30px;
-  height: 12px;
-  width: 12px;
```
schroerbrian, line 38: I removed sizing of the close icon on the QR Code Modal to align with the sizing on existing modals in our code base. In discussions with Dongyeon for this ClinicianAction modal, we decided to use the existing "close" icons for our modals rather than his new designs. Thus I changed it here as well for consistency.

schroerbrian:
Changing to draft while I add some more work to this PR

app/components/ucsf/ClinicianActions/ClinicianActions.module.scss
```diff
@@ -0,0 +1,50 @@
+@import '~styles/utils/_helpers.scss';
```
schroerbrian, line 1: see parent component Modal for general styles

app/components/ui/Modal/Modal.tsx
```diff
@@ -0,0 +1,31 @@
...
+}) => (
+  <ReactModal
+    isOpen={isOpen}
+    className={`${styles.modal} ${addModalClass}`}
+    overlayClassName={styles.modalOverlay}
+    onRequestClose={closeModal}
+  >
+    <div role="button" tabIndex={0} onClick={closeModal}>
+      <img src={icon('close')} alt="close" className={styles.closeModalIcon} />
+    </div>
```
richardxia, line 28: Why not use an actual `<button>` instead of a `<div>` that mimics a button? I'm guessing you copied this from the existing modals on the app, so we probably don't need to address this here, and instead we can probably take of this in a general cleanup of all the modals on the site to use a single set of base styles.

app/utils/whitelabel.ts
```diff
@@ -66,6 +69,9 @@ const SFFamilies: WhiteLabelSite = {
   logoLinkDestination: 'https://www.sffamilies.org/',
   navLogoStyle: styles.navLogoSFFamilies,
   showBanner: false,
+  showClinicianAction: false,
```
richardxia, line 72: I'm starting to think that we should have a base object with default values that we clone and modify for each of the individual sites, since it makes it difficult to see what the actual differences are between the sites when 90% of them are the same.

Just broke out https://github.com/ShelterTechSF/askdarcel-web/issues/1164 to capture this as a followup ticket.

app/components/ui/QrCodeModal/QrCodeModal.tsx
```diff
@@ -20,34 +19,28 @@ export const QrCodeModal = ({
...
-          tabIndex={0}
-        >
-          CLOSE
-        </Button>
+      <div className={styles.modalContent}>
+        <div className={styles.description}>
+          <h2 className={styles.title}>Resource Code</h2>
+          <p className={styles.body}>
+            Scan this QR code with a mobile phone
+            <br />
```
richardxia, line 30: Is this necessary? It may be difficult in general to ensure that this breaks at a reasonable place when using different font sizes (remember, the user may want to increase the site font for accessibility) and if the text changes in the future.

When I removed this `<br />` tag locally, it seemed to put a line break at a reasonable place to me.

If we're trying to avoid a line that just has a single word on it, one trick that I learned a long time ago is to put a non-breaking space (`&nbsp;`) between the last and second to last words, which will ensure that the final line will have at least two words.

app/components/ucsf/ClinicianActions/ClinicianActions.module.scss
```diff
@@ -0,0 +1,50 @@
...
+
+.clinicianActions {
+  position: relative;
```
richardxia, line 4: This is already being set by the base Modal, right? Could we get rid of this? I think it's in general tricky to keep track of relative positioning in CSS when split across a base CSS class and an extended one, and it'll be easier to maintain if only the base CSS class is in charge of both this element's relative positioning as well as anything else that relies on that relative positioning.

I don't see you relying on relative positioning for any of the styles in this file, since I don't see anything with an explicit top/right/bottom/left attribute, z-index, or any of the obvious ones, so I think this component can be agnostic of the relativeness of the parent modal.

app/components/ucsf/ClinicianActions/ClinicianActions.module.scss
```diff
@@ -0,0 +1,50 @@
...
+.buttonBar {
+  display: flex;
+  justify-content: end;
+}
+
+.closeBtn {
+  width: 128px;
+  height: 53px;
+  border-radius: 3px;
+  font-size: 16px;
```
richardxia, line 49: Do we actually need to set all these properties? From a design perspective, it would be better if the buttons on our site were as consistent as possible and that we stick to as few unique styles as possible.

Regarding width specifically, it looks like our default button styles already choose 120px as the default width. I'm not sure it's worth overriding the width when we have an existing width that is so close. For height, we have a similar thing where the existing button styles set 56px as the default, so I don't know if overriding it to 53px is necessary.

In addition, for height, we are kind of already redundantly setting it in the base button component, since the height is redundant with (top padding + bottom padding + line height * font-size), and setting both sets of properties makes it less clear what the desired final height should be. That said, this particular issue affects the base button component and should probably be resolved in a separate PR. Just created https://github.com/ShelterTechSF/askdarcel-web/issues/1165 to track this.

app/components/ui/ClientHandoutsModal/ClientHandouts.tsx
```diff
@@ -0,0 +1,52 @@
...
+          </a>
+        ))}
+
+        <Button
+          onClick={closeModal}
+          addClass={styles.closeBtn}
+          tabIndex={0}
+        >
+          Close
+        </Button>
```
richardxia, line 46: For consistency with the Clinician Actions modal, do you think it'd be a good idea to wrap this in a `.buttonBar` element?

richardxia (approved):
LGTM, and sorry for taking such a long time to review this. I've broken out a few other tickets to keep track of miscellaneous and minor refactoring tasks, but overall this PR LGTM!

app/components/listing/TableOfContactInfo.tsx
schroerbrian: Good point and thanks for sharing these resources! I try adding the `RiExternalLinkFill`. It looked a little odd with links that took up more than one line. I made a note to talk to the designers during the next Wednesday meeting to get an idea of what they think of opening in a new tab or not or if they want to provide a design

app/components/ui/ClientHandoutsModal/ClientHandouts.tsx
schroerbrian: Good idea. I have added it and now control positioning via the `.buttonBar`

app/components/ucsf/ClinicianActions/ClinicianActions.module.scss
schroerbrian: I've gone ahead and removed the width/height as well as `border-radius` and `font-size`. I have made a note to bring up button standardization during the next design meeting or two

app/components/ucsf/ClinicianActions/ClinicianActions.module.scss
schroerbrian: Nope, no need! I think it was left over before I switched over to using the parent component.

app/components/ui/QrCodeModal/QrCodeModal.tsx
schroerbrian: Thanks, good trick! Done.

app/utils/whitelabel.ts
schroerbrian: Thank you. Great idea!

PR #1158, opened 2022-09-01
Printouts modal

schroerbrian:
This also includes a parent modal component that I created to standardize modal styles across the search results page – the QRCode modal, the Clinician Actions modal, and the Handouts modal. I also began adding it to the Text Me modal, but that is slightly more complicated. I will continue working on that while this PR is up and will hopefully get it in before review is done. If not, I think it can be its own self-contained PR soon.

I also altered some styles on the QRCode modal to reflect changes that the designers made to standardize modal styles as well as to introduce improved patterns that I've learned since completing that work.


schroerbrian:
Dupe of https://github.com/ShelterTechSF/askdarcel-web/pull/1155 after synthesizing changes

PR #1160, opened 2022-09-06
Add textarea to edit page for adding clinician/covid actions to service.

schroerbrian:


schroerbrian (commented):
Small change

richardxia (approved):
This LGTM, but is there a backend change needed for this to work?

schroerbrian:
> This LGTM, but is there a backend change needed for this to work?

Actually there was a slight discrepancy between what the f/e was expecting and what the b/e is passing back. So thank you for the callout. The b/e is passing back an array whereas I was expecting a string. I updated the logic to display to the instruction prop of the first element of an array, and to edit that same value on change. There should really only be one instructions per service. I'm going to bring this up with the b/e team. We might need to refactor this in the future, but I'll just merge in for now. 

Joe is still working on the CRUD endpoint for updating instructions, but that will be done any day now.

richardxia:
> Actually there was a slight discrepancy between what the f/e was expecting and what the b/e is passing back. So thank you for the callout. The b/e is passing back an array whereas I was expecting a string. I updated the logic to display to the instruction prop of the first element of an array, and to edit that same value on change. There should really only be one instructions per service. I'm going to bring this up with the b/e team. We might need to refactor this in the future, but I'll just merge in for now.

Oh nice, I had checked out the code locally, tried adding a Clinician action and saw that it failed, but I didn't realize that it was because there was a mismatch in API expectations. I assumed that it just wasn't implemented at all in the backend. Nice job catching it and making a quick fix/workaround here.

Issue #640, opened 2019-02-25
Notes hyperlinks in services


Maxastuart:
Make hyperlinks display & be useable on Organization page when they’re added to individual Service “notes” sectiosn. Example: https://staging.askdarcel.org/services/260

PR #1162, opened 2022-09-12
Ucsf subcategory eligibility endpoints

schroerbrian:
This PR includes several changes: 
1. It adds a request to the subcategory endpoint to fetch a UCSF service's subcategories to present as refinements on the `ServiceType` page rather than using mock data
2. It passes these subcategory refinements to the Algolia query generated when visiting the /results page
3. It still uses the `ucsfEligibilities` (now `ucsfEligibilitiesMap`) file as a source of truth for UCSF service eligibilities, given current API limitations. It adds a slight modification to this file to include a checked property, which maps an eligibility's checked state on the `ClientEligibility` page
4. It passes checked eligibilities to the Algolia query on the /results page as eligibility refinements 
5. Lastly, it refactors the code on the `ClientEligibility` and `ServiceType` pages to more closely follow how the `ServiceDiscoveryForm` page manages refinements, which is to use a separate object of refinement ID keys to boolean values. I think this improves code clarity. It also reduced total LOCs
6. It adds a Google Analytics event trigger on the `ServiceType` page, which follows the example of the event triggered on the `ServiceDiscoveryForm`
7. It makes a change to the `refinementMapping` file given that we will be updating a couple eligibilities that are referenced in that file

This branch should work with the most recent updates on the API Master branch `13c8c578f17274b59b9421cb6d60102b237a668c` commit message `Include documents in service presenter (#640)`

I used a copy of the Staging DB for development. Since there are currently no tagged UCSF services, you will not see any results. I tagged one service locally with UCSF category, subcategories, and eligibilities and was able to see that in my results.


app/pages/ServiceDiscoveryForm/constants.ts
```diff
@@ -93,36 +93,34 @@ export const CATEGORIES: Readonly<ServiceCategory[]> = [
     subcategorySubheading: 'If you need shelter, then tell us more about who you are. Select one answer.',
   },
   {
-    algoliaCategoryName: 'Covid-longterm-housing',
+    algoliaCategoryName: 'Covid-longtermhousing',
```
schroerbrian, line 96: Oops. This is the correct category name as defined in the DB

app/utils/refinementMappings.js
```diff
@@ -3,7 +3,7 @@ export const eligibilitiesMapping = {
   Families: ['Families', 'Families with Babies'],
   Homeless: ['Homeless'],
   Immigrants: ['Immigrants', 'Undocumented'],
-  LGBTQ: ['LGBTQ', 'Transgender'],
```
schroerbrian, line 6: `Transgender` is not currently an eligibility in the DB, so I've removed it. I've replaced it with `Transgender and Gender Non-Conforming`, which does exist in the DB. As part of this work, we plan to change the `LGBTQ` eligibility to `LGBTQ+`, so I've added that here

app/pages/UcsfClientEligibilityPage/UcsfClientEligibilityPage.tsx
```diff
@@ -7,108 +7,95 @@ import { Section } from 'components/ucsf/Section/Section';
...
-      updatedEligibilities = massToggleGroupEligibilities(
-        eligibilityGroup.eligibilities,
-        targetToggleState,
-      );
+  const handleEligibilityClick = (
+    eligibility: Eligibility,
+    eligibilities: Eligibility[],
+  ) => {
+    const seeAllEligibilityItem = eligibilities.find(e => e.isSeeAll);
+    const eligibilityCheckedId = eligibility.checkedId;
```
richardxia, line 41: Nit: Is this variable providing any value over just writing out `eligibility.checkedId` in the places where it is used? The variable's name is basically the exact same thing, so it's neither more expressive nor more concise.

app/pages/UcsfClientEligibilityPage/UcsfClientEligibilityPage.tsx
```diff
@@ -7,108 +7,95 @@ import { Section } from 'components/ucsf/Section/Section';
...
+    const targetValue = !selectedEligibilities[eligibilityCheckedId];
```
richardxia, line 42: Nit: Maybe `newValue` would be a clearer name for this variable? "Target" has specific meaning when it comes to HTML forms and events, so I thought it was related to that at first, but I think this is just what the new value of the checkbox state will be.

app/pages/UcsfClientEligibilityPage/UcsfClientEligibilityPage.tsx
```diff
@@ -7,108 +7,95 @@ import { Section } from 'components/ucsf/Section/Section';
...
-        { ...seeAllEligibility },
-        ...eligibilityGroup.eligibilities.slice(1, eligibilityIndex),
-        updatedEligibility,
-        ...eligibilityGroup.eligibilities.slice(eligibilityIndex + 1),
-      ];
-    }
+      // If target checked value is false, uncheck "See All" box as well
+      if (!targetValue) {
+        // N.B. Use of "!" assumes that every Eligibility array will have a "See All" element
+        updatedEligibilities[seeAllEligibilityItem!.checkedId] = false;
```
richardxia, line 56: I think it might be clearer to do the not-null assertion when you first assign the variable, since it would be right next to where you fetched the eligibility item in the first place and if in the future we wanted to use `seeAllEligibilityItem` in multiple places, it would also avoid having to do the not-null assertion at every place where you  use the variable.

app/pages/UcsfClientEligibilityPage/UcsfClientEligibilityPage.tsx
```diff
@@ -7,108 +7,95 @@ import { Section } from 'components/ucsf/Section/Section';
...
                   <Checkbox
-                    onChange={() => setEligibilityGroup(index, toggleChecked(eligibilityGroup, i))}
+                    onChange={() => handleEligibilityClick(
+                      eligibility,
+                      eligibilityGroup.eligibilities,
+                    )}
                     name={eligibilityGroup.label}
                     id={`${eligibilityGroup.label}-${eligibility.name}`}
-                    checked={eligibility.checked}
+                    checked={selectedEligibilities[eligibility.checkedId] || false}
```
richardxia, line 98:
```suggestion
                    checked={selectedEligibilities[eligibility.checkedId] ?? false}
```

I forget if we had issues with tooling, but it'd be better to use `??` over `||` here, since it more explicitly indicates that you just want a fallback for the case where the key doesn't exist, in which case an `undefined` will be returned.

app/pages/UcsfClientEligibilityPage/UcsfClientEligibilityPage.tsx
```diff
@@ -124,23 +111,42 @@ const ClientEligibilities = ({ rawEligibilityData, resourceSlug }: {
...
   }
 
   const history = useHistory<LocationState>();
   const { state } = history.location;
   const selectedResourceSlug = state && state.selectedResourceSlug;
+  const resourceEligibilityGroups = eligibilityMap[selectedResourceSlug];
+
   const goToServiceTypePage = (slug: string) => {
-    history.push('/service-type', { selectedResourceSlug: slug });
+    const flattenedEligibilityGroups = resourceEligibilityGroups.reduce<Eligibility[]>(
```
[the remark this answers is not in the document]
richardxia: Also, I think you might just want to call the [`flat()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flat) method defined on arrays.

app/pages/UcsfClientEligibilityPage/UcsfClientEligibilityPage.tsx
```diff
@@ -124,23 +111,42 @@ const ClientEligibilities = ({ rawEligibilityData, resourceSlug }: {
...
+
+    const selectedEligibilityNames = flattenedEligibilityGroups.reduce<string[]>((
+      result,
+      eligibility,
+    ) => {
+      if (selectedEligibilities[eligibility.checkedId] && !eligibility.isSeeAll) {
+        return [...result, eligibility.name];
+      }
+      return result;
+    }, []);
```
richardxia, line 140: Similar to my `flat()` suggestion, could this be expressed with `filter()`?

richardxia (commented):
Haven't done a full review of all the code yet, but I thought I'd leave my comments so far.

app/pages/UcsfClientEligibilityPage/UcsfClientEligibilityPage.tsx
schroerbrian: Since I only want to return the eligibility names, using filter would have required using map as well (I think). But I think I found a better solution in the `flatMap` method. See my comment above

app/pages/UcsfClientEligibilityPage/UcsfClientEligibilityPage.tsx
schroerbrian: Good call. For now, since that var exists only within the fairly small scope of this function, I have just added a check in the `if` condition right above line 55 to ensure that the `seeAllEligibilityItem` exists before accessing the `checkedId` prop. I could see a case where it may not exist in some eligibility groups in the future.

app/pages/UcsfClientEligibilityPage/UcsfClientEligibilityPage.tsx
schroerbrian: Thank you, I agree. I tend to overuse "target*" in general :D

app/pages/UcsfClientEligibilityPage/UcsfClientEligibilityPage.tsx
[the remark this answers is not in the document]
richardxia: Ah, I see, I hadn't looked closely at the actual types involved, but this makes sense. It does look cleaner with `flapMap()`, though!

richardxia (approved):
LGTM! Thanks for looking into the `.reduce()` changes!

app/pages/UcsfServiceTypePage/UcsfServiceTypePage.tsx
```diff
@@ -1,89 +1,76 @@
...
-    let updatedList;
-    if (updatedServiceType.name === 'See all') {
-      updatedList = serviceTypeList.map((item: serviceTypeItem) => (
-        {
-          ...item,
-          checked: updatedServiceType.checked,
-        }
-      ));
-      updatedList = massToggleServiceList(serviceTypeList as [], updatedServiceType.checked);
+const seeAllPseudoId = -1;
```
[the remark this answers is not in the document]
schroerbrian: This looks very interesting. Thanks for the detailed explanation. I briefly looked into somehow setting a default ID value on a See All type but I did not find anything conclusive. This looks like a great idea for a refactor or a similar feature in the future. Thanks!

schroerbrian:
Thanks for the review!

PR #1166, opened 2022-09-19
Migrate Texting to TypeScript.

richardxia:
This also makes a small modification to app/index.d.ts, where we make the type stub for CSS modules more precise by specifying that the object has string keys and string values. This was copied from the sheltertech.org codebase in
https://github.com/ShelterTechSF/sheltertech.org/blob/c12c90bbff7d0f8370ecd35d1bdec534a1c9f53a/src/shims.d.ts#L1-L15


t-petersen (approved):

tpetersen0308 (approved):

richardxia:
Thanks for reviewing, @tpetersen0308!

schroerbrian:
Thanks for updating this! Looks good. Sorry for not getting to it sooner

PR #1168, opened 2022-09-23
Edit breaking news page

dfloo:


schroerbrian (approved):
This looks excellent! Sorry for leaving this one hanging for so long; I wasn't sure if it was ready or not. I have pushed an update to this branch that fixes some merge conflicts that added up over the past few months.

A couple things I would mention -- it is our convention to use React functional components with React hooks over class-based components, and we prefer using SCSS modules over universal SCSS. I think this is fine to go in as is, but if you have the time and inclination, updating either or both would be great as well. You can take a look at `app/pages/ServiceDiscoveryResults/ServiceDiscoveryResults.tsx` to see how we use a functional component with `useState` and `useEffect` hooks, and check out `app/components/search/SearchResults/SearchResults.jsx` and `app/components/search/SearchResults/SearchResults.module.scss` to see how we use SCSS modules.

Anyhow, thanks for your contribution, this looks great overall. Depending on how you feel, I think you can merge or make the changes mentioned. I'm tagging @richardxia here as well in case he has any thoughts one way or the other.

app/pages/EditBreakingNewsPage.jsx
```diff
@@ -0,0 +1,172 @@
...
+  }
+
+  async onDelete(item) {
+    await dataService.APIDelete(`/api/news_articles/${item.id}`);
+
+    this.setState(prevState => ({
+      breakingNewsItems: prevState.breakingNewsItems
+        .filter(oldItem => oldItem.id !== item.id),
+    }));
+  }
```
richardxia, line 66: All of these methods are concurrency/async unsafe due to accepting the entire `item` as an argument. React does not guarantee that a component will always be rerendered immediately after a state update (it happens to do this by chance on the ancient version of React we're still running, but the main React feature development over the past few years has been to explicitly not make this happen in order to support asynchronous rendering, which can greatly improve performance).

You have a data race in that the value of `item` comes from its value at _render_ time, as it is captured into the closures that you construct in the `onClick` properties, but this can drift from the _actual_ state of the component. For example, if you edit a field on a news item, it will trigger the `onFieldChange` method, but if you also hit the save button _before_ the component is rerendered, then the value of `item` passed to `onSave()` will be outdated compared to its corresponding `state` counterpart.

What you should really do is change each of these methods to accept [a handle](https://en.wikipedia.org/wiki/Handle_(computing)) to the news article, not a full copy of the news article data, either using the article's `id` property or its array index as the handle. Then you should only make modifications to the news articles from within the `setState()` call, referencing the previous state of the news article from the `prevState` argument to the function being passed to `setState()`. This ensures that you're always operating on the latest version of the item, even if the state were modified before the component is rerendered.

app/pages/EditBreakingNewsPage.scss
```diff
@@ -0,0 +1,58 @@
...
+button {
+    margin-left: 20px;
+    :first-child {
+        margin-right: 5px
+    }
+}
+
+.flex-grow {
+    flex-grow: 1;
+}
```
richardxia, line 58: We generally avoid CSS utility classes in this codebase, preferring instead to give semantic names to CSS classes (e.g. `.form-header`) and applying styles to the semantic names.

app/pages/EditBreakingNewsPage.jsx
```diff
@@ -0,0 +1,172 @@
...
+
+  async onAddNew() {
+    const response = await dataService.post('/api/news_articles', {});
```
richardxia, line 69: I'm a little bit surprised that the API works this way, since I think it complicates things when we don't do any validation or even have a database column constraint that requires that any of the fields be set.

@jjfreund, did we specifically design the API such that none of the fields is mandatory, and if not, could we add some validation both at the Rails controller layer and the DB schema layer to require that these fields not be null?

I'm also noticing some unintuitive behavior on the edit page where if I click "Add Breaking News Story" and refresh before actually populating the fields or saving, then the empty item disappears from the UI. It looks like the empty items still get created in the database, but they never get displayed because the `NewsArticlesController#index` method apparently filters out any articles that have a null `effective_date`. Basically, if you fail to set the "effective date" and save the article before refreshing the page, it gets lost to the DB and cannot be surfaced in the edit UI.

I don't know if this was intentional, since we [only do an explicit null check for `expiration_date`, not `effective_date`](https://github.com/ShelterTechSF/askdarcel-api/blob/6a64dfa11d732e8c880f298e634ef4d24c355ac1/app/controllers/news_articles_controller.rb#L13), and I guess SQL treats `effective_date < ?` as "false" if `effective_date` is null.

app/pages/EditBreakingNewsPage.jsx
```diff
@@ -0,0 +1,172 @@
...
+  }
+
+  render() {
+    const { breakingNewsItems } = this.state;
+
+    return (
+      <div className="edit-breaking-news-page">
+        <div className="edit-breaking-news-header">
+          <h1>Edit Breaking News Items</h1>
+        </div>
```
richardxia, line 157: There's no point in creating a wrapper `<div>` for this because you haven't attached any distinct styling for the `<h1>` element and because both `<div>` and `<h*>` elements are block elements. Could you just squash these into a single element?

```suggestion
        <h1 className="edit-breaking-news-header">
          Edit Breaking News Items
        </h1>
```

Also, by only using a single element for the header, you can center the text much more simply by just using `text-align: center` instead of having to use flexbox. Your footer should also use `text-align: center`, but you need to change your footer's `<button>` style to not be `display: flex`.

app/pages/EditBreakingNewsPage.scss
```diff
@@ -0,0 +1,58 @@
...
+        font-weight: 600;
+    }
+    label {
+        font-size: 18px;
+    }
+}
+
+.news-item-form:not(:last-child) {
+    border-bottom: 1px solid grey;
+}
```
richardxia, line 30: I think it might be visually clearer to always insert the bottom border instead of omitting it for the last child, since otherwise there's nothing separating the "Add Breaking News Story" from individual news article form. This could be confusing to users because it would make the "Add Breaking News Story" button look like it'll just save the last news article, but instead it'll actually create a brand new one.

I think part of the disconnect is that most users probably would not expect a blank news article to be created and saved to the database upon clicking the "Add Breaking News Story". Creating an actual divider between the edit forms and the "Add Breaking News Story" button will at least help separate the two, even if it's still potentially misleading to a user.

app/pages/EditBreakingNewsPage.scss
```diff
@@ -0,0 +1,58 @@
...
+    flex-direction: row;
+    align-items: center;
+}
+
+.form-section {
+    justify-content: space-between;
+    div:not(:last-child) {
+        padding-right: 20px;
+    }
+}
```
richardxia, line 43: This comment may no longer be applicable if you make the other changes I suggested regarding getting rid of the flexbox styling for the header and footer, but in general, I don't think it's a great idea to split styling definitions for the same class (`.form-section` in this specific case) across multiple blocks like this. It would be more OK to use multiple blocks if you were using different kinds of CSS properties in each, such as font styling in one and spacing in another.

However, in this particular case, you set `display: flex` in the first block but `justify-content` in the second, where `justify-content` only makes sense for flexbox or grid. This makes the CSS difficult to follow, especially if it's not immediately obvious that `.form-section` is also being defined elsewhere.

I think you should remove `.form-section` from the first block, and just compile define the `.form-section` in a single block.

Also, you probably do not want `align-items: center` for `.form-section`. This will cause the fields to line up strangely if they are of different sizes:

Screen Shot 2023-02-20 at 11 46 55 PM

You probably really want `align-items: flex-end` so that the bottoms are all aligned.

app/pages/EditBreakingNewsPage.scss
```diff
@@ -0,0 +1,58 @@
...
+
+.edit-breaking-news-header {
+    font-size: 28px;
+    font-weight: 600;
+    padding: 10px 0px 0px;
+}
+
+.edit-breaking-news-footer {
+    padding-bottom: 20px;
```
richardxia, line 13: Use margin, not padding, to separate elements from each other. The difference would be much more obvious if you had a background color or a border on the element, but there are other subtle reasons why you should still prefer margin for separate elements even if you don't have a background color or a border.

app/pages/EditBreakingNewsPage.scss
```diff
@@ -0,0 +1,58 @@
...
+
+.form-section {
+    justify-content: space-between;
+    div:not(:last-child) {
+        padding-right: 20px;
+    }
+}
+
+.form-header {
+    padding-bottom: 10px;
```
richardxia, line 46: Use margin, not padding, to separate elements from each other.

richardxia (changes requested):
Thanks for the ping, @schroerbrian, and thanks for the original PR submission, @dfloo. Unfortunately, I'm going to take a harder stance here and strongly request that we make some changes to this PR before merging. Of all my comments, I think one about not using the solitary `button` selector in the CSS file is absolutely critical, since that one leaks into the styling of all button elements on the entire site, and that one would not even be fixed by CSS modules because CSS modules only affects class names, not tag names.

I also echo, though even more strongly, the request to rewrite the React component as a function-based component, and to use CSS modules, and I would also add rewriting it in TypeScript. Function vs. class based components is more of a matter of style, but I would strongly suggest we follow the predominant conventions in our codebase unless there's a good reason to diverge. I feel stronger about requiring CSS modules and TypeScript, since these specifically impact the maintainability of the codebase.

I'm happy to help with making any of these changes if you run into any issues or if you don't have the bandwidth to make them, @dfloo. In general, it's a good idea to add a pull request description so that the reviewers understand the context, intent, and status of a PR.

app/pages/EditBreakingNewsPage.tsx
```diff
@@ -0,0 +1,154 @@
...
+        return {
+          ...article,
+          [field]: value
+        };
+      }
+      return article;
+    }));
+  }
+
+  const renderForm = (article: NewsArticle, index: number): any => (
```
richardxia, line 71: Could you extract this into a separate component? While small functions that return JSX elements are fine, this is quite large and I think it would be more readable to pull it out into a separate component.

app/pages/EditBreakingNewsPage.tsx
```diff
@@ -0,0 +1,154 @@
...
+    <form className="form" key={article.id}>
```
richardxia, line 72: Keys should be assigned by expression where the `.map()` method is called, not the rendering of the individual items, since only the `.map()` expression can know what a unique key would be.

If you extract this into a component, then you can attach the key to the instantiation of that component, not the internal JSX that it is rendering.

app/pages/EditBreakingNewsPage.tsx
```diff
@@ -0,0 +1,154 @@
...
+
+  const onDelete = async (articleId: number) => {
+    await dataService.APIDelete(`/api/news_articles/${articleId}`);
+    const updatedNewsArticles = breakingNewsArticles.filter(({ id }: any) => id !== articleId);
+    console.log(updatedNewsArticles);
+
+    setBreakingNewsArticles(updatedNewsArticles);
+  };
+
+  const onFieldChange = (articleId: number, { target }: any) => {
```
richardxia, line 56: Please give the event a proper type. Even when I don't remember the exact event type off the top of my head, I can usually always just hover over the expression where you call this function (down in the JSX) and my editor will show me the type of this argument.

package-lock.json
richardxia, line 1: Could you revert the changes to this file? It doesn't look like you are updating package.json, so we shouldn't update package-lock.json.

richardxia (commented):
Thanks for pushing some updates, I've left another round of feedback here.

schroerbrian:
Thanks @dfloo for your great work and @richardxia for all the great feedback. Our org wants to get this in, so I'm merging this in as is. I merged in a [PR of my own](https://github.com/ShelterTechSF/askdarcel-web/pull/1253) into this branch to fix up a few last things. @dfloo -- if you have time, please feel free to respond to Richard's suggestions. Thanks all this is a very important feature!

richardxia:
Thanks, @schroerbrian, for getting this merged in, and thanks @dfloo, for doing the bulk of the work starting this!

PR #1169, opened 2022-09-30
1157 translation widget

schroerbrian:
Adds logic to the white label sites to enable/disable the Google Translate widget. Also adds translate=no attributes to HTML elements where we do not want translation to occur (addresses, org names, etc). Lastly, this adds a whiteLabelDefaults object with default props that are merged into all of our white label site objects.

Also this changes how layout is handled for services on the search result page – it replaces Flex with Grid

The first screenshot shows comparison between my local dev and Staging to demonstrate that mobile layout is the same after changing to grid. The second screenshot is of my local environment using grid on a standard desktop screen size.

Screen Shot 2022-10-04 at 11 23 46 AM

Screen Shot 2022-10-04 at 11 48 02 AM

### DIFF
--- a/app/components/search/SearchResults/SearchResults.module.scss
+++ b/app/components/search/SearchResults/SearchResults.module.scss
@@ -138,6 +138,9 @@
 .sideLinks {
   flex-shrink: 0;
   width: auto;
+  @media screen and (min-width: $min-tablet-p) {
+    max-width: 325px;
+  }
   margin-left: 25px;
   font-weight: bold;
   font-size: 18px;
--- a/app/components/search/SearchResults/SearchResults.module.scss
+++ b/app/components/search/SearchResults/SearchResults.module.scss
@@ -51,24 +51,23 @@
 }
 
 .searchResult {
-  display: flex;
-  flex-direction: row;
-  align-items: stretch;
-  justify-content: space-between;
-  padding-top: 21px;
-  padding-bottom: 24px;
-  padding-left: 24px;
+  display: grid;
+  grid-template-columns: 7fr 3fr;
+  gap: 25px;
+  justify-items: end;
+  padding: 21px 0 24px 24px;
   border-left: 1px solid $color-grey3;
   border-bottom: 1px solid $color-grey3;
 
   @media screen and (max-width: $break-desktop) {
-    flex-direction: column;
+    gap: 32px;
+    grid-template-columns: none;
+    justify-items: start;
   }
 
   @media screen and (max-width: $break-tablet-l) {
     padding-left: 30px;
   }
-
 
   @media screen and (max-width: 600px) {
     padding: 18px 20px;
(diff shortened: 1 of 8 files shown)


app/components/search/SearchResults/SearchResults.module.scss
```diff
@@ -138,6 +138,9 @@
 .sideLinks {
   flex-shrink: 0;
   width: auto;
+  @media screen and (min-width: $min-tablet-p) {
```
schroerbrian, line 141: The translation of "send a text message" link was rather lengthy in Spanish. I set this max-width to force the link to wrap as it was extending behind the search results map. I spent about an hour fooling around with `flex-wrap/grow/shrink`, `overflow`, `white-space`, `word-break` but with no success. Thus, I just decided to set the max-width. I'm open to better solutions if anyone has any, though!

richardxia (approved):
Nice job on refactoring the search results page to use Grid instead of Flexbox!

All the things you added `translate="no"` to LGTM! It's not blocking, but I'm wondering if there's some way to systematically ensure that certain fields don't get translated. The main drawbacks with this `translate="no"` approach are that:

1. They have to be added to each individual HTML element, even though the underlying data property/database column may show up in multiple places. I don't know how to phrase this concisely, but it'd be nicer if we could somehow annotate the properties that we don't want to translate and somehow have that either automatically add the `translate="no"` to the HTML or at least have something yell if we forget to set `translate="no"` for an HTML element that wraps a data property that should not be translated
2. It's going to be hard for developers to remember to appropriately add this `translate="no"` for new HTML that they write.

This might be moot if you're looking at using the API version of Google Translate, since then we could probably actually annotate things at the data level rather than at the presentation level.

schroerbrian:
> Nice job on refactoring the search results page to use Grid instead of Flexbox!
> 
> All the things you added `translate="no"` to LGTM! It's not blocking, but I'm wondering if there's some way to systematically ensure that certain fields don't get translated. The main drawbacks with this `translate="no"` approach are that:
> 
> 1. They have to be added to each individual HTML element, even though the underlying data property/database column may show up in multiple places. I don't know how to phrase this concisely, but it'd be nicer if we could somehow annotate the properties that we don't want to translate and somehow have that either automatically add the `translate="no"` to the HTML or at least have something yell if we forget to set `translate="no"` for an HTML element that wraps a data property that should not be translated
> 2. It's going to be hard for developers to remember to appropriately add this `translate="no"` for new HTML that they write.
> 
> This might be moot if you're looking at using the API version of Google Translate, since then we could probably actually annotate things at the data level rather than at the presentation level.

That makes a lot of sense. I could see this being quite difficult to maintain. I've created a [ticket](https://github.com/ShelterTechSF/askdarcel-web/issues/1181) for this work. It'd be great if the API version of Google Translate works, otherwise perhaps I or someone working on the ticket can devise another creative solution.

PR #1173, opened 2022-10-06
Add print results icon

schroerbrian:
NB: Please review the [translation PR](https://github.com/ShelterTechSF/askdarcel-web/pull/1169) first if it has not been merged in yet. This PR branches off of that PR as it adds to the whitelabel defaults object created in the former PR. 

This PR adds the print results button on the search results page. I couldn't really figure out an elegant way to hide non-relevant elements in the print view other than setting body and descendants visibility as hidden and then overriding that with the search results elements. I'm open to suggestions on improving this, though.

Firefox had some interesting bugs where certain elements remain on the print view given that they have a "transition: all" rule, which delays their visibility value change. I had to add some CSS overrides in the my print media query block to rectify this.

Checked across Safari, Chrome, and Firefox. Screenshot below of Chrome:
image


app/components/search/SearchResults/SearchResults.module.scss
```diff
@@ -179,3 +157,60 @@
...
   }
 }
+
+@media print {
+  // This below styles are for printing out the results on our service/search results page.
+
+  // This sets visibility hidden on all elements except for the searchResultsContainer
+  // and its children to avoid printing irrelevant sections of the page.
+  // This follows the example of SO this answer https://stackoverflow.com/a/2618980
+  body {
```
schroerbrian, line 167: See comment in description about using this solution to show only the search results in the print view.

richardxia:
> NB: Please review the https://github.com/ShelterTechSF/askdarcel-web/pull/1169 first if it has not been merged in yet. This PR branches off of that PR as it adds to the whitelabel defaults object created in the former PR.

Thanks for the heads up. I hope you don't mind, but I'm going to change the base branch for this PR so that it targets #1169 so that the diff is easier for me to review. In case you didn't know, you can set the base branch for a PR, and if the base branch is later merged into the main branch, then GitHub will automatically update your PR's base branch to the main branch, though you might have to rebase and push back up to GitHub due to old commits getting stale.

If you need to change the base branch back to `master`, here are the instructions for changing the base branch of an already open PR: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/changing-the-base-branch-of-a-pull-request

richardxia:
> I hope you don't mind, but I'm going to change the base branch for this PR so that it targets #1169 so that the diff is easier for me to review.

Nevermind, I see that this is not much easier to review because this branch also includes an extra merge of `master` that #1169 didn't have. I'm going to switch the base branch back to `master`.

app/components/search/SearchResults/SearchResults.module.scss
```diff
@@ -179,3 +157,60 @@
...
+
+@media print {
+  // This below styles are for printing out the results on our service/search results page.
+
+  // This sets visibility hidden on all elements except for the searchResultsContainer
+  // and its children to avoid printing irrelevant sections of the page.
+  // This follows the example of this answer https://stackoverflow.com/a/2618980
+  body {
+    visibility: hidden;
+  }
```
richardxia, line 169: This is kind of dangerous because it's not scoped to just this page, so it actually impacts the print view of _all_ pages on the site. I confirmed this by attempting to print the Home Page, which is totally blank in the print view. I think the general approach in the StackOverflow link you wrote in the comment is probably fine, but it was probably written for an era before SPAs, where you could just have the backend serve a different CSS file per page.

I think one way we could still use this approach but scope stuff to just this page would be to attach a custom CSS class to the `<body>` element. Since `<body>` is outside of the React root element, you'll have to use a React Effect to mutate the `<body>` element, and you'll have to remember to undo the CSS class when the component unloads, but it should be pretty easy to do. Here's an article showing how to use the `useEffect` hook to add CSS class to the `body` on component load and remove it on component unload: https://bobbyhadz.com/blog/react-set-body-style. Once you do this, then I think we can change this CSS selector from `body` to whatever class name you chose (note that since we're using CSS modules, you'll probably need to get the class name from the `styles` object so that you have the mangled name).

I'm not sure if this is the best way of doing it, but I think it should be safe, since React Hooks should guarantee that we run the effect on load and unload.

app/components/search/SearchResults/SearchResults.module.scss
```diff
@@ -179,3 +157,60 @@
...
+
+  .searchResultsContainer {
+    visibility: visible;
+  }
+
+  .searchResultsContainer {
```
richardxia, line 175: These are the same CSS class, right? These two CSS blocks can be combined.

app/components/search/SearchResults/SearchResults.module.scss
```diff
@@ -179,3 +157,60 @@
...
+    position: absolute;
+    top: 0;
+    bottom: 0;
+    left: 0;
+    right: 0;
+    margin: 50px auto;
+    width: 700px;
```
richardxia, line 182: Pixels are a bit weird of a unit to use for printing. There is actually an [`in` unit](https://developer.mozilla.org/en-US/docs/Web/CSS/length#in) for inches, which is defined as 96px. We could maybe use that unit here, or maybe we should just use relative units, at least for the horizontal dimension. I have no idea if the `vw` unit works for printing, but that's potentially an option. We might also be able to just use percents.

app/components/search/SearchResults/SearchResults.module.scss
```diff
@@ -179,3 +157,60 @@
...
+    // delay on changing the visibility value.
+    transition: none !important;
+  }
+
+  a[href]::after {
+    // Our _reset.scss file displays the whole href of a link in the ::after selector.
+    // This is confusing for our service and resource links which are just routes
+    // to other parts of our site, etc. Thus, overriding that here.
+    content: "" !important;
+  }
```
richardxia, line 215: Each of these is also affected by the global CSS issue that I mentioned in my comment about `body`, though these could possibly be fixed without having to attach a CSS class to the `body` element. If you just take the CSS class for the root element of this page (`.searchResultsContainer`?) and put it before each of these selectors, then it will scope it everything down to just inside of this component.

richardxia (approved):
Overall, this LGTM. Thanks for doing all the testing across multiple browsers.

My main concern is with the global CSS selectors in the print styles, which actually leak to other pages because we have a single-page app (SPA). I think my suggestion to attach a custom CSS class to the `<body>` element via a React Effect should work, but there's probably many ways to do it, so if you come up with a better way of doing scoping the CSS styles, feel free to implement it in any other way.

schroerbrian:
> > NB: Please review the #1169 first if it has not been merged in yet. This PR branches off of that PR as it adds to the whitelabel defaults object created in the former PR.
> 
> Thanks for the heads up. I hope you don't mind, but I'm going to change the base branch for this PR so that it targets #1169 so that the diff is easier for me to review. In case you didn't know, you can set the base branch for a PR, and if the base branch is later merged into the main branch, then GitHub will automatically update your PR's base branch to the main branch, though you might have to rebase and push back up to GitHub due to old commits getting stale.
> 
> If you need to change the base branch back to `master`, here are the instructions for changing the base branch of an already open PR: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/changing-the-base-branch-of-a-pull-request

Oops thanks for that. I forgot about doing that as a better alternative.

app/components/search/SearchResults/SearchResults.module.scss
schroerbrian: I did a bit of research and couldn't find anything better than `useEffect` so I went with that. Thanks for the suggestion!

app/components/search/SearchResults/SearchResults.module.scss
schroerbrian: Thanks, fixed!

app/components/search/SearchResults/SearchResults.module.scss
schroerbrian: Oops, good catch.

PR #1174, opened 2022-10-07
Disable Deactivate button when resource is already deactivated. Perfo…

schroerbrian:
…rm redirect on resource and service listing page when either is deactivated.

Pull this API branch down locally if you'd like to test: https://github.com/ShelterTechSF/askdarcel-api/pull/652
Both this and that PR are designed to go in together.

### DIFF
--- a/app/models/Service.ts
+++ b/app/models/Service.ts
@@ -35,6 +35,7 @@ export interface Service {
   resource: Organization;
   schedule: Schedule;
   source_attribution: string;
+  status: string;
   updated_at: string;
   url: string | null;
   verified_at: any;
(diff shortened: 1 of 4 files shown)


app/models/Service.ts
```diff
@@ -35,6 +35,7 @@ export interface Service {
   resource: Organization;
   schedule: Schedule;
   source_attribution: string;
+  status: string;
```
richardxia, line 38:
```suggestion
  status: 'pending' | 'approved' | 'rejected' | 'inactive';
```

If you feel comfortable with it, do you think we could turn this into a union of string literals? This both serves as documentation on the frontend side and it ensures that we don't attempt to compare it to an invalid string value (I think, at least; we might have to turn on some TypeScript or ESLint options for that).

richardxia (approved):
LGTM

app/models/Service.ts
schroerbrian: Great suggestion! I have added this. I tested this by comparing a `status` prop against a junk string; the TS compiler complained, so i think we're good.

Issue #1175, opened 2022-10-10
Migrate all code to TypeScript


richardxia:
This is a mega issue that tracks all of the remaining files that need to be migrated to TypeScript. This probably shouldn't all be done in a single PR, since there are many files left and migrating them may interfere with any concurrent work on new features, so if you want to work on this, I would suggest working on groups of related files and to add your name to each file you're working one:

- [ ] `assets/index.js`
- [x] `components/edit/EditAddress.jsx` @richardxia
- [x] `components/edit/EditCollection.jsx` @richardxia
- [x] `components/edit/EditNote.jsx` @richardxia
- [x] `components/edit/EditNotes.jsx` @richardxia
- [x] `components/edit/EditPhones.jsx` @richardxia
- [x] `components/edit/EditSchedule.jsx` @richardxia
- [x] `components/edit/EditScheduleDay.jsx` @richardxia
- [x] `components/edit/EditServices.jsx` @richardxia
- [x] `components/edit/EditSidebar.jsx` @richardxia
- [x] `components/edit/FormTextArea.jsx` @richardxia
- [x] `components/edit/MultiSelectDropdown.jsx` @richardxia
- [x] `components/edit/ProvidedService.jsx` @richardxia
- [x] `components/listing/__tests__/RelativeOpeningTimes.spec.jsx` @richardxia
- [x] `components/listing/feedback/FeedbackForm.jsx` @richardxia
- [x] `components/listing/feedback/FeedbackSteps.jsx` @richardxia
- [x] `components/listing/feedback/constants.js` @richardxia
- [ ] `components/search/Refinements/ClearAllFilters.jsx` @richardxia
- [ ] `components/search/Refinements/FacetRefinementList.jsx` @richardxia
- [ ] `components/search/Refinements/OpenNowFilter.jsx` @richardxia
- [ ] `components/search/Refinements/OpenNowRefinementList.jsx` @richardxia
- [ ] `components/search/Refinements/RefinementListFilter.jsx` @richardxia
- [ ] `components/search/SearchMap/SearchEntry.jsx` @richardxia
- [ ] `components/search/SearchResults/SearchResults.jsx` @richardxia
- [x] `hooks/APIHooks.js` @richardxia 
- [x] `models/User.js` @richardxia
- [x] `models/__tests__/User.spec.js` @richardxia
- [x] `pages/HomePage/components/ResourceCard/ResourceCard.jsx` @richardxia 
[...]

PR #1180, opened 2022-10-10
Migrate home page components to TypeScript

richardxia:
Refs #1175.

This migrates all of the widgets on the home page to TypeScript. The biggest change is that I upgraded the Typeform package from 0.13.1 to 2.0.0, which actually spans a ton of versions. The reason I did this is because the 0.13.1 version of the package lacks any type definitions, but the newer versions have TypeScript type definitions that come with the package. Fortunately, our usage of Typeform is pretty minimal, so it wasn't bad at all to upgrade.

I did some manual testing of the Typeform buttons (the stuff under the "Get step-by-step help" section on the page) to make sure that they still behave the same and look right.


richardxia:
> Looks great! Thanks for making these changes. This is funny timing considering the execs have [just asked](https://sheltertech.slack.com/archives/CSKTQR54L/p1665425501836129) for the `Typeforms` to be removed from the home page. If you have the bandwidth and desire, I'd say you could just remove `Typeform` references outright; however, if not, I will get to them soon enough.

Yeah, I thought it was pretty funny that the request to remove the Typeform forms came in right after I upgraded Typeform. If you don't mind, I think it might be nice to just merge this in now, since this updates several other files unrelated to Typeform, and then we can remove the entire Typeform section of the page in a followup PR, just to keep the tasks separated.

richardxia:
I'm going to hit the merge button, but feel free to revert if you wanted to remove the Typeform stuff before merging this other stuff in.

PR #1185, opened 2022-10-31
Migrate Feedback components to TypeScript.

richardxia:
Refs #1175.

This migrates the Feedback components to TypeScript. There are a few minor, non-type changes that I made in order to get everything to pass type checking, but they shouldn't cause any functional changes to the app.


schroerbrian (approved):
Looks good! Thanks for making these changes.

richardxia:
Thanks for the review!

PR #1196, opened 2022-11-22
Swap see all logic in UCSF eligibility and subcategory pages - if see…

schroerbrian:
… all is selected, deselect all other refinements so as to not restrict results.

UCSF gave us feedback that the current flow was confusing. Turns out, I had the logic wrong here. When "See All" is selected, all refinements should then be unselected. This is because selecting  refinement(s) can restrict query results


richardxia (approved):
LGTM, and nice catch!

PR #1199, opened 2022-12-01
Migrate ListingDemoPage to TypeScript.

richardxia:
Refs #1175.

This migrates a couple of testing and debug-only files to TypeScript. It was honestly kind of difficult to figure out the exact types, since they're pretty dynamic, so I'm glad that I did this because now it's easier to understand the types of the test data here.

### DIFF
--- a/app/pages/debug/ListingDemoPage.tsx
+++ b/app/pages/debug/ListingDemoPage.tsx
@@ -8,6 +8,10 @@ import {
 } from '../../models';
 import './ListingDemoPage.scss';
 
+type Interval = readonly [number, number];
+type DaySchedule = null | Interval | readonly Interval[];
+type WeekSchedule = readonly DaySchedule[];
+
 // Is sorted so that the first day is today
 /* eslint-disable max-len */
 export const simpleScheduleLookup = {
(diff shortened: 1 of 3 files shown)


app/pages/debug/ListingDemoPage.tsx
```diff
@@ -24,29 +28,41 @@ export const simpleScheduleLookup = {
...
-export const createScheduleFromShorthand = shedule_shorthand => {
+/** Return true if x is an Interval[]; false if it is only an Interval.
+ *
+ * This is a user-defined type guard, since TypeScript is unable to infer the
+ * correct type normally.
+ *
+ * For more information on user-defined type guards, see:
+ * https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates
+ */
+function isIntervalArray(x: Interval | readonly Interval[]): x is readonly Interval[] {
```
davidagustin, line 44: I still don't understand why the return type here is not a Boolean.

davidagustin (changes requested):

app/pages/debug/ListingDemoPage.tsx
richardxia: It's just special syntax in TypeScript for user-defined type guards. I linked to the TypeScript documentation on user-defined type guards in the comment for this function. Is there a particular part of that documentation that you wanted a more detailed explanation for?

davidagustin: Got it. I saw a part of the documentation but there are different kinds of narrowing.

davidagustin (approved):

davidagustin:
Learnt more typescript today. LGTM

schroerbrian:
Looks like a robust and helpful change. Thank you!

PR #1202, opened 2022-12-05
Allow lang query param to translate sites

drcaramelsyrup:
* Enables `lang=` query param to specify a target translation language so SFFamilies portal site may link to the visitor's preferred language

This is done by overriding the (undocumented, as far as I can tell) `googtrans` cookie that the Google Translate widget inspects - this seems to be the only method out of other [suggested options to set a target language](https://stackoverflow.com/questions/1773687/google-translate-set-default-language/8493765#8493765) that works reliably.

Technically only the SFFamilies whitelabel is requesting this feature, but I think the language override could potentially be useful across sites. This shouldn't do anything if translation isn't enabled.

For example, with `lang=zh-TW`:
image

This session cookie should be scoped to the domain (and root + child paths):
image

### DIFF
--- a/app/components/ui/Translate.tsx
+++ b/app/components/ui/Translate.tsx
@@ -1,16 +1,38 @@
 import React from 'react';
+import { useCookies } from 'react-cookie';
 import { Helmet } from 'react-helmet-async';
+import { useLocation } from 'react-router-dom';
 import { whiteLabel } from '../../utils';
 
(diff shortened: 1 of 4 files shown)


schroerbrian (approved):
LGTM! Thank you, this is a great enhancement.

drcaramelsyrup:
Thanks! Will merge it shortly.

PR #1204, opened 2022-12-09
Add refinementListLimit value to whitelabel config and pass to eligib…

schroerbrian:
…ility and subcategory list in sidebar.

The Algolia refinement list widget has a default list limit of 10. This increases the limit for UCSF refinements as some were being cut off. In the future, when we have a better, more consolidated eligibility/subcategories, we can consider removing this config prop.


drcaramelsyrup (approved):
reran flaky e2e build, lgtm!

schroerbrian:
> reran flaky e2e build, lgtm!

Thanks checking and for rerunning the test! It has been failing a lot recently

PR #1205, opened 2022-12-13
Allow Texting modal to be scrollable.

richardxia:
This is important for shorter screens (< ~700px), where the submit button on the Texting modal may not be visible.

This also adds a somewhat unrelated (though kind of related) fix to the Texting modal, where the Filters bar from the search results page was above the Texting modal along the z dimension. Forcing the main page and the modal to be in different stacking contexts fixes this issue and overall makes the stacking and scrolling a bit easier to reason about.

### Before

Screen Shot 2022-12-12 at 5 43 00 PM

### After (Filters menu z-index fix)

Screen Shot 2022-12-12 at 5 43 14 PM

### After (Demonstrating that modal can be scrolled)

Screen Shot 2022-12-12 at 5 43 20 PM

---

FWIW, the way I tested this was with Chrome's responsive mode, setting the resolution to 1024 x 600 px, since my display resolution is far taller than 600 px.

### DIFF
--- a/app/App.module.scss
+++ b/app/App.module.scss
@@ -1,0 +1,7 @@
+.outerContainer {
+  /* Creates a stacking context such that everything inside of the
+   * .outerContainer is either above or below anything outside of the
+   * .outerContainer. One example of something outside of the .outerContainer is
+   * modals. */
+  z-index: 0;
+}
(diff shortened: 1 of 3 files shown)


schroerbrian:
LGTM! I tested using Chrome's Pixel simulator on a 650px height. It does not render very well for landscape mode, but I don't think we need to worry about that at this point.

schroerbrian (approved):

PR #1207, opened 2022-12-16
Use single page for UCSF subcategory and eligibility. Move subcategor…

schroerbrian:
…y and eligibility pages into components and share code where appropriate.

The impetus for this consolidation was that the Immigration tile only has eligibility refinements, so I needed to find a way to skip the subcategory refinements page. I'm sorry to make you review some duplicated code twice! At least this change also removes a lot of more or less duplicated code and also improves the experience (for instance subcategories are fetched prior to the user nav'ing to the subcategory page so there is no flash after they load). 

Testing locally, everything looks and feels fine and looks identical to what is on QA One at the moment. Let me know if you have any questions or requests to make this easier. Thanks!


app/components/search/SearchResults/SearchResults.module.scss
```diff
@@ -54,7 +54,7 @@
   display: grid;
   grid-template-columns: 7fr 3fr;
   gap: 25px;
-  justify-items: end;
+  justify-items: stretch;
```
schroerbrian, line 57: This fixes a bug where certain search results would return and be justified towards the map (when the text was minimal and thus the search results body was narrow to move to the end)

app/components/ucsf/RefinementLists/EligibilityRefinements.tsx
```diff
@@ -91,74 +87,3 @@ const ClientEligibilities = ({
     </div>
   );
 };
-
```
schroerbrian, line 94: Page component refactored/consolidated with `ServiceType` Page component and  moved to UcsfDiscoveryForm.tsx

app/components/ucsf/RefinementLists/SubcategoryRefinements.tsx
```diff
@@ -0,0 +1,74 @@
+import React from 'react';
```
schroerbrian, line 1: I'm not sure why Git recognized the EligibilityRefinements file as being moved from a previous file and updated, but not the SubcategoryRefinements file. Sorry about that! There should not have been any major changes to this file other than removing the `Page` component and editing some class names :(

schroerbrian:
Approving for now, to begin QA testing

richardxia (commented):
Sorry for the late review, but LGTM! Thanks for pushing on this and getting this through the finish line. I'm excited for this to finally launch soon!

PR #1209, opened 2022-12-23
Update Cypress to v 12. Add TS interface to app Config const.

schroerbrian:
Makes updates in reference to @richardxia 's helpful comments on https://github.com/ShelterTechSF/askdarcel-web/pull/1206

I updated Cypress from v10 to 12 which fixed an issue with caching the response to Cypress intercepts and negated a need for us to add a funky fix for that. Bumping the version caused a few of our tests to faill; we were using a `.should` after requests where we should have been using a `.then`. i'm not sure why this was done to begin with but it seemed like a pretty straightforward and reasonable change to make as part of this PR.

I also added an interface for our config object.

### DIFF
--- a/app/config.ts
+++ b/app/config.ts
@@ -3,6 +3,7 @@ declare global {
 }
 
 interface Config {
+  GOOGLE_ANALYTICS_ID: string;
   LINKSF_DOMAIN: string;
   MOHCD_DOMAIN: string;
   MOHCD_SUBDOMAIN: string;
--- a/app/config.ts
+++ b/app/config.ts
@@ -2,7 +2,16 @@ declare global {
   const CONFIG: any;
 }
 
-const config = {
+interface Config {
+  LINKSF_DOMAIN: string;
+  MOHCD_DOMAIN: string;
+  MOHCD_SUBDOMAIN: string;
+  SFFAMILIES_DOMAIN: string;
+  UCSF_DOMAIN: string;
+  [key: string]: any;
+}
+
+const config: Config = {
   ALGOLIA_APPLICATION_ID: CONFIG.ALGOLIA_APPLICATION_ID,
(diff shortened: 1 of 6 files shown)


app/config.ts
```diff
@@ -2,7 +2,16 @@ declare global {
...
 }
 
-const config = {
+interface Config {
+  LINKSF_DOMAIN: string;
+  MOHCD_DOMAIN: string;
+  MOHCD_SUBDOMAIN: string;
+  SFFAMILIES_DOMAIN: string;
+  UCSF_DOMAIN: string;
+  [key: string]: any;
```
schroerbrian, line 11: It seems like all of these values are probably strings, but I didn't want to make that assumption considering some of them are passed in dynamically by the app's ENV object (though I guess all env args are probably strings by default??)

richardxia: Yeah, I think these will probably always be strings in practice, but it is hard to be 100% type safe here because these all ultimately come from a Webpack plugin that injects these values in at build time.

Another alternative to `any` is [`unknown`]([url](https://mariusschulz.com/blog/the-unknown-type-in-typescript)). Whereas `any` will always pass all type checks, `unknown` basically won't let you do anything with it unless you either use a `typeof` expression, a custom type guard, or make a type assertion. Although this may cause more friction to a developer, it might serve as a reminder for us to go back and update this interface with the precise type.

richardxia (approved):
LGTM, and my one and only comment doesn't need to be addressed, since we've got a lot of `any` gaps in our codebase today, anyway. I hope to get all of our vanilla JS files converted to TypeScript before we start tackling the `any`'s, since that's probably a higher value task.

app/config.ts
schroerbrian: Awesome, that's a great suggestion. Thanks for the feedback.

PR #1212, opened 2023-01-03
Fix React Rules of Hooks violations

richardxia:
Note: This PR is not quite ready to merge yet, since there's one remaining rule violation that I wasn't sure what the right fix is. I'm going to leave an inline comment in the diff on the block of code in question.

There are a few rules that we have to follow when using React Hooks, documented at https://reactjs.org/docs/hooks-rules.html. While working on something else, I realized that we were violating some of these rules, in particular, the rule that hooks cannot be conditionally executed. This means that every component that calls a hook must do so unconditionally; you should not call them from within `if` blocks, `for` loops, and you also can't do an early return before all hooks have executed.

If you're interested in _why_ this rule exists, this article I read several years ago helped me understand the mental model of what exactly is going on with React Hooks: https://medium.com/@ryardley/react-hooks-not-magic-just-arrays-cd4f1857236e.

Fortunately, React has published an official ESLint plugin for detecting violations: https://www.npmjs.com/package/eslint-plugin-react-hooks

This PR installs and enables that plugin, and it fixes the violations in our codebase. Aside from fixing conditionally executed hooks, the other main type of error we had was not properly declaring the dependencies list in `useEffect()` calls. This latter one may be reasonable to waive in certain scenarios, but I don't think any of the violations in our codebase were worth waiving. See https://github.com/facebook/react/issues/14920#issuecomment-471070149 for the official GitHub issue where the React team collected a lot of feedback on this rule and tried to provide detailed responses to specific scenarios.

Note that I haven't done any testing of these changes, so in addition to the remaining rule violation that needs to be fixed, I want to double check that these changes don't cause any new bugs in the app. Almost all of these changes will have functional changes on the code, and I want to make sure that we weren't relying on certain hooks not firing due to (accidentally?) omitting dependencies in the dependency lists or due to (accidentally?) conditionally executing hooks.

### DIFF
--- a/app/App.tsx
+++ b/app/App.tsx
@@ -82,7 +82,7 @@ export const App = () => {
       ReactGA.set({ page });
       ReactGA.pageview(page);
     });
-  }, []);
+  }, [history]);
 
   return (
     <div id={outerContainerId} className={styles.outerContainer}>
--- a/app/components/search/SearchResults/SearchResults.jsx
+++ b/app/components/search/SearchResults/SearchResults.jsx
@@ -31,10 +31,8 @@ const transformHits = hits => hits.map(hit => {
 // Todo: setExpandList will be used as part of next stage of multiple location work
 // eslint-disable-next-line no-unused-vars
 const SearchResults = ({ searchResults, expandList, setExpandList }) => {
-  if (!searchResults) return null;
   const [centerCoords, setCenterCoords] = useState(null);
   const [googleMapObject, setMapObject] = useState(null);
-  const hits = transformHits(searchResults.hits);
 
   useEffect(() => {
     if (centerCoords) {
@@ -44,7 +42,11 @@ const SearchResults = ({ searchResults, expandList, setExpandList }) => {
     return () => {
       document.body.classList.remove('searchResultsPage');
     };
-  }, []);
+  }, [googleMapObject, centerCoords]);
+
+  if (!searchResults) return null;
+
+  const hits = transformHits(searchResults.hits);
 
   return (
     <div className={styles.searchResultsAndMapContainer}>
--- a/app/components/ucsf/RefinementLists/EligibilityRefinements.tsx
+++ b/app/components/ucsf/RefinementLists/EligibilityRefinements.tsx
@@ -1,4 +1,4 @@
-import React, { useState, useEffect } from 'react';
+import React from 'react';
 
 import { Checkbox } from 'components/ui/inline/Checkbox/Checkbox';
 
@@ -14,22 +14,12 @@ interface SelectedEligibilities {
 }
 
 export const EligibilityRefinements = ({
-  resourceEligibilityGroups, selectedEligibilities, setSelectedEligibilities, resourceSlug,
+  resourceEligibilityGroups, selectedEligibilities, setSelectedEligibilities,
 }: {
   resourceEligibilityGroups: EligibilityGroup[];
   selectedEligibilities: SelectedEligibilities;
   setSelectedEligibilities: (categories: SelectedEligibilities) => void;
-  resourceSlug: string;
 }) => {
-  // Why is eligibilityGroupList a state variable and not just a prop?
-  useEffect(() => {
-    setEligibilityGroupList(resourceEligibilityGroups);
-  }, [resourceSlug]);
-
-  const [eligibilityGroupList, setEligibilityGroupList] = useState<EligibilityGroup[]>(
-    resourceEligibilityGroups,
-  );
-
   const handleEligibilityClick = (
     eligibility: Eligibility,
     eligibilities: Eligibility[],
@@ -59,7 +49,7 @@ export const EligibilityRefinements = ({
     <div className={styles.refinementsBox}>
       <div className={styles.refinementsBox_title}>Client Identity</div>
       <ol className={styles.refinementsLabels}>
-        {eligibilityGroupList.map(eligibilityGroup => (
+        {resourceEligibilityGroups.map(eligibilityGroup => (
           <li key={eligibilityGroup.label} className={styles.listContainer}>
             <span className={styles.refinementsGroupLabel}>
               {eligibilityGroup.label}
--- a/app/components/ucsf/RefinementLists/EligibilityRefinements.tsx
+++ b/app/components/ucsf/RefinementLists/EligibilityRefinements.tsx
@@ -21,6 +21,7 @@ export const EligibilityRefinements = ({
   setSelectedEligibilities: (categories: SelectedEligibilities) => void;
   resourceSlug: string;
 }) => {
+  // Why is eligibilityGroupList a state variable and not just a prop?
   useEffect(() => {
     setEligibilityGroupList(resourceEligibilityGroups);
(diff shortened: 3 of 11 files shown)


app/components/ucsf/RefinementLists/EligibilityRefinements.tsx
```diff
@@ -21,6 +21,7 @@ export const EligibilityRefinements = ({
   setSelectedEligibilities: (categories: SelectedEligibilities) => void;
   resourceSlug: string;
 }) => {
+  // Why is eligibilityGroupList a state variable and not just a prop?
   useEffect(() => {
     setEligibilityGroupList(resourceEligibilityGroups);
   }, [resourceSlug]);
```
richardxia, line 27: This line is causing the last remaining error, due to `resourceEligibilityGroups` not being present in the dependencies list. As written today, this hook only fires when `resourceSlug` is changed, or on the very first render of this component after it is mounted.

I'm not sure if adding `resourceEligibilityGroups` to the dependencies list is safe, in that I don't know if we were relying on only actually updating the `resourceEligibilitiesGroups` when the `resourceSlug` is also updated at the same time.

I tried reading the surrounding code, and I'm a little bit confused by what this `useEffect()` call is trying to accomplish in the first place. It might just be a bit of weird, vestigial code arising from multiple independent refactorings, and unfortunately, the git history is also difficult to follow because this code has come from multiple splicing/desplicing and file renames.

The main weirdness to me is that `setEligibilityGroupList()` is only ever called here in this `useEffect()` block. It's not clear to me why this even needs to be a state variable in the first place, since it's basically only initialized whenever `resourceSlug` changes. `resourceEligibilityGroups` itself is a prop of the this component, and there's some nontrivial code in `UcsfDiscoveryForm.tsx` that produces and operates on it.

@schroerbrian, since you've touched this code a lot lately, I'm wondering if you could provide a bit of insight here? Without completely grokking the calling code, I think the simplest fix would be to remove the `eligibilityGroupList` state variable from this component and instead just use the `resourceEligibilityGroups` prop in its place. However, I don't know if there's something more nuanced going on here, so any knowledge you have about this code would be great hear!

app/hooks/APIHooks.js
```diff
@@ -14,9 +14,11 @@ export const useEligibilitiesForCategory = categoryID => {
   const [eligibilities, setEligibilities] = useState(null);
 
   useEffect(() => {
-    dataService.get(`/api/eligibilities?category_id=${categoryID}`).then(response => {
-      setEligibilities(response.eligibilities);
-    });
+    if (categoryID !== null) {
```
richardxia, line 17: For a bit more information about this change, because this custom hook was originally being called conditionally in the `*DiscoveryForm` components, when I changed those components to unconditionally call this hook, I had to now guard against the case where `categoryID` is null. Same goes for the other custom hook lower in this file.

This is actually how I initially discovered the React Hooks violations, since I was working on porting this file to TypeScript, and I was looking at how these hooks were being called, when I realized that we were illegally conditionally calling the hooks.

app/components/ucsf/RefinementLists/EligibilityRefinements.tsx
schroerbrian: Thanks for catching this, Richard! I really can't recall why I used `useEffect` here though I vaguely remember there being a reason for this. I switched to using the `resourceEligiblityGroups` prop locally, and it seems to be working fine. I have created a [PR](https://github.com/ShelterTechSF/askdarcel-web/pull/1213/files) into this branch that makes the switch.

app/App.tsx
```diff
@@ -82,7 +82,7 @@ export const App = () => {
       ReactGA.set({ page });
       ReactGA.pageview(page);
     });
-  }, []);
+  }, [history]);
```
schroerbrian, line 85: This is unrelated to your PR, but I'm wondering if the `getLocation` call should be run within a `useMemo` hook. It's my understanding that the `useMemo` hook fires regardless of render status, so it would return the location data faster. If this sounds good, I will work on changing it in a future PR

schroerbrian (approved):
This looks great! Thanks for making these very important changes.

app/App.tsx
richardxia: According to the [React docs](https://reactjs.org/docs/hooks-reference.html#usememo), `useMemo()` should only be used for memoizing expensive computations, and you shouldn't use it to do anything that shouldn't happen during rendering, such as side effects. Since our `getLocation()` function makes an API call to the browser location API and then falls back to making an API call to Google's location API, this is side effectful, so I don't think we want to use `useMemo()` for this. The React docs also say that `useMemo()` is not guaranteed to memoize all previous values, so it may have to rerun the function again, even if the dependencies haven't changed. This would be bad for side effectful things things like querying the browser location or firing off an API request to an external service.

I think if we wanted to fire off the location query before the initial render of the app, we might be able to do it by having a wrapper component whose only job is to fire off the location query, where we keep around a state variable that prevents the rest of the app from actually rendering until the effect has fired first. Something like:

```jsx
const LocationWrapper = () => {
  const [isFirstRender, setIsFirstRender] = useState(true);
  const [userLocation, setUserLocation] = useState(null);
  useEffect(() => {
    getLocation().then((x) => setUserLocation(x));
    setIsFirstRender(false);
  });

  if (isFirstRender) return null;
  return <App />;
}

File: app/components/search/SearchResults/SearchResults.jsx [hunk 4]
```diff
@@ -44,7 +42,11 @@ const SearchResults = ({ searchResults, expandList, setExpandList }) => {
     return () => {
       document.body.classList.remove('searchResultsPage');
     };
-  }, []);
+  }, [googleMapObject, centerCoords]);
```
richardxia, line 45: I was looking at manually testing this, but it doesn't look like `setCenterCoords()` is ever called. I'm not going to remove this in this PR, but I think it's something we may want to do when we migrate this file to TypeScript, where it should hopefully be a lot more obvious when something is unused.

richardxia:
I'm a little less certain on how to manually test the UCSF pages, since I don't have the DB populated locally with either real or fake data for that site, but it sounded like you were able to manually test some of it when you submitted that other PR, so I'm going to assume that everything should be OK.

Thanks again for reviewing this and for submitting the other PR to get rid of the unnecessary `useEffect()` call!

app/components/search/SearchResults/SearchResults.jsx
schroerbrian: I think it may be leftover from the multiple location work that was never fully implemented.  I will make a note to look into whether it can be of use to us.

schroerbrian:
> I'm a little less certain on how to manually test the UCSF pages, since I don't have the DB populated locally with either real or fake data for that site, but it sounded like you were able to manually test some of it when you submitted that other PR, so I'm going to assume that everything should be OK.
> 
> Thanks again for reviewing this and for submitting the other PR to get rid of the unnecessary `useEffect()` call!

Yep, I was able to look into testing the UCSF white label. I will also merge this into qa-one and push up so that Fred and others can have a look during their QA testing.

PR #1213, opened 2023-01-03
Change unneeded state value to component prop in EligibilityRefinemen…

schroerbrian:
…ts component to fix hook violation

I changed this to be a component prop rather than a state var. I have a vague recollection of needing to use the `useEffect` before, but that was 1 or 2 refactors ago. I have tested this out locally and all seems fine.


richardxia (approved):
Thanks for making this change! I'm going to merge this and then do one last sanity check of my other PR!

PR #1214, opened 2023-01-09
Mmigrate APIHooks and DataService to TypeScript

richardxia:
Refs #1175.

This migrates the APIHooks and DataService files over to TypeScript. There should be no functional change, as the only changes are to the type annotations. Some of the more notable changes include adding the `@types/lodash` package as a dependency and fixing a slightly incorrect type annotation `ServiceDiscoveryResults.tsx`, which was claiming that the argument are of type `CategoryRefinement`, but in reality, it is just `Category`, lacking the Algolia-specific fields that `CategoryRefinement` has.


schroerbrian (approved):
LGTM! Thanks for making this change.

richardxia:
Thanks for reviewing!

PR #1216, opened 2023-01-19
1176 setup prettier

schroerbrian:
Here we are. Finally. For ease of reviewing: 

1. commit 1 includes the actual installation of prettier, prettier config changes, and .eslintrc changes to get the prettier and eslint to play nicely together.
2. commit 2 is solely formatting changes from running prettier.

One question going forward is if we might want to run eslint/prettier as a commit hook in the future. 

Thanks all, I'm sure this will speed up and improve our developing!


.prettierrc.json
```diff
@@ -0,0 +1,4 @@
+{
+  "singleQuote": true,
+  "trailingComma": "es5"
+}
```
richardxia, line 4: I would totally be cool with not overriding either of these and just using the default values. For code style, I generally prefer to do as little customization as possible so that the code style is as similar to other projects that use the tool as well. If these conflict with our ESLint rules, I prefer we disable/change the ESLint configuration over changing the Prettier settings, since it's trivial to fix Prettier issues (just run Prettier on the code), while not all ESLint issues have automatic fixes available, since they're not all about code formatting.

What do you think?

richardxia (approved):
💯!! Thanks for doing this! I'm overall really happy with this, and my only comment is to suggest sticking with the default Prettier settings, changing the ESLint ones if they conflict.

> One question going forward is if we might want to run eslint/prettier as a commit hook in the future.

I think I'm fine either way with this. Automated commit hooks may be a little bit tricky to do with our dev setup, since our preferred setup involves running the app in Docker, so in order to even have Prettier locally, you'd need to separately install the dependencies.

I think most editors, including VS Code, also have easy-to-install plugins for running Prettier on a project. Another benefit with the editor plugins is that you can get an even tighter feedback loop by enabling format-on-save, rather than only format-on-commit.

.prettierrc.json
schroerbrian: That is a good point. I think I set these more by force of habit. I have removed the custom configs!

PR #1221, opened 2023-02-06
Swap sffamilies white label logo for our415

drcaramelsyrup:
Swap the SFFamilies logo with Our415 logo. Soon we should consider changing the styling to match our415.org as well. At the least the logo swap is necessary prior to the Our415 event on 2/11.

I'm going to keep the `sf-families.svg` asset for use with the partners logos, until we get confirmation that we should change it.
Pasted image 20230205161506

Also actually use the `logoLinkDestination` when set, instead of the `siteUrl` to make sure clicking the logo takes the user to the correct destination.


schroerbrian (approved):
LGTM! Leland is going to look into doing a deploy for this tonight.

PR #1226, opened 2023-02-15
Add slug to url in ucsf white label so that user can nav directly to …

schroerbrian:
…discovery form

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -3,3 +3,4 @@ build
 !.*.js
 cypress.config.ts
 node_modules
+.eslintrc.js
--- a/app/App.tsx
+++ b/app/App.tsx
@@ -201,7 +201,7 @@ export const App = () => {
               {/* UCSF white label paths */}
               <Route
                 exact
-                path="/find-services"
+                path="/find-services/:selectedResourceSlug"
                 component={UcsfDiscoveryForm}
               />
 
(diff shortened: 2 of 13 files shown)


.prettierignore
```diff
@@ -3,3 +3,4 @@ build
 !.*.js
 cypress.config.ts
 node_modules
+.eslintrc.js
```
richardxia, line 6: Hey @schroerbrian, just wondering, is there a reason you excluded this file from Prettier? I actually normally prefer to have various .js-based config files to be formatted by Prettier, since it helps ensure that they're always formatted neatly.

schroerbrian: There is a clash between prettier and eslint where prettier wants to remove quotes from all single word key names (like "camelcase") whereas eslint then complains about `Inconsistently quoted property 'camelcase' found.`  In a future PR, I think I will just disable eslints `quote-props` rule for the offending lines (there are only 2), if that sounds like a better solution?

richardxia: Oh, I see, yeah, I would prefer to just eliminate the ESLint rule so that we can run prettier on the .eslintrc.js file.

I did discover that Prettier actually has [consistent quoting as a configuration option](https://prettier.io/docs/en/options.html#quote-props), but I don't have a super strong opinion about changing it from the default of "as-needed".

PR #1225, opened 2023-02-13
Enable more languages for Our415 whitelabel

drcaramelsyrup:
* Vietnamese (vi), Arabic (ar), Russian (ru)
* Enable setting language codes per whitelabel with `enabledTranslations` (to replace `enableTranslation` bool)

PR #1227, opened 2023-02-16
Minor tweaks to whitelabel enabledTranslations property.

richardxia:
These are the minor changes that I suggested as a post-merge review comment in https://github.com/ShelterTechSF/askdarcel-web/pull/1225#pullrequestreview-1300787255.

- Change the spelling of the type to `readonly string[]`, which is equivalent
- Change the comment to a documentation comment, which allows it to show up on hover
- Remove an unnecessary nullness check, since the type cannot be null

Here's an example of what the documentation comments look like in my editor of choice (vim), but it should be similar in other modern editors:
Screen Shot 2023-02-15 at 10 28 50 PM

### DIFF
--- a/app/components/ui/Translate.tsx
+++ b/app/components/ui/Translate.tsx
@@ -9,7 +9,7 @@ const Translate = () => {
   const { search } = useLocation();
 
   const languages = whiteLabel.enabledTranslations;
-  if (languages && languages.length > 0) {
+  if (languages.length > 0) {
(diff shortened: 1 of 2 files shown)


schroerbrian (approved):
LGTM!

drcaramelsyrup:
Thanks for the cleanup work!

drcaramelsyrup (approved):

richardxia:
Thanks, everyone!

PR #1228, opened 2023-02-17
Ucsf UI improvements

schroerbrian:
This PR adds code to select "See All" checkboxes by default on the UCSF Discovery Form page, per product discussion -- the eligibility and subcategory refinements state vars are initialized with a hash that has all See All refinements selected. 

This PR also implements floating buttons on the UCSF Discovery Form page per the [Figma designs](https://www.figma.com/file/HUTUyVuJf8lctj5fAFIY32/UCSF?node-id=5626%3A10024&t=pdo86f0CTBULiX2V-0) and product discussion.

Screen Shot 2023-02-16 at 8 51 02 PM

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,3 +1,5 @@
+// Disabling eslint quote props because prettier clashes with eslint here
+/* eslint-disable quote-props */
 const path = require("path");
 
 module.exports = {
@@ -20,7 +22,7 @@ module.exports = {
     },
   },
   rules: {
-    "camelcase": "off",
+    camelcase: "off",
     "import/extensions": [
       "error",
       {
@@ -144,7 +146,7 @@ module.exports = {
       rules: {
         "newline-per-chained-call": "off",
         "lines-between-class-members": "off",
-        "indent": "off",
+        indent: "off",
       },
     },
     // Node.js scripts
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,5 +1,3 @@
-// Disabling eslint quote props because prettier clashes with eslint here
-/* eslint-disable quote-props */
 const path = require("path");
 
 module.exports = {
@@ -174,10 +172,6 @@ module.exports = {
         // CJS-style imports.
         // https://github.com/benmosher/eslint-plugin-import/issues/1469
         "import/no-unused-modules": "off",
-        // Most ESLint rules names have characters which must be quoted, such as
-        // '-' or '/', so it's easier to read if all things under "rules" are
-        // consistently quoted.
-        "quote-props": ["error", "consistent-as-needed"],
       },
     },
   ],
--- a/app/components/ucsf/RefinementLists/SubcategoryRefinements.tsx
+++ b/app/components/ucsf/RefinementLists/SubcategoryRefinements.tsx
@@ -13,11 +13,11 @@ export interface SelectedSubcategories {
   [key: number]: boolean;
 }
 
-export const defaultSelectedSubcategories: Readonly<SelectedSubcategories> = {
-  "-1": true,
-};
-
 const seeAllPseudoId = -1;
+
+export const defaultSelectedSubcategories: Readonly<SelectedSubcategories> = {
+  [seeAllPseudoId]: true,
+};
 
 export const SubcategoryRefinements = ({
   subcategories,
--- a/app/components/ucsf/RefinementLists/SubcategoryRefinements.tsx
+++ b/app/components/ucsf/RefinementLists/SubcategoryRefinements.tsx
@@ -9,9 +9,13 @@ interface SubcategoryRefinement {
   id: number;
 }
 
-interface SelectedSubcategories {
+export interface SelectedSubcategories {
   [key: number]: boolean;
 }
+
+export const defaultSelectedSubcategories: Readonly<SelectedSubcategories> = {
+  "-1": true,
+};
 
 const seeAllPseudoId = -1;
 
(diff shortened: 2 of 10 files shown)


app/pages/UcsfDiscoveryForm/UcsfDiscoveryForm.tsx
```diff
@@ -155,18 +158,23 @@ const Page = () => {
...
-          </Button>
-          <Button
-            addClass={styles.nextBtn}
-            onClick={() => {
-              goToNextStep(selectedResourceSlug);
-            }}
-          >
-            {nextButtonText}
-          </Button>
+        <div className={styles.buttonBarBorder}>
```
richardxia, line 161: My reading of the Figma design shows that the button bar should be completely outside of the `.discoveryFormPage`. In the Figma document, you can see that the bar extends to the full width of the browser's viewport. Here, because `.discoveryFormPage` only spans columns 2 through 11 of the page grid, the button bar isn't wide enough. In order to achieve the same layout as the Figma document, I think you need the button bar to be a sibling of the `.grid` (Layout) element.

app/styles/utils/_layoututils.scss
```diff
@@ -20,6 +20,7 @@ $header-height: 70px;
 $header-mobile-height: 50px;
 
 // Z-Index values
+$z-index-button-bar: 1;
```
richardxia, line 23: This comment might not be relevant if you pull the button bar out to be a sibling of the Layout component, but I wanted to leave this comment in case you didn't want to do that in this PR and/or for educational purposes.

We probably shouldn't be defining this constant in this file because the element that you are applying this to is not in the same [stacking context](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Positioning/Understanding_z_index/The_stacking_context) as the others, and therefore has no relation to the other z-index values described below. z-indexes aren't actually global; they're only relative to other elements within the same stacking context, and, as the name implies, there can be multiple stacking contexts within a page. As a concrete example, even if you set this constant to a very large z-index value, it would still stack below the `$z-index-modal: 5` constant, since the entire stacking context is still below a z-index of 5.

My mental heuristic for knowing whether an element is in a different stacking context is whether any ancestor of the element has been taken out of the [normal flow](https://developer.mozilla.org/en-US/docs/Learn/CSS/CSS_layout/Normal_Flow) of the layout engine. Absolutely/relatively positioning an element will take that element out of normal flow, and flexbox/grid will take the children of an element out of normal flow.

richardxia (commented):
Aside from my comment about the width of the button bar not matching the Figma design, this PR mostly LGTM!

I do actually have one remaining comment, but it's more about the design itself. I'll write it here in my PR review right now, but I'll copy it over to the Figma design so that we can get our designers to comment on it.

One issue I see with the sticky bar at the bottom is that it may cut off content in a way that does not make it obvious that there is more content. In both your screenshot and the Figma document, the button bar partially overlaps a checkbox item, which makes it obvious that a user should scroll to see more info, but by pure chance, here's what it looks like on my browser:

Screen Shot 2023-02-20 at 9 23 48 PM

There's quite a bit more content below what's visible on screen, since there's an entire 4th eligibility category, but it's not obvious that it's there because on a Mac, scroll bars don't appear (unless you have a mouse plugged in).

I've usually seen one of two solutions to this problem:

1. Always make scrollbars visible when there's more content. This goes against the Apple/Mac design guidelines, but it's arguably more accessible.
2. Add some kind of visual hint that there's more content, such as a drop shadow. Here's my crappy mockup that puts an inset box-shadow on the `.refinementsBox` element:

Screen Shot 2023-02-20 at 9 33 00 PM

Screen Shot 2023-02-20 at 9 33 08 PM

I'm sure a designer could come up with a more aesthetically pleasing version of this, but the general idea is that we need to make it visually obvious that we haven't seen the bottom of the content yet.

app/components/ucsf/RefinementLists/SubcategoryRefinements.tsx
```diff
@@ -9,10 +9,14 @@ interface SubcategoryRefinement {
...
-interface SelectedSubcategories {
+export interface SelectedSubcategories {
   [key: number]: boolean;
 }
 
+export const defaultSelectedSubcategories: Readonly<SelectedSubcategories> = {
+  "-1": true,
+};
+
 const seeAllPseudoId = -1;
```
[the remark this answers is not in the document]
schroerbrian: Ah! I had not known that one could use a var for a key name. Great to know, thanks!

app/pages/UcsfDiscoveryForm/UcsfDiscoveryForm.tsx
schroerbrian: Per our conversation, I've just removed the sticky button bar altogether. That said, I've left in button styling changes that I made to align with Figma design updates 
image

app/styles/utils/_layoututils.scss
schroerbrian: Since we decided to not float the button bar, I've removed the z-index const. But thank you for the explanation nonetheless!

.eslintrc.js
```diff
@@ -1,3 +1,5 @@
+// Disabling eslint quote props because prettier clashes with eslint here
+/* eslint-disable quote-props */
```
richardxia, line 2: Instead of disabling this with a comment up here, shouldn't we just turn off the configuration value at the bottom of this file? There's a section dedicated to just the lint settings just for this file, so it doesn't make sense to turn on a rule only for this file only to disable it via a comment :)

richardxia (approved):
LGTM!

.eslintrc.js
schroerbrian: ohhh haha! i did not know

PR #1229, opened 2023-02-23
BreakingNewsDraft

davidagustin:
No Issue was made for this ticket. Needs Cheveron icon component


package.json
```diff
@@ -10,6 +10,7 @@
   "dependencies": {
     "@babel/polyfill": "^7.4.4",
     "@sentry/browser": "^4.0.6",
+    "@typeform/embed": "^0.13.1",
```
richardxia, line 13: We removed the Typeform integration from our app a little while ago. Did you perhaps have a merge conflict in which you accidentally added this back in? Could you remove this dependency?

app/pages/HomePage/components/BreakingNews/BreakingNews.tsx
```diff
@@ -0,0 +1,43 @@
...
+  CarouselProvider,
+  Slider,
+  Slide,
+  ButtonBack,
+  ButtonNext,
+} from 'pure-react-carousel';
+
+import styles from './BreakingNews.module.scss';
+
+const BreakingNews = () => (
```
richardxia, line 12: I left a similar comment on https://github.com/ShelterTechSF/askdarcel-web/pull/1168#discussion_r1112604637, but could we name this NewsArticle to make it consistent between the frontend and backend? Or if it's acceptable, renaming the backend API routes and database table could also be acceptable. I just think it'll cause confusion to people reading the code if the frontend and backend call this something different.

app/pages/HomePage/components/BreakingNews/BreakingNews.module.scss
```diff
@@ -0,0 +1,61 @@
...
+  width: 90%;
+  border-radius: 8px;
+  border: solid 1px transparent;
+  box-shadow: 0 6px 16px rgb(0 0 0 / 20%);
+}
+
+.title {
+  font-weight: 600;
+  font-size: 42px;
+  @media screen and ( max-width: 767px ) {
```
richardxia, line 46: Let's use the constants for the breakpoints defined in https://github.com/ShelterTechSF/askdarcel-web/blob/5604e3f3618a2a7c2c5107b7d20bdf2da9c451c4/app/styles/utils/_layoututils.scss#L4-L17 rather than hardcoding magic numbers in this file.

richardxia (commented):
I see that this is still an early draft, but here are some initial comments that I have. Thanks for getting this started!

schroerbrian:
Starting work on this per discussion with @davidagustin

schroerbrian:
Thanks @davidagustin for getting this started! This is now ready for review.

app/components/ui/NewsArticles/NewsArticles.tsx
```diff
@@ -0,0 +1,141 @@
...
+  >([]);
+
+  useEffect(() => {
+    const fetchBreakingNewsArticles = async () => {
+      const response = await dataService.get("/api/news_articles");
+      const { news_articles }: { news_articles: NewsArticle[] } = response;
+      setBreakingNewsArticles(news_articles);
+    };
+
+    fetchBreakingNewsArticles();
```
richardxia, line 65: Is there a reason for wrapping this in a function and immediately calling it, as opposed to just running the stuff in the body?

app/components/ui/NewsArticles/NewsArticles.tsx
```diff
@@ -0,0 +1,141 @@
...
+
+  if (breakingNewsArticles.length === 0) return null;
+
+  return (
+    <div
+      className={`${styles.container} ${
+        collapseCarousel && styles.carouselCollapsed
+      }`}
+    >
+      <div>
```
richardxia, line 76: Is there a purpose to this layer of `<div>`, or are you able to eliminate it?

app/components/ui/NewsArticles/NewsArticles.tsx
```diff
@@ -0,0 +1,141 @@
...
+                collapseCarousel
+                  ? styles.expandCarouselChev
+                  : styles.collapseCarouselChev
+              }`}
+            >
+              keyboard_arrow_down
+            </i>
+          </button>
+        </div>
+        {!collapseCarousel && (
```
richardxia, line 95: If you do the collapsing purely in CSS and not JS, then we can put in a nice animation for the collapsing that wouldn't be possible if you actually removed the DOM element. Here's an article that shows one approach of doing a CSS animation for a collapsible component: https://jdsteinbach.com/css/holy-grail-css-animation/

app/components/ui/NewsArticles/NewsArticles.tsx
```diff
@@ -0,0 +1,141 @@
...
+      <p className={styles.articleTitle}>
+        {article.headline} <span className={styles.dateSeperator}>|</span>{" "}
+        <span className={styles.articleDate}>
+          {formatDate(article.effective_date)}
+        </span>
+      </p>
+      <p className={styles.articleBody}>{article.body}</p>
+    </div>
+    {article.url && (
+      <a href={article.url} className={styles.articleLink}>
```
richardxia, line 44: I think we'd probably want these links to open in a new window, since they'll almost always link to an external site, right?

richardxia (commented):
Very nice and clean implementation!

I noticed a few visual oddities when checking this out locally. These might be a bit on the extreme side due to me putting kind of unusual data into the DB, but I think they should be relatively quick to fix.

Here are the two news articles I have locally:

Screen Shot 2023-04-08 at 8 26 39 PM

Screen Shot 2023-04-08 at 8 26 49 PM

The visual issues are:

1. When one news item is significantly larger than others, it causes the height of the whole NewsArticle container to match that large news item. This causes the left and right arrows to look misaligned with the shorter news articles. I think this can be fixed by ensuring that the white box of the shorter news articles is as tall as the tallest one.
2. In the larger news item, the "Read more ->" link is scrunched up. I think we just need to make sure that that link is given as much horizontal space as it wants.
3. Lastly, I guess it matches the Figma designs, but it seems strange to me that on mobile, we don't render the "Read more ->" link, so the only info we can communicate is the very short article heading. I'm a little bit concerned that if we don't have any way of viewing the larger description or the "Read more ->" link, that when an internal ShelterTech user modifies the Breaking News content, if they're using a desktop/laptop, they won't realize that very little of the content is actually viewable on mobile.

app/components/ui/NewsArticles/NewsArticles.tsx
```diff
@@ -0,0 +1,141 @@
...
+
+interface NewsArticle {
+  id: string;
+  headline: string;
+  effective_date: string;
+  expiration_date: string;
+  body: string;
+  priority: number;
+  url?: string;
+}
```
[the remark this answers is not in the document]
schroerbrian: Done. I added that to the Meta file, whereas https://github.com/ShelterTechSF/askdarcel-web/pull/1168 has it in its own file, but I can fix it up when 1168 is merged in. I will make a comment on #1168 's PR

app/components/ui/NewsArticles/NewsArticles.tsx
schroerbrian: Good call, updated.

app/components/ui/NewsArticles/NewsArticles.tsx
richardxia: Ah, I see, yeah, that makes sense. I haven't used `async` much myself, so I wasn't familiar with this pattern of using async inside of React hooks.

app/components/ui/NewsArticles/NewsArticles.module.scss
```diff
@@ -0,0 +1,277 @@
...
+}
+
+.carouselContainer {
+  max-width: 1160px;
+  margin: auto;
+  width: 90%;
+  margin-bottom: 9px;
+  opacity: 1;
+  transform: scale(1);
+  animation: grow 0.3s ease-in-out;
```
richardxia, line 68: This renders weirdly on my computer (macOS with Chrome 112). When I collapse the breaking news component, it collapses instantaneously, with no animation, but when I expand it, it expands with kind of a weird animation where the gray background instantly appears, but then the white boxes grow gradually with an animation.

I'm not too familiar with animation keyframes, but the approach I've used for collapsing animations in the past was the `max-height` + `transition: max-height ...` styles that the article I linked to mention.

Not sure if you're experiencing something different on your computer, but I just wanted to point it out.

app/components/ui/NewsArticles/NewsArticles.tsx
```diff
@@ -0,0 +1,148 @@
...
+
+    fetchBreakingNewsArticles();
+  }, []);
+
+  if (!breakingNewsArticles || breakingNewsArticles.length === 0) return null;
+
+  return (
+    <div
+      className={`${styles.container} ${
+        collapseCarousel && styles.carouselCollapsed
```
richardxia, line 80: While looking at the dynamic CSS classes on the actual HTML, I noticed that this literally renders as `false` in the class attributes. I think you may need to turn this into a ternary expression where you explicitly return the empty string if the condition is false.

richardxia (approved):
Thanks for addressing my comments! I have one last comment about the animation looking kind of weird, at least on my computer, but I think it's the last thing that needs to be fixed. Thanks for getting this through to the finish line!

app/components/ui/NewsArticles/NewsArticles.tsx
schroerbrian: Ahhh I hadn't thought of that -- good callout on that one

app/components/ui/NewsArticles/NewsArticles.module.scss
schroerbrian: I went into a little animation hole and added some transitions using max-height and opacity. Now the entire Breaking News element animates and it looks smoother. I also added an animation for when the element initially appears on screen after the data has been fetched; I used a class rather than returning null if there are no news articles. See animation here: [BreakingNewsDraft](https://www.loom.com/share/9a89fd614bc94322b22812740bd4a3d5)
I will demo this at the product meeting tomorrow to see what folks think about the animations and can adjust as needed before going to prod.

richardxia: Awesome, that looks great to me!

PR #1230, opened 2023-02-23
Convert Edit page files to .tsx with weak types.

richardxia:
Refs #1175

This converts all of the Edit page JSX files to TSX, but without adding the necessary type annotations everywhere. Instead, this opts out of lint rules and even downgrades the tsconfig settings in order to make minimal code canges.

Note that the tsconfig.json changes to disable `noImplicitAny` affect the whole codebase; there's no way to change the tsconfig settings for a subset of files without creating a separate tsconfig project. This will only be temporarily enabled until we add stronger types to the Edit page code.

I'm planning to immediately work on adding stricter types after this merges. I just think doing this first will make the work easier, since some types are hard to figure out without working on multiple layers of these components at the same time. For example, the OrganizationEditPage may pass a prop down to EditServices, which may pass it down to ProvidedService, which may pass it down to EditAddresses. Being able to work on a set of props at a time and add type annotations on it all the way up and down the stack would be easier than working on a file at a time (and believe me, I have some stashed commits where I tried doing that first!).


schroerbrian (approved):
Looks great. I'm a fan of doing this piece by piece to get the momentum going. Thanks!

richardxia:
Thanks for the review! I'll probably start submitting PRs to gradually improve the typing on these files in the next few days!

PR #1231, opened 2023-02-23
Run Prettier in CI.

richardxia:
I noticed that Prettier doesn't seem to actually be checked in CI, since `npm run lint` only runs ESLint, not Prettier. This adds new `prettier` and `prettier:fix` scripts to our package.json, and then it adds `npm run prettier` to our CI job.


schroerbrian (approved):
Oops thanks -- i had been meaning to add this!

richardxia:
Whoops, I should have merged my PRs in the opposite order. I am now causing the `master` branch to fail lint now. I'll send a fix PR right away.

PR #1233, opened 2023-02-24
Pass language param to pdf endpoint

schroerbrian:


app/pages/Pdf/ServicePdfPage.tsx
```diff
@@ -45,9 +50,23 @@ export const ServicePdfPage = () => {
...
+          .replace(/;}/g, "}")
+          .replace(/\/\*.*?\*\//g, "");
+      }
+
       // Map image has been fetched. We're now ready to make our request to the PDF conversion endpoint
       fetch(`/api/services/html_to_pdf`, {
-        body: JSON.stringify({ html: ref.current?.outerHTML }),
+        body: JSON.stringify({
+          html: ref.current?.outerHTML,
+          target_language: targetLangCode ?? null,
```
richardxia, line 68: I think this `?? null` doesn't do anything, since it looks like [`URLSearchParams.get()`](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams/get) will already return `null` if the parameter doesn't exist. `?? null` would also catch `undefined` and turn it into a `null`, but it doesn't actually look like it's possible for it to be `undefined` in the first place.

app/pages/Pdf/ServicePdfPage.tsx
```diff
@@ -111,14 +130,13 @@ export const ServicePdfPage = () => {
...
       )}
-      ;
       {!pdfSource && (
         <div
           ref={ref}
           className={`serviceHandoutPdf ${styles.handoutInvisible}`}
         >
           {/* The below styles contain some overrides of global .renderedMarkdown and table tag selector styles */}
-          <style>
+          <style translate="no" ref={styleRef}>
```
richardxia, line 139: I'm laughing a little because I'm imagining how you discovered that we need `translate="no"` here 😂.

richardxia (approved):
LGTM!

PR #1234, opened 2023-02-24
Fix Prettier errors.

richardxia:
This was caused by several PRs merging while the PR that turned on Prettier in CI was in flight.


richardxia:
Whoops, autosquash worked differently than what I am used to from my day job. Well, I accidentally merged this in even before CI ran, but hopefully that passes.

PR #1235, opened 2023-02-26
Migrate Edit Page to TypeScript Part 2

richardxia:
Refs #1175.

This deals with some of the lower-hanging fruit on the Edit Page, where either the type is totally known due to the values coming purely from the frontend (as opposed to the backend API), such as event handlers or `setState()` being called with known keys on the object. Notably, this adds a type for the props of the main page component as well as one for the state of that component, though the current version of the state type definition has a lot of `any`'s in it.

I've been cautious about not using the types defined in `app/models` until I can confirm that they actually match the Edit Page, since the `app/models` types mostly reflect what the API sends back, but the Edit Page performs a lot of transformations internally. For now, leaving the types as `any` or even `Record<any, any>` makes it a bit more typesafe since we at least acknowledge the existence of certain properties.

One other thing to note is that I've added an `assertDefined()` helper function for purely type checking purposes. It's similar to the `!` (postfix) operator in TypeScript, which will promote a type of `T | undefined` to just `T`, but my version is real JavaScript, so it performs an actual runtime check, which makes it a bit safer than the `!` operator. This was mostly used to deal with `this.state.resource`, which is only initialized after the `componentDidMount()` hook fires, which makes a request to the API endpoint to get the existing resource data.

### DIFF
--- a/app/components/edit/FormTextArea.tsx
+++ b/app/components/edit/FormTextArea.tsx
@@ -1,7 +1,12 @@
-import React, { Fragment } from "react";
-import PropTypes from "prop-types";
+import React from "react";
 
-const FormTextArea = ({ label, placeholder, value, setValue }) => (
+type Props = {
+  label: string;
+  placeholder: string;
+  value: string;
+  setValue: (x: string) => void;
+};
+const FormTextArea = ({ label, placeholder, value, setValue }: Props) => (
   <>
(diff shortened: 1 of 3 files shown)


schroerbrian (approved):
Looks great! Thanks for helping to usher this work along, especially for the edit page

richardxia:
Thanks for the speedy reviews!

PR #1236, opened 2023-03-01
Migrate Edit Page to TypeScript Part 3

richardxia:
Refs #1175.

This is the next unit of work I've done in terms of migrating the Edit Page to TypeScript. Here, I've defined types for the internal versions of Schedule, ScheduleDay, Organization (Resource), and Service, which differ from the versions that the API endpoints respond with. I won't detail the differences in this PR description, but I wrote comments next to the type definitions explaining the differences. The short summary is that the Edit Page has to keep track of certain state in its UI, so it can't maintain that state using the exact same shape of the data as stored in the DB.

These internal data type aren't great, and if I were to rewrite this page from scratch, I would do it very differently, but I am glad that TypeScript has allowed me to at least annotate exactly what things are like today.

One other notable change is that for some of the `<input>` fields directly on the Organization, I had to add a `resource.some_field ?? ""` to places where we were setting the `defaultValue` prop on the `<input>` element. This was because the builtin type definitions say that `defaultValue` cannot be `null`, but some of our resource fields can be `null`. I did a bit of debugging in a live browser, and it looks like if you pass in a `null`, the net effect is that it basically gets treated as the empty string, so I decided to explicitly coerce them to the empty string.

I did some light testing of this locally, and I think it works, but I can't honestly say that I've done any comprehensive testing, since there's so many possible scenarios to test.

### DIFF
--- a/app/components/edit/ProvidedService.tsx
+++ b/app/components/edit/ProvidedService.tsx
@@ -1,4 +1,4 @@
-import React, { Fragment } from "react";
+import React from "react";
 import PropTypes from "prop-types";
 import EditNotes from "./EditNotes";
 import EditSchedule from "./EditSchedule";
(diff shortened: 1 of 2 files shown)


schroerbrian (approved):
Looks great. Thanks for adding the comprehensive comments!

richardxia:
Thank _you_ for reviewing the changes!

PR #1238, opened 2023-03-08
Include legal status in create resource request

schroerbrian:
This also disables the UI for adding a service on a new resource. I looked into allowing users to add services with a new resource. It didn't work off the bat. Thus, I figured this was never allowed, rather than whacking through and trying to fix it. Our code is already using `!newResource` as [condition for displaying the services list](https://github.com/ShelterTechSF/askdarcel-web/blob/23d2e043ae2c218a4462b0da6460faf3d38db93f/app/pages/OrganizationEditPage.tsx#L1590-L1596), which also implies it was never intended.

https://github.com/ShelterTechSF/askdarcel-web/blob/23d2e043ae2c218a4462b0da6460faf3d38db93f/app/pages/OrganizationEditPage.tsx#L1590-L1596

image

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -87,7 +87,10 @@ module.exports = {
     // Temporarily disable certain ESLint rules for Edit page files until we
     // improve the type annotations.
     {
-      files: ["app/components/edit/*.tsx", "app/pages/OrganizationEditPage.tsx"],
+      files: [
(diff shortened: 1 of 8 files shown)


richardxia (approved):
Thanks for digging into the add services button when creating new resources! LGTM!

PR #1242, opened 2023-03-09
Add GA4 Google Analytics tag to tracking code

schroerbrian:
Our `react-ga` package [has no plans to support the new GA 4 property](https://github.com/react-ga/react-ga/issues/460), so I needed to use a package that does. Come July when Google sunsets its UA tracking ,we can remove the `react-ga` package and its references.


richardxia (approved):
LGTM!

Issue #1177, opened 2022-10-10
Redo global and base CSS styles


richardxia:
When this project was first created in 2016, we had a base set of CSS files, mostly under `app/styles`. The web, CSS, and frontend tooling have evolved a lot since then, and many of the practices in those files are quite outdated, including:

- Avoiding the use of flexbox, which was not supported widely enough by enough browsers back then (mostly IE's fault)
- Making very heavy use of advanced Sass features such as mixins, includes, and extensions
- Setting very specific styles on generic tag selectors, which are global and often cause us to have to override those specific styles for new pages

I'd like to propose a project for significantly reworking the CSS on the site. I'll first describe the end goal, and then subsequently I'll describe a migration path.

## End Goal

- Very minimal global CSS, which may not be much more than picking a reset stylesheet and setting some useful global constants (e.g. fonts, colors, etc.).
- All component- or page-specific styles go into CSS modules
- Use standard CSS features when possible, making use of transpilers/polyfills for CSS features that may not be supported in browsers today. The nice thing about using standard features is that the polyfills can go away in the future without changing the code today.
- Have a saner convention for managing common values. I am possibly interested in [Tailwind CSS](https://tailwindcss.com/), but even just defining an enum-style set of constants could be enough here. The sheltertech.org global [CSS file](https://github.com/ShelterTechSF/sheltertech.org/blob/c12c90bbff7d0f8370ecd35d1bdec534a1c9f53a/src/stylesheets/global.css) basically only contains variables and shows how this could look like.

## Migration path

1. Contain the current global CSS styles so that they are no longer global. This could be done, for example, but turning every generic tag selector into a nested selector that depends on a `.legacy` class, and then individual pages can be opted in and out of those styles.
2. Introduce a new reset stylesheet and fix up existing pages and components that may need to explicitly set some new properties.
3. Componentize common components such as buttons and text boxes, making sure that they work with only the reset stylesheet.
4. Gradually migrate the pages that depend on the previously global CSS styles so that they are not dependent on them.

richardxia:
One thing we'll have to be careful about is the styling on links, especially their pseudo-classes like `:hover` or `:active`. See https://github.com/ShelterTechSF/askdarcel-web/pull/1245#discussion_r1141563784 for an example of where we had to explicitly override a global pseudo-class's styling for the text color.

[...]

PR #1245, opened 2023-03-17
Add report street crisis button to navigation.

tpetersen0308:
https://www.loom.com/share/b6d4a53274934960b01bd8209d9579ea

### DIFF
--- a/app/components/ui/HamburgerMenu.module.scss
+++ b/app/components/ui/HamburgerMenu.module.scss
@@ -13,8 +13,4 @@
   &.emphasized {
     font-weight: 600;
   }
-
-  &.hidden {
-    display: none;
-  }
 }
--- a/app/components/ui/HamburgerMenu.module.scss
+++ b/app/components/ui/HamburgerMenu.module.scss
@@ -13,4 +13,8 @@
   &.emphasized {
     font-weight: 600;
   }
+
+  &.hidden {
+    display: none;
+  }
 }
--- a/app/components/ui/HamburgerMenu.module.scss
+++ b/app/components/ui/HamburgerMenu.module.scss
@@ -9,4 +9,8 @@
     background: #276ce5;
     color: #ffffff;
   }
+
+  &.emphasized {
+    font-weight: 600;
+  }
 }
--- a/app/components/ui/HamburgerMenu.tsx
+++ b/app/components/ui/HamburgerMenu.tsx
@@ -22,8 +22,18 @@ const burgerStyles = {
   },
 };
 
-const links = [
-  { to: "/", text: "Home", exact: true },
+const links: {
+  to: string;
+  text: string;
+  exact?: boolean;
+  external?: boolean;
+  linkStyles?: string;
+}[] = [
+  {
+    to: "/",
+    text: "Home",
+    exact: true,
+  },
   { to: "/about", text: "About" },
   { to: "https://help.sfserviceguide.org", text: "FAQ" },
   {
@@ -38,6 +48,14 @@ const links = [
   { to: "https://twitter.com/sheltertechorg", text: "Twitter", external: true },
   { to: "/terms-of-service", text: "Terms of Service" },
   { to: "/privacy-policy", text: "Privacy Policy" },
+  {
+    to: "https://sf.gov/information/reporting-concerns-about-street-crises-and-conditions",
+    text: "Report Street Crisis",
+    external: true,
+    linkStyles: `${styles.menuItem} ${styles.emphasized} ${
+      !showReportCrisis && styles.hidden
+    }`,
+  },
 ];
 
 export const HamburgerMenu = ({
@@ -62,28 +80,25 @@ export const HamburgerMenu = ({
     styles={burgerStyles}
     width="275px"
   >
-    {links.map(({ to, text, exact = false, external = false }) => (
-      <MenuItem
-        key={to}
-        to={to}
-        onClick={toggleHamburgerMenu}
-        exact={exact}
-        linkStyles={styles.menuItem}
-        external={external}
-      >
-        {text}
-      </MenuItem>
-    ))}
-    {showReportCrisis && (
-      <MenuItem
-        to="https://sf.gov/information/reporting-concerns-about-street-crises-and-conditions"
-        onClick={toggleHamburgerMenu}
-        exact={false}
-        linkStyles={`${styles.menuItem} ${styles.emphasized}`}
-        external={true}
-      >
-        Report Street Crisis
-      </MenuItem>
+    {links.map(
+      ({
+        to,
+        text,
+        exact = false,
+        external = false,
+        linkStyles = styles.menuItem,
+      }) => (
+        <MenuItem
+          key={to}
+          to={to}
+          onClick={toggleHamburgerMenu}
+          exact={exact}
+          linkStyles={linkStyles}
+          external={external}
+        >
+          {text}
+        </MenuItem>
+      )
     )}
   </Menu>
 );
@@ -114,7 +129,7 @@ const MenuItem = ({
     </a>
   ) : (
     <NavLink
-      className={styles.menuItem}
+      className={linkStyles}
       activeClassName={styles.active}
       to={to}
       onClick={onClick}
--- a/app/components/ui/HamburgerMenu.tsx
+++ b/app/components/ui/HamburgerMenu.tsx
@@ -48,15 +48,16 @@ const links: {
   { to: "https://twitter.com/sheltertechorg", text: "Twitter", external: true },
   { to: "/terms-of-service", text: "Terms of Service" },
   { to: "/privacy-policy", text: "Privacy Policy" },
-  {
+];
+
+if (showReportCrisis) {
+  links.push({
     to: "https://sf.gov/information/reporting-concerns-about-street-crises-and-conditions",
     text: "Report Street Crisis",
     external: true,
-    linkStyles: `${styles.menuItem} ${styles.emphasized} ${
-      !showReportCrisis && styles.hidden
-    }`,
-  },
-];
+    linkStyles: `${styles.menuItem} ${styles.emphasized}`,
+  });
+}
 
 export const HamburgerMenu = ({
   isOpen,
--- a/app/components/ui/HamburgerMenu.tsx
+++ b/app/components/ui/HamburgerMenu.tsx
@@ -1,7 +1,12 @@
 import React, { ReactNode } from "react";
 import { NavLink } from "react-router-dom";
 import { push as Menu, State } from "react-burger-menu";
+import whiteLabel from "../../utils/whitelabel";
 import styles from "./HamburgerMenu.module.scss";
+
+const {
+  showReportCrisis,
+} = whiteLabel;
 
 const burgerStyles = {
   bmBurgerButton: {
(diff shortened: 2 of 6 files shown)


app/components/ui/Navigation.module.scss
```diff
@@ -137,6 +139,10 @@
   display: inline-block;
   height: 24px;
   width: 24px;
+
+  @include r_min($break-tablet-s) {
+    display: none !important;
+  }
```
tpetersen0308, line 145: The above changes are just to adjust the responsive breakpoints, since this added some width to the nav content.

app/components/ui/Navigation.module.scss
```diff
@@ -172,3 +178,29 @@
...
+    color: white;
+    box-shadow: 0 4px 8px rgba(0, 0, 0, 0.2);
+  }
+  &:active {
+    box-shadow: 0 1px 2px 0 rgba(0, 0, 0, 0.15);
+  }
+  @media screen and (max-width: 600px) {
+    display: none;
+  }
+}
```
tpetersen0308, line 206: This is somewhat copied/pasted from the search button, with a few tweaks. Not sure if we want to extract it to a shared style or not. Do we use this button styling elsewhere much?

app/components/ui/HamburgerMenu.tsx
```diff
@@ -56,10 +59,26 @@ export const HamburgerMenu = ({
...
     ))}
+    {showReportCrisis && (
+      <MenuItem
+        to="https://sf.gov/information/reporting-concerns-about-street-crises-and-conditions"
+        onClick={toggleHamburgerMenu}
+        exact={false}
+        linkStyles={`${styles.menuItem} ${styles.emphasized}`}
+      >
+        Report Street Crisis
+      </MenuItem>
```
richardxia, line 80: I'm wondering if it'd be better to (conditionally) add this to the `links` array and just add extra properties for the things that are special about this item. This would keep this component data-driven and more consistent with the other link items we have.

app/components/ui/Navigation.module.scss
```diff
@@ -172,3 +178,29 @@
...
+
+.buttonLink {
+  appearance: button;
+  text-decoration: none;
+  color: white;
+  background-color: $color-brand;
+  border-radius: 3px;
+  font-size: 16px;
+  font-weight: 600;
+  height: 60px;
```
richardxia, line 190: I see that this matches the height in the Figma design, but I also noticed that in the Figma design, the header is now 100px tall instead of the 70px we have today, and this makes a huge difference in terms of the perceived size of the button relative to the whitespace. Could you increase the height of that header to match the Figma design?

app/components/ui/HamburgerMenu.tsx
richardxia: On a separate note, it looks like Dongyeon responded to your question on Figma, and he does think that these links should open in another window/tab, so we probably need to add an extra property to control that.

app/components/ui/Navigation.module.scss
```diff
@@ -172,3 +178,29 @@
...
+  width: 200px;
```
richardxia, line 191: I generally avoid setting both the height and width of elements, since that can cause undesirable behavior when the element's content is larger than height and/or width. Even if it looks fine right now, we might change the text later, or it may look wrong on a different browser/OS that has a different text rendering algorithm.

For buttons, I normally like controlling the height via padding, not height, which allows it to look natural in the case where you decide to allow the text in the button to wrap to multiple lines. I also usually like primarily controlling the width via padding as well, although sometimes it's nice to set a max and/or min width.

app/components/ui/Navigation.module.scss
```diff
@@ -172,3 +178,29 @@
...
+  display: flex;
+  justify-content: center;
+  align-items: center;
```
richardxia, line 194: You may not need these if we decide to using both vertical and horizontal padding, since that will naturally center the text in both dimensions. If we do still want to use flexbox, then we should at least make this an `inline-flex` so that it behaves like a normal button when inside of [normal flow](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Flow_Layout) (e.g. a non-flexbox, non-grid, non-table element).

app/components/ui/Navigation.module.scss
```diff
@@ -172,3 +178,29 @@
...
+  &:hover {
+    color: white;
+    box-shadow: 0 4px 8px rgba(0, 0, 0, 0.2);
+  }
+  &:active {
+    box-shadow: 0 1px 2px 0 rgba(0, 0, 0, 0.15);
+  }
+  @media screen and (max-width: 600px) {
+    display: none;
+  }
```
richardxia, line 205: We probably don't want this, right? It feels strange to me to hide a button via the button's own styles through a media query on the page width, since I would normally think that only "layout-level" components should really be doing stuff like that. It doesn't really matter in this case because we already switch to the hamburger menu at <= 600px, but I think this would get in the way if we wanted to reuse these styles in other contexts.

app/components/ui/Navigation.module.scss
richardxia: Yeah, I think it'd be good to extract this into a shared style, but it doesn't need to happen in this PR, since we're already duplicating it in several places, probably with subtle and possibly unintentional differences between them. Outside of the search button near the bottom of the home page, I know that it's also present on the Resource Guides (ServiceDiscoveryForm in the code) and the Edit Pages, and I'm sure we have buttons elsewhere too.

app/components/ui/Navigation.module.scss
```diff
@@ -172,3 +178,29 @@
...
+  font-size: 16px;
+  font-weight: 600;
+  height: 60px;
+  width: 200px;
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  &:focus,
+  &:hover {
+    color: white;
```
richardxia, line 197: ~~I'm guessing this is one of the pieces that you said was copied and pasted from the Search Button, but this line doesn't have any effect since the non-pseudoclass styles already set it to white~~ Nevermind, I thought this wasn't necessary, but I forgot we have some global styles that we keep on having to override. I'm still going to post this comment, even though there's no action to take right now, but I just wanted something to link to in https://github.com/ShelterTechSF/askdarcel-web/issues/1177, which is a ticket I created somewhat recently to clean up our global CSS styles.

richardxia (commented):
Thanks for quickly getting this out! This looks pretty good to me, but I think the two main blocking items are to increase the header size and to make the links open in a new window/tab.

Most of my other comments, like putting the new button link in the `links` array or the nitty-gritty CSS comments, are a bit more subjective, so feel free to ignore them if you have a different preference.

app/components/ui/Navigation.module.scss
schroerbrian: There is a $header-height constant in the our `_layoututils.scss` file that you can adjust this on.

app/utils/whitelabel.ts
```diff
@@ -98,6 +99,7 @@ services and re-entry programs.`,
   showHeaderQrCode: false,
   showMobileNav: true,
   showSearch: true,
+  showReportCrisis: true,
```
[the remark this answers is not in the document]
schroerbrian: Yep, I think that is definitely the case

schroerbrian (changes requested):
Looking great. I think defaulting `showReportCrisis` to false is the only major blocker that I see, beyond addressing Richard's comments. Thanks for the good work!

app/components/ui/Navigation.module.scss
tpetersen0308: I'll take a look at that!

app/utils/whitelabel.ts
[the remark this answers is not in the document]
tpetersen0308: Thanks, I wasn't sure which versions of the site these defaults were servicing

app/components/ui/HamburgerMenu.tsx
tpetersen0308: Sounds good. Should I update the other external links to open in a new tab as well?

app/components/ui/Navigation.module.scss
tpetersen0308: If we only want to change the header height when the button is present, will this constant be useful?

app/components/ui/Navigation.module.scss
tpetersen0308: I'll try to get it centered with padding 👍

app/components/ui/HamburgerMenu.tsx
richardxia: I think the Twitter and Facebook links would make sense to open in a new tab, but ones under the help.sfserviceguide.org domain maybe should stay in the same tab, since even though it takes you to an external site, it's still conceptually part of the same site? But I don't think I have a super strong opinion about this and could go either way.

app/components/ui/Navigation.module.scss
richardxia: If you're talking about the button not being present on most of the whitelabel sites, then I think it's probably fine to increase the header nav height on all the sites, even if there's no button, just for consistency. But this might be a question more for the design team.

app/components/ui/Navigation.module.scss
tpetersen0308: I think this was leftover from some tinkering, and I forgot to remove it when I got the higher-level responsiveness worked out

app/components/ui/HamburgerMenu.module.scss
```diff
@@ -9,4 +9,12 @@
...
     color: #ffffff;
   }
+
+  &.emphasized {
+    font-weight: 600;
+  }
+
+  &.hidden {
+    display: none;
+  }
```
tpetersen0308, line 19: This felt like a weird way to feature flag the button, but conditionally adding it to the links array wasn't working out. Was trying to concat it onto links if `showReportCrisis === true`, but it wasn't showing up. I haven't done react in a bit--is that something that needs to run in a lifecycle hook or something?

richardxia: Were you literally using the [`Array.concat()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/concat) method? That method doesn't mutate the original array, and instead returns a new one, so it'd have to be done in the same expression as the one where you assign to `links`.

If that doesn't work, then we're admittedly doing some weird stuff in JavaScript-land by having lots of top-level/import-time logic for the whitelabel, and I wonder if that's causing the issue you're seeing. If you're still hitting issues after making sure that you are in fact mutating/initializing the `links` array properly, then I wonder if moving the `links` array definition into the body of the HamburgerMenu component would avoid the import-time issues.

Failing both of those, then I think what you have now is probably fine for getting this feature out the door, but it'd be nice to create a ticket to investigate the issue further.

richardxia (approved):
Aside from maybe giving the `links` array thing one last shot, this LGTM! Thanks for working on this!

app/utils/whitelabel.ts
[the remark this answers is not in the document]
schroerbrian: Oops my bad for not elaborating. While we do want to set this as false for the `whiteLabelDefaults` -- which is sort of a base object designed to avoid repetition in our different white label objects -- we want to overwrite that value in the `SFServiceGuide` and in the `defaultWhiteLabel` const. The `SFServiceGuide` whitelabel maps to our prod site www.sfserviceguide.org and the `defaultWhiteLabel` is our dev environment. As this stands now, the button won't show up on our prod site.

Sorry this is a bit confusing, maybe we can make this code somehow more descriptive in a future ticket.

schroerbrian (approved):
Other than my comment about changing the whitelabel `showReportCrisis` in the `SFServiceGuide` and `defaultWhiteLabel` to `true`, this looks great! Approving preemptively. Thanks!

app/components/ui/HamburgerMenu.module.scss
tpetersen0308: Ahh dangit I forgot that concat is non-destructive. Let me try that again.

app/utils/whitelabel.ts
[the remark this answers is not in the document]
tpetersen0308: Ahh thanks, I think I got it right this time! Yeah, I think I was confused mostly by the multiple defaults. There may be a better was to control it by environment.

PR #1246, opened 2023-03-29
Add timeout to GA page view event to delay event trigger for a reason…

schroerbrian:
…able period of time so that our Helmet package will have time to update the page title.

I add a 500ms delay to our GA page view event code to give time for our page titles to resolve prior to sending the event.  I also updated our Helmet package to see if it would fix things. It did not but I decided updating it can't hurt so included it in this PR. We have already been using the forked version of Helmet.

I also added more descriptive page titles to our search pages. 

Lastly, I changed our whitelabel default title from AskDarcel, which is only displayed locally, to SF Service Guide. I think we may have outgrown this title, and it could be confusing for newer devs. Any objections here?

### DIFF
--- a/app/App.tsx
+++ b/app/App.tsx
@@ -88,7 +88,7 @@ export const App = () => {
 
         ReactGA.set({ page });
         ReactGA.pageview(page);
-      }, 500)
+      }, 500);
     });
   }, [history]);
 
--- a/app/App.tsx
+++ b/app/App.tsx
@@ -80,6 +80,9 @@ export const App = () => {
     ReactGA_4.initialize(config.GOOGLE_ANALYTICS_GA4_ID);
     return history.listen((loc) => {
       setTimeout(() => {
+        /* We call setTimeout here to give our views time to update the document title before
+           GA sends its page view event
+        */
         const page = loc.pathname + loc.search;
         ReactGA_4.send({
           hitType: "pageview",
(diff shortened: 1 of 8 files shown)


app/components/listing/ServiceDetails.tsx
```diff
@@ -16,14 +17,10 @@ export const ServiceDetails = ({ service }: { service: Service }) => {
       id={`service-${service.id}`}
       data-cy="service-list-item"
     >
-      {/* <div className="service--meta disabled-feature">
-        <p><ServiceCategory category={service.category} /></p>
-        <p>updated {service.updated_at}</p>
-      </div> */}
       <h2 className="service--header">
-        <a href={`/services/${service.id}`} translate="no">
+        <Link to={{ pathname: `/services/${service.id}` }} translate="no">
```
schroerbrian, line 21: During my GA debugging I found that clicking on this particular anchor link would trigger a refetch of the entire document; whereas using `Link to` does not. It did not solve our GA problem but is a good improvement nonetheless.

app/App.tsx
```diff
@@ -79,14 +79,16 @@ export const App = () => {
...
     return history.listen((loc) => {
-      const page = loc.pathname + loc.search;
-      ReactGA_4.send({
-        hitType: "pageview",
-        page,
-      });
-
-      ReactGA.set({ page });
-      ReactGA.pageview(page);
+      setTimeout(() => {
```
richardxia, line 82: Maybe add a comment here describing why this timeout is needed? It'll be helpful in case if anyone besides us two is reading this code.

richardxia (approved):
LGTM!

app/App.tsx
schroerbrian: Good call, thanks!

PR #1247, opened 2023-04-02
Migrate Edit Page to TypeScript Part 5

richardxia:
Refs #1175

Major Edit: The original version that I submitted as a PR here was incorrect and was causing the Edit Page to completely break. I've pushed up a new version to this branch, and I've rewritten this PR description to reflect the new changes.

This adds TypeScript type annotations to the ProvidedService component.

In order for the types to validate, this required modifying the definition of InternalFlattenedService to make the `schedule` property required. This appears to be true in practice, where either the `schedule` property is defined by the original API response, or it gets set by the frontend code when creating a new Service, so there shouldn't be a case where it is undefined.

This adds an additional runtime check to the `applyChanges()` method to assert that the final, flattened service indeed has a `schedule` property.

### DIFF
--- a/app/components/edit/EditServiceChildCollection.tsx
+++ b/app/components/edit/EditServiceChildCollection.tsx
@@ -31,7 +31,7 @@ export const EditServiceChildCollection = <T extends CollectionItem>({
   buttonText,
   propertyKeyName,
 }: {
-  initialCollectionData: any[];
+  initialCollectionData: any[] | undefined;
   handleCollectionChange: (field: string, value: T[]) => void;
   CollectionItemComponent: FC<ComponentProps>;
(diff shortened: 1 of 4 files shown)


schroerbrian (approved):
Looks great! Thanks again for continuing this project and sorry for the slight delay in reviewing this. 

Side note: to test this, I pulled it down and built out reactivation functionality for deactivated services. PR here: https://github.com/ShelterTechSF/askdarcel-web/pull/1249. Everything seems to be working smoothly

richardxia:
No worries, and thanks for doing some tests on this branch! I also did a little bit more extra testing myself, and everything seemed to be fine, so I'll be merging this now.

PR #1248, opened 2023-04-06
Unset banner content alignment in mobile view

candywang:
Changing the property `align-items: unset` in the mobile view top aligns the icon. However, if this is not the best implementation I will take a look at the resource provided again.

### DIFF
--- a/app/components/ui/Banner.module.scss
+++ b/app/components/ui/Banner.module.scss
@@ -29,6 +29,7 @@
   @media screen and (max-width: $break-tablet-s) {
     width: 96%;
     gap: 12px;
+    align-items: unset;
   }
 }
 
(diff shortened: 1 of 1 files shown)


richardxia:
Thanks for getting this out quickly! Sorry if there was ambiguity in the ticket description, but we actually wanted this for both desktop and mobile views. Could you make this work on desktop as well?

candywang:
@richardxia Does adding a top margin in the desktop view work to align the icon and title text?

richardxia (approved):
> @richardxia Does adding a top margin in the desktop view work to align the icon and title text?

Yeah, this should be fine. I think there may be a way to get it to align more naturally, i.e. with `vertical-align: baseline`, but it's tricky when the icon is in a different grid column than the text that we want it to be aligned with. This LGTM!

schroerbrian (approved):
LGTM!

PR #1249, opened 2023-04-07
Expose activate function on edit page when resource is deactivated

schroerbrian:
Built this out as a way to test https://github.com/ShelterTechSF/askdarcel-web/pull/1247 changes and to fix something that's been in the backlog.

Displays activate button if resource has been deactivated and enable re-activating a resource. 

I also made a couple small TS changes to the handleDeactivate function for greater type specificity. Let me know if they get in your way all and ill revert them.

image

### DIFF
--- a/app/pages/OrganizationEditPage.tsx
+++ b/app/pages/OrganizationEditPage.tsx
@@ -84,7 +84,18 @@ const applyChanges = (
       (item) => !deletedIds.has(item.id)
     );
   }
-  return transformedItems;
+
+  // This block of code allows us to assert that `schedule` is always defined on
+  // the flattened service while preserving the type information for TypeScript.
+  const newTransformedItems: InternalFlattenedService[] = [];
+  transformedItems.forEach((flattenedService) => {
+    const { schedule } = flattenedService;
+    if (schedule === undefined)
+      throw new Error("Expected flattened service to have a schedule");
+    newTransformedItems.push({ ...flattenedService, schedule });
+  });
+
+  return newTransformedItems;
 };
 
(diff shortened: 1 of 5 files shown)


app/pages/OrganizationEditPage.tsx
```diff
@@ -1329,17 +1329,17 @@ class OrganizationEditPage extends React.Component<Props, State> {
 
   handleDeactivation(type: "resource" | "service", id: number): void {
     const { history } = this.props;
-    let confirmMessage: string | null = null;
+    let confirmMessage: string | undefined;
     let path: string | null = null;
```
richardxia, line 1333: Looking at this code more carefully, we can actually avoid having a `| null` or `| undefined` in the types at all if we just throw an exception in the `if` and `else if` chains. If you add this after the `else if` block

```typescript
    } else {
      throw new Error(`Unexpected type: ${type}`);
    }
```

then both `confirmMessage` and `path` can have a type of just `string` and we don't need to set an initial value here. The reason this works is because we're proving to TypeScript that either the `type === "resource"` or the `type === "service"` branch has to be taken, and therefore both `confirmMessage` and `path` are guaranteed to be given a real value. This should also allow you to remove the `as` assertions on them later in this function.

If it's not too much trouble, could you make this change instead of the change you had here?

richardxia (approved):
LGTM, thanks for fixing this!

app/pages/OrganizationEditPage.tsx
schroerbrian: Good thinking. Done!

PR #1250, opened 2023-04-07
Use PDF generation for all PDFs. Small UI improvements

schroerbrian:
We are now going to generate PDFs for all UCSF services. I have set up usage alerts in a slack channel #google-cloud-alerts and will be monitoring usage closely. There is also a daily cap. I do not think UCSF will get too close to our limits for now.


richardxia (approved):
Nice, LGTM!

PR #1251, opened 2023-04-09
Migrate Edit Page to TypeScript Part 6

richardxia:
Refs #1175.

This migrates a couple of generic editing widgets to TypeScript, `InputField` and `MultiSelectDropdown`. Most of the diff is in the latter, and most of the diff is actually just comments from me trying to reverse engineer what was going on. It looks like we have an expected shape for the objects being passed to `MultiSelectDropdown`, and then `react-select` imposes its own requirements on the values passed to it, so a good amount of our component is bridging the two formats.


schroerbrian (approved):
Thanks for making these changes, LGTM!

PR #1256, opened 2023-05-01
Edit Page: Add strict type for ProvidedService's handleChange() function.

richardxia:
Refs #1175.

The new type for `handleChange()` requires that:

1. the field name actually matches the name of a real field on the InternalTopLevelService type, and
2. the type of the new value matches the type of the field on the InternalTopLevelService type.

This gives us much stricter type checking, since different fields of InternalTopLevelService have very different types, and this allows us to actually check those types against the values that we are passing in.

Several of the other changes in this commit are to pass type checking, now that we're more strictly enforcing calls to `handleChange()`. Two notable changes are that we've removed the `propertyKeyName` prop to `EditServiceChildCollection`, since it can be baked into the `handleCollectionChange` function, and slightly changing the type of the `editServiceById` prop on the `EditServices` componen to accept an `InternalTopLevelService` rather than an `InternalFlattenedService`.

### DIFF
--- a/app/components/edit/EditServiceChildCollection.tsx
+++ b/app/components/edit/EditServiceChildCollection.tsx
@@ -29,15 +29,13 @@ export const EditServiceChildCollection = <T extends CollectionItem>({
   label,
   blankItemTemplate,
   buttonText,
-  propertyKeyName,
 }: {
   initialCollectionData: any[] | undefined;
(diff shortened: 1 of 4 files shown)


schroerbrian (approved):
Part 7, wow! LGTM

richardxia:
Thanks, as always, for reviewing!

PR #1258, opened 2023-05-01
Add SFSeal footer to Our415, allow configuring whitelabel footer options

drcaramelsyrup:
Our415 has requested that we add the City of San Francisco seal as a footer on their listing pages. (They don't want any of the other stuff on there, like links to ShelterTech social media.) Verified default whitelabel looks the same on home page and listing pages.

As for Our415:

### DIFF
--- a/app/components/ui/Footer.scss
+++ b/app/components/ui/Footer.scss
@@ -1,3 +1,5 @@
+@import "../../styles/utils/_helpers.scss";
+
 .site-footer {
   display: flex;
   align-items: center;
@@ -51,7 +53,7 @@
   height: 100%;
   width: 1280px;
   margin: 0 auto;
-  @media screen and (max-width: 767px) {
+  @media screen and (max-width: $break-tablet-s) {
     width: auto;
   }
 }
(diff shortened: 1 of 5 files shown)


schroerbrian (approved):
LGTM! Thanks for verifying the standard SFSG whitelabel footer as well.

My only comment is that, for future reference, we have CSS constants for our breakpoints, including a 767px one, which are defined [here](https://github.com/ShelterTechSF/askdarcel-web/blob/master/app/styles/utils/_layoututils.scss). 

It looks like the Footer.scss was already using this 767px breakpoint, which you probably referenced for these changes. Since it was a quick change, and I had to look up where we define the consts anyway, I went ahead and pushed up changes to use the const in this file instead, hope that is cool!

drcaramelsyrup:
Perfectly fine, thank you :)

richardxia (approved):
LGTM! My only comment is that if we end up customizing the footer for other whitelabel sites any further, we might want to make the entire footer swappable, rather than adding more parameters to it.

drcaramelsyrup:
Thanks for your reviews! I'll try to get this one merged after another proofread this weekend.

PR #1259, opened 2023-05-02
Adjust listing page side icons to more closely match figma designs

schroerbrian:
Looking at the large side icons on listing pages finally drove me crazy, so I dug up an old [Figma design](https://www.figma.com/file/0X2KzS0qDEfZPRRm0TQyd5/SFSG-Website?type=design&node-id=1458-235&t=8TEfnDAk8dkge9k9-0) and updated the styles. There are many more differences between the design and our current listing page (and I don't know how current it is), but I find the side icons size to be so noticeable that I'd like to quickly fix them.

Screen Shot 2023-05-02 at 11 48 15 AM

I also fixed a UI bug where the service description margin wasn't aligning with the service title (too far to the left, see example below):
Screen Shot 2023-05-02 at 11 48 09 AM

### DIFF
--- a/app/components/listing/ActionBar.scss
+++ b/app/components/listing/ActionBar.scss
@@ -5,32 +5,35 @@
 
   li {
     text-align: center;
-    border: 1px solid $color-grey2;
-    width: 128px;
-    height: 128px;
+    border: 1px solid $color-grey3;
+    width: 110px;
+    height: 110px;
 
     a {
       color: $color-textfade;
-      padding: $padding-default 0;
+      padding-top: $padding-default;
       display: flex;
+      gap: 12px;
       flex-direction: column;
-      width: 128px;
-      height: 128px;
+      justify-content: center;
+      width: 100%;
+      height: 100%;
 
       img {
-        display: block;
-        margin: $padding-default $padding-default $padding-large;
-        font-size: $padding-xxlarge;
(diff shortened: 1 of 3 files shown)


app/components/listing/ActionBar.scss
```diff
@@ -1,36 +1,39 @@
 @import "../../styles/utils/helpers";
 
 .action-sidebar {
-  margin: $padding-xlarge 0;
+  margin-bottom: $padding-xlarge 0;
```
richardxia, line 4:
```suggestion
  margin-bottom: $padding-xlarge;
```

I think you meant to drop the second number as well, since `margin-bottom` only takes a single value, whereas before we were using the `margin` shorthand for declaring vertical and horizontal margin with two values.

Actually, do you understand why we declare a bottom margin in the first place? There doesn't seem to be anything below the action sidebar, and it looks like we have an entirely separate component for the mobile version, so this wouldn't apply to that. If this is unused, then let's delete it altogether.

richardxia (approved):
LGTM! Thanks for fixing this, it's been bothering me for a while, too!

app/components/listing/ActionBar.scss
schroerbrian: Ahhh good catch, there's no need for that rule. Thanks!

PR #1260, opened 2023-05-10
Complete search translation

schroerbrian:
API PR here: https://github.com/ShelterTechSF/askdarcel-api/pull/686

### DIFF
--- a/app/components/search/SearchResults/SearchResults.jsx
+++ b/app/components/search/SearchResults/SearchResults.jsx
@@ -221,32 +221,32 @@ const SearchResult = ({ hit, index, setCenterCoords }) => {
           {
             key: -2,
             description: "Spanish",
-            url: `${handoutApiUrl}?lang=es`,
+            url: `${handoutApiUrl}?handoutLanguage=es`,
           },
           {
             key: -3,
             description: "Tagalog",
-            url: `${handoutApiUrl}?lang=tl`,
+            url: `${handoutApiUrl}?handoutLanguage=tl`,
           },
           {
             key: -4,
             description: "Chinese (Traditional)",
-            url: `${handoutApiUrl}?lang=zh-TW`,
+            url: `${handoutApiUrl}?handoutLanguage=zh-TW`,
           },
           {
             key: -5,
             description: "Vietnamese",
-            url: `${handoutApiUrl}?lang=vi`,
+            url: `${handoutApiUrl}?handoutLanguage=vi`,
           },
           {
             key: -6,
             description: "Russian",
-            url: `${handoutApiUrl}?lang=ru`,
+            url: `${handoutApiUrl}?handoutLanguage=ru`,
           },
           {
             key: -7,
             description: "Arabic",
-            url: `${handoutApiUrl}?lang=ar`,
+            url: `${handoutApiUrl}?handoutLanguage=ar`,
           },
         ]}
       />
--- a/app/pages/SearchResultsPage/SearchResultsPage.tsx
+++ b/app/pages/SearchResultsPage/SearchResultsPage.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useState } from "react";
+import React, { useEffect, useRef, useState } from "react";
 import { useHistory, useLocation } from "react-router-dom";
 import { useCookies } from "react-cookie";
 import { Helmet } from "react-helmet-async";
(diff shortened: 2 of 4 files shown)


app/pages/SearchResultsPage/SearchResultsPage.tsx
```diff
@@ -34,22 +36,64 @@ interface SearchState extends ParsedQs {
...
+    } else {
+      setSearchState(queryParams);
+      setSearchStateResolved(true);
+    }
+  }, [cookies.googtrans, search, untranslatedQuery]);
+
+  if (!searchStateResolved) {
+    // In some case(s) -- e.g. when waiting to receive a translation from the API -- the search state
+    // is set asynchronously; thus, we want to wait until the search state resolves. Otherwise,
+    // our code will query Algolia twice - first with an empty string and secondly with the translated
```
richardxia, line 89: Not to nitpick grammar for grammar's sake, but did you mean "our code _would_ query Algolia twice"? As in, you're saying that if this code didn't have this early `return null`, then it would result in a UI flash and unnecessary call to Algolia?

Mostly asking because on my first read of these comments, I think I misread this as saying that we are intentionally acknowledging that we are going to cause a UI flash and unnecessary call to Algolia. Maybe you could reword this comment to be a bit more precise?

app/pages/SearchResultsPage/SearchResultsPage.tsx
```diff
@@ -34,22 +36,64 @@ interface SearchState extends ParsedQs {
...
+        }).then((resp) =>
+          resp.json().then((body) => {
+            setSearchState({
+              ...queryParams,
+              query: body.result,
+            });
+            setSearchStateResolved(true);
+          })
+        );
+      }
```
richardxia, line 79: Don't we need an `else` to go with this inner `if`? Otherwise, wouldn't `searchStateResolved` never get set to `true`, hanging the page?

app/pages/SearchResultsPage/SearchResultsPage.tsx
```diff
@@ -34,22 +36,64 @@ interface SearchState extends ParsedQs {
...
   const [searchRadius, setSearchRadius] = useState(
     searchState?.configure?.aroundRadius ?? "all"
   );
+  // In cases where we translate a non-English query into English (to query Alglolia), we use this
+  // state var to cache the user's orginal, untranslated query to display on the results page
+  const [untranslatedQuery, setUntranslatedQuery] = useState("");
+  const [searchStateResolved, setSearchStateResolved] = useState(false);
+
+  useEffect(() => {
+    const queryParams = qs.parse(search.slice(1));
```
richardxia, line 55: Could you remind me how the flow of data works on the Search Results page, specifically how the initial page URL, the internal React State, and the Algolia InstantSearch widget works?

I forget who was in charge of updating the URL's query parameters, but I just wanted to make sure that we are properly updating the SearchState here when the user interacts with the Algolia widgets like changing filters or changing the search query. I'm a little worried that this `qs.parse()` code will run every time this effect runs, which I think is different from what we had before, where I think it was only running once on initial page load.

app/pages/SearchResultsPage/SearchResultsPage.tsx
schroerbrian: Ooof, good catch

schroerbrian:
Converting to draft while i fix this up

app/pages/SearchResultsPage/SearchResultsPage.tsx
schroerbrian: Update to this comment, specifically ```Toggling a filter triggers a search via the InstantSearch widget, which then triggers the onSearchStateChange callback, which we use to sync the URL and our component's state with Algolia's InstantSearch widget.```

Actually, my order is a bit wrong. The correct order, as I understand it (now), is: Toggling a filter does the following: 

1. Triggers the onSearchStateChange event, which passes a `nextSearchState` param to its callback function
2. We check the nextSearchState value passed by the event and update our URL accordingly 
3. We push the URL onto the history stack which causes our page to re-render
4. During the render, we update the searchState React state value using the new URL params
5. This in turn triggers an Algolia search, I'm assuming because it contains a reference to the searchState state prop

schroerbrian: I should hopefully have this finished up by tomorrow

app/components/search/SearchResults/SearchResults.jsx
```diff
@@ -221,32 +221,32 @@ const SearchResult = ({ hit, index, setCenterCoords }) => {
           {
             key: -2,
             description: "Spanish",
-            url: `${handoutApiUrl}?lang=es`,
```
schroerbrian, line 224: changed the handout query param to handoutLanguage to disambiguate from the lang query param that is used by front-end for Google translate and WeGlot

app/pages/SearchResultsPage/SearchResultsPage.tsx
```diff
@@ -34,22 +36,80 @@ interface SearchState extends ParsedQs {
...
 export const SearchResultsPage = () => {
+  const [cookies] = useCookies(["googtrans"]);
+  const history = useHistory();
+  const { search } = useLocation();
   const { userLocation } = useAppContext();
   const [lastPush, setLastPush] = useState(Date.now());
-  const { search } = useLocation();
   const [expandList, setExpandList] = useState(false);
 
-  const searchState: SearchState = useMemo(
```
schroerbrian, line 42: I'm not exactly sure why useMemo was used here to begin with. Anyhow, `qs.parse...` was run each time there was any kind of editing of the filters in this previous setup, so AFAIK, there was no reason to memoize this value. For the new implementation, search updates, both to the actual search query and the filters, are handled by the searchState state variable and a useEffect block

schroerbrian:
Should be good to go! @richardxia let me know if you have any questions about the relationship between the URL, Algolia's Instant Search widget and the internal React state. I had to dig into this quite a bit so understand the relationship a lot better now

app/pages/SearchResultsPage/SearchResultsPage.tsx
richardxia: Oh geez, thanks for that update. I don't think I actually realized that our page worked that way, and I probably thought it was something closer to your original comment. Thanks again for digging into this deeper!

app/pages/SearchResultsPage/SearchResultsPage.tsx
```diff
@@ -34,22 +36,80 @@ interface SearchState extends ParsedQs {
...
   );
 
+  // In cases where we translate a query into English, we use this ref
+  // to represent the user's original, untranslated input. The untranslatedQuery
+  // is displayed in the UI and stored in the URL params.
+  // This untranslatedQuery value is also checked against when a new search is triggered
+  // to determine if the user has input a different query (vs. merely selecting refinements),
+  // in which case we need to call the translation API again
+  const untranslatedQuery = useRef<string | null | undefined>(null);
+  const translatedQuery = useRef(null);
```
richardxia, line 58: Just wondering, what's the rationale behind using refs instead of state for these variables? I'll have to admit that I haven't used refs much in React, and I think the only times I have ever used it were to get references to DOM elements.

app/pages/SearchResultsPage/SearchResultsPage.tsx
```diff
@@ -34,22 +36,80 @@ interface SearchState extends ParsedQs {
...
+
+  useEffect(() => {
```
richardxia, line 60: I wonder if this logic could be simplified if we broke it up into independent `useEffect()` blocks, where we make the control flow similar in both the English and non-English cases. I think there's four phases/data dependencies:

1. URL query parameters -> untranslated query
3. untranslated query + translation language -> translated query (no-op for the `en` case)
4. translated query + other search params -> search state

I think a rough sketch of the separate `useEffect()` calls I was thinking of would look like this:

```tsx
useEffect(() => {
  setUntranslatedQuery(qs.parse(search.slice(1)).query)
  // Maybe we should simultaneously set a second state variable holding
  // just the non-query parts of `search`?
}, [search])

useEffect(() => {
  // Grab query language from cookie
  if (queryLanguage === "en") {
    setTranslatedQuery(untranslatedQuery)
  } else {
    // Do the API call to the translate_text/ endpoint
  }
}, [untranslatedQuery, cookie.gooogtrans, queryLanguage])

useEffect(() => {
  setSearchState({...urlParams, query: translatedQuery}); 
}, [translatedQuery, search /* maybe instead of search, we should depend on the non-query parts of search? */])
```

I'm not sure if my suggestion here is any less code, but I think it makes it a little bit clearer which variables are dependent on other variables, and it also leverage's React's `useEffect()`'s second argument for keeping track of when things need to be updated. I think that would make it less error-prone compared to us basically reimplementing that ourselves in a single `useEffect()` block. I think this would bring this code closer to the [Functional Reactive Programming (FRP) paradigm](https://en.wikipedia.org/wiki/Functional_reactive_programming) that React is loosely based on, which is all about declaring data dependencies between variables and letting the system figure out which ones need to be updated due to dependent variables being updated.

I think the main downside with my suggestion is that this would cause up to four re-renders when updating the search params. However, I think this is mitigated by the

```tsx
  if (searchState === null) {
    return null;
  }
```

since at least for the first page load, the first few rerenders would be cheap. If performance is an issue for unnecessarily rerendering the child component (`InnerSearchResults`), then I think we could make use of [`React.memo()`](https://react.dev/reference/react/memo#reference) (not the same thing as `useMemo()`).

Anyway, this is just a thought I had, and I don't think it's strictly necessary to do it this way. I just thought that this could be a way of structuring this `useEffect()` block differently such that makes the dependencies a little clearer.

richardxia (approved):
LGTM! I left a pretty large comment suggesting an alternative way of structuring your main `useEffect()` block, but it's just a suggestion, and I think what you have here already is good to merge, too. Thanks for working on this and digging into how the state is managed on this page!

app/pages/SearchResultsPage/SearchResultsPage.tsx
schroerbrian: Updating refs does not trigger a state update. Since our code updates the `untranslatedQuery` and `translatedQuery` value within the useEffect hook, I was running into this React error: ```react-dom.development.js:23093 Uncaught Error: Maximum update depth exceeded. This can happen when a component repeatedly calls setState inside componentWillUpdate or componentDidUpdate. React limits the number of nested updates to prevent infinite loops.```

app/pages/SearchResultsPage/SearchResultsPage.tsx
schroerbrian: Great suggestion. I agree with you that breaking it up into multiple useEffect blocks clarifies our code. I've gone ahead and followed your suggestion and it seems to work well. Qualitatively speaking, performance looks good. Thanks for taking the time to lay all of this out.

app/pages/SearchResultsPage/SearchResultsPage.tsx
schroerbrian: (though this is now sort of moot as I've switched these to useState per your suggested changes to the useEffect block)

PR #1262, opened 2023-05-15
Edit Page: Add type annotations to EditNotes.

richardxia:
Refs #1175.

This is a mostly straightforward change to add strict type annotations to anything related to editing notes. There were some surprising things discovered from adding strict types, mostly involving the direct mutation of objects that should not be mutated in React. We leave this behavior in for now, but with comments describing what should eventually be fixed.


schroerbrian (approved):
LGTM! Thanks for adding the explanatory comments.

richardxia:
Thanks for reviewing!

PR #1263, opened 2023-05-17
Edit Page: Add type annotations to EditSchedule.

richardxia:
Refs #1175.

This adds type annotations to EditSchedule and related components on the Edit Page. It doesn't add any new types, but it loosens a few types in existing places to account for nullness, which the code already safely handled before.

### DIFF
--- a/app/components/edit/ProvidedService.tsx
+++ b/app/components/edit/ProvidedService.tsx
@@ -14,7 +14,7 @@ import type {
 
 import s from "./ProvidedService.module.scss";
 
-interface InternalScheduleDay {
+export interface InternalScheduleDay {
   opens_at: number | null;
   closes_at: number | null;
   /** The DB ID of the ScheduleDay. */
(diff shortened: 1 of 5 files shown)


app/components/edit/ProvidedService.tsx
```diff
@@ -50,6 +50,15 @@ export interface InternalSchedule {
   Sunday: InternalScheduleDay[];
 }
 
+/** If the argument is either an InternalSchedule or an empty object, return
+ *  true if it is an InternalSchedule.
+ */
+export function isNonEmptyInternalSchedule(
```
schroerbrian, line 56: Very small nit but it might be more clear to name this function `isEmptyInternalSchedule` and to check if "Monday" is not in the schedule?

schroerbrian (approved):
LGTM! Great work

app/components/edit/ProvidedService.tsx
richardxia: Yep, good idea, flipped it in 817f0d76dab5f1482f55c20e4f00fe2fa99be72e.

PR #1270, opened 2023-06-01
Edit Page: Add type annotations to EditAddress

richardxia:
Refs #1175.

I feel like I'm finally getting close to the end here. This should be the last subcomponent of the Edit Page that needs type annotations, and I think the bulk of the remaining missing type annotations are all on the top-level OrganizationEditPage itself.

This PR is a bit bigger than the last several that I've submitted. There were a couple refactoring changes I made before actually adding type annotations to the code:

- 129159041b8c73eb9afbeeacb389c89b302b015a removes the `country` field from the `Address` type. We had actually removed this field from the API side while working on the Multiple Addresses project in https://github.com/ShelterTechSF/askdarcel-api/pull/586, since it was confusing to users on the edit page, and [we left some vestiges of it on the frontend](https://github.com/ShelterTechSF/askdarcel-web/pull/1044) just to decouple the merging of the PRs. This commit finally removes those vestiges from the frontend code
- 4435c9b32637ba765de21a5be466e6b43d5be0de corrects a few of the type annotations on the `Address` type by making them nullable. This makes it match the API backend's [DB schema](https://github.com/ShelterTechSF/askdarcel-api/blob/ed2b739b9c1cca65e3dd5972a527306a7d793a1f/db/schema.rb#L25-L45)
- 874ae97669165ca34a3fc206cfc03f561fad8614 is finally the bulk of the changes, which are to add type annotations to the EditAddress component and all components related to it. I think probably the most notable thing here is that I created a new `InternalAddress` type, matching the convention for other types I've defined that are derived from the common ones but specialized to the Edit Page. Here, the main differences are that these reflect only the fields that are visible on the Edit Page, as we don't even display a number of the fields available in the DB table, such as address lines 3 and 4 and the attention field.

  There are other little gotchas related to the type system that I had to deal with, but I believe any of the really tricky ones should have comments above them.

---

One more thing that I feel I should provide a bit of context for is the notion of an "address handle", since some of the code in the diff is related to that. This is something I added during the Multiple Addresses work, and it's supposed to be a field on Services which hold the integer index of an Address defined on the Resource. We need this on the Edit Page in order to simplify the logic on the Edit Page.

Normally, the API returns copies of an Address on the Resource and on the Services, meaning that address information is duplicated between Resources and Services and even across different Services at the same address. On the read/listing pages, this is fine and even a bit convenient, but on the edit page, this makes it extremely problematic to keep track of edits to addresses, since the edits may be made to one address but not the other copies. The way we were originally handling this on the Edit Page was to compute diffs, which was somewhat tractable to do when we could only have a single Address per Resource, and Services could only be at that same address.

To make this work with multiple addresses, right after the Edit Page loads and when the API response with the initial page data comes back from the API, we transform the data such that we remove the copies of the Addresses on all the Services, and we replace them with these address handles. These are like "pointers" in languages that have pointers. By removing all the Addresses on the Services, we ensure that there is only one copy of any particular Address, and it is stored on the Resource. (Actually, it's more complicated than that because we always have two copies of everything on the Edit Page: the original, as it came from the API, and a modified version) This way, when we edit an Address, we only edit a single copy of it, and when we need to determine whether we need to send an update request to the API, we just need to check that one Address.

I hope that wasn't too long-winded of an explanation. I think this is documented in the comments on the Edit Page somewhere, but I wanted to summarize all that in case you were wondering just what an "address handle" is.

### DIFF
--- a/app/components/edit/EditAddress.tsx
+++ b/app/components/edit/EditAddress.tsx
@@ -1,15 +1,34 @@
 import React, { useState } from "react";
 import ReactModal from "react-modal";
-import PropTypes from "prop-types";
 import {
   RiDeleteBin5Line,
   RiEditBoxLine,
   RiArrowLeftLine,
 } from "react-icons/ri";
+import type { InternalAddress } from "../../pages/OrganizationEditPage";
+import type { Address } from "../../models";
 
 import s from "./EditAddress.module.scss";
 
 // Subcomponents
+
+// Note: This must be manually updated to match the actual <input> fields in the
+// EditAddressModal component below.
+interface EditAddressFormData {
+  name: string;
+  address_1: string;
+  address_2: string;
+  city: string;
+  state_province: string;
+  postal_code: string;
+}
+
+type EditAddressModalProps = {
+  isOpen: boolean;
+  onRequestClose: () => void;
+  defaultData: InternalAddress | Record<string, never>;
+  onSave: (address: InternalAddress) => void;
+};
 
 /** Modal containing the edit and add new address form.
  *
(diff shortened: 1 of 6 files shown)


app/components/edit/EditAddress.tsx
```diff
@@ -1,31 +1,62 @@
...
-    const newData = Object.fromEntries(formData.entries());
+    const formData = new FormData(e.currentTarget);
+    // We need to perform a double type assertion with the intermediate `as
+    // unknown` because the types for FormData.entries() don't exactly line up
+    // with EditAddressFormData. FormData.entries() allows values to either be
+    // strings or files, but EditAddressFormData's values can only be strings.
+    // Our form doesn't include a file upload, so we cast it away.
+    const newData: EditAddressFormData & { id?: number } = Object.fromEntries(
+      formData.entries()
+    ) as unknown as EditAddressFormData;
```
schroerbrian, line 59: Wow, didn't know you could do this!

schroerbrian (approved):
Sorry for the late review, I was away this weekend. This looks great! This is some heavy duty typescript wizardry. I'm excited that this is near the finish line.

app/components/edit/EditAddress.tsx
richardxia: Yeah, this got a bit more complicated than I really wanted it to be, but I didn't want to have to have to define and name a new type, so I went with creating an anonymous type intersection.

As cool as it is to be able to do this kind of stuff, I generally think that if we need complicated types like this, then the design of our program isn't exactly right. Hoping to tackle actual refactoring soon, after this TypeScript migration is done!

richardxia:
> Sorry for the late review, I was away this weekend. This looks great! This is some heavy duty typescript wizardry. I'm excited that this is near the finish line.

No problem, and thanks as always for reviewing these PRs!

PR #1275, opened 2023-06-13
Remove EditPatientHandout.

richardxia:
This is effectively unused, since the API field that it edits is no longer used on the non-edit pages. This also removes EditServiceChildCollection, since this was the only component it was used on.


schroerbrian (approved):
LGTM!

PR #1276, opened 2023-06-13
Migrate Edit Page to TypeScript Part XII

richardxia:
Refs #1175.

I think this will be the penultimate PR for the Edit Page TypeScript migration. This adds type annotations everything except for the code path for editing existing Resources. Specifically, this fixes up some `any` types on the EditSidebar component, it adds type annotations to some generic utility functions (e.g. `deepClone()`), and it adds type annotations to the code that's called when creating a brand new Resource.

After this is merged, I think I should be able to do the last set of changes all in one shot.

### DIFF
--- a/app/pages/OrganizationEditPage.tsx
+++ b/app/pages/OrganizationEditPage.tsx
@@ -118,7 +118,9 @@ const applyChanges = (
   return newTransformedItems;
 };
 
-function getDiffObject(curr, orig) {
+/** Return an object that contains only the key-value pairs in `curr` that are different than `orig`.
+ */
+function getDiffObject<T extends {}>(curr: T, orig: T): Partial<T> {
   return Object.entries(curr).reduce((acc, [key, value]) => {
     if (!_.isEqual(orig[key], value)) {
       acc[key] = value;
@@ -413,7 +415,8 @@ const prepSchedule = (scheduleObj) => {
   return newSchedule;
 };
 
-const deepClone = (obj) => JSON.parse(JSON.stringify(obj));
+const deepClone = <T extends unknown>(obj: T): T =>
+  JSON.parse(JSON.stringify(obj));
 
 type AddressChangeType =
   | { type: "markedForRemoval"; handle: number }
@@ -486,7 +489,7 @@ const computeTypeOfChangeToAddresses = (
  * This recipe was copied from
  * https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set
  */
-const setDifference = (set1, set2) =>
+const setDifference = <T extends unknown>(set1: Set<T>, set2: Set<T>): Set<T> =>
   new Set([...set1].filter((x) => !set2.has(x)));
 
 // Helper functions for computing views on state.
@@ -973,7 +976,7 @@ class OrganizationEditPage extends React.Component<Props, State> {
         postDocuments(currentService.documents, promises);
         delete currentService.documents;
         delete currentService.shouldInheritScheduleFromParent;
-        const { addressHandles } = currentService;
+        const { addressHandles }: { addressHandles: number[] } = currentService;
         delete currentService.addressHandles;
         if (!_.isEmpty(currentService)) {
           promises.push(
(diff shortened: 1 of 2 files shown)


app/pages/OrganizationEditPage.tsx
```diff
@@ -118,7 +126,9 @@ const applyChanges = (
   return newTransformedItems;
 };
 
-function getDiffObject(curr, orig) {
+/** Return an object that contains only the key-value pairs in `curr` that are different than `orig`.
+ */
+function getDiffObject<T extends {}>(curr: T, orig: T): Partial<T> {
```
schroerbrian, line 131: I have some idea of how this "T" pattern works and I think I have even used it but my memory is a little fuzzy about details. Do you have a resource you can link offhand that explains it? I am Googling a bit but perhaps there is something out there that you can vouch for.

app/pages/OrganizationEditPage.tsx
```diff
@@ -392,17 +402,54 @@ function createFullSchedule(scheduleObj) {
...
+/** A new ScheduleDay object that is about to be POSTed to the API. */
+type NewScheduleDay = Omit<ScheduleDay, "id">;
+
+const prepSchedule = (
+  scheduleObj: Record<string, never> | InternalSchedule
+): NewScheduleDay[] => {
+  const newSchedule: NewScheduleDay[] = [];
+  Object.keys(scheduleObj).forEach((untypedDay) => {
+    // Object.keys() always returns strings, even when we know(?) that the keys
+    // are a more precise type, since object types in TypeScript only desrcibe
```
schroerbrian, line 441: small nit - "describe" is misspelled

schroerbrian (approved):
LGTM! I'm excited that we're almost at the finish line!

app/pages/OrganizationEditPage.tsx
richardxia: Yeah, the most common name for this feature across most languages is [generics](https://en.wikipedia.org/wiki/Generic_programming), and it's the official name in TypeScript as well. (The more academic name for it is [parametric polymorphism](https://en.wikipedia.org/wiki/Parametric_polymorphism), in case that term ever pops up.) Here are [TypeScript's docs on Generics](https://www.typescriptlang.org/docs/handbook/2/generics.html). One thing to keep in mind about TypeScript's implementation that is different than in most languages is that the TypeScript implementation does not affect the actual behavior of the code, since a "TypeScript compiler" basically just strips away all the TypeScript syntax. In most statically typed languages with generics, the concrete value of the generic type can influence which method you actually call.

`T` in this example would be called a _type parameter_, and it is similar to normal function parameters except that it only holds types and it generally can only be used in places where types are allowed, like after a `:` in a variable/parameter declaration, in a `as ...` type assertion, or when explicitly passing a type parameter to a function. Normally you don't have to explicitly pass in type parameters because they get inferred from the types of the arguments.

The `extends ...` is a type constraint, which says that `T` must be compatible with the type on the right-hand side of `extends`. Without writing the `extends`, you would normally read it as "for any `T`", and with the `extends` you would read it as "for any `T` that extends ...".

The difference between `getDiffObject<T extends {}>(curr: T, orig: T)` and `getDiffObject(curr: {}, orig: {})` is that in the latter, `curr` and `orig` can be any object type (note that things like `number` and `string` are not considered objects in JavaScript), but we aren't guaranteeing that they are necessarily the same object type. With the type parameter, we are saying that `curr` and `orig` must be the same type `T`, and `T` can be any type that extends `{}` (a.k.a. any object type).

One last thing to mention is that there are unfortunately conflicts in TypeScript's syntax and React's JSX syntax, since the `<>` can sometimes be mistaken for HTML tags. This specifically affects arrow functions, since the `<>` would go immediately after the `=` and before the opening parentheses of the parameter list, e.g. `const foo = <T>(x: T) => ...`. To be able to use type parameters on arrow functions, you have to add a bit of superfluous code to allow the TypeScript compiler disambiguate the two. I've been using `const foo = <T extends unknown>(x: T)` as the workaround, since that's what I originally learned, but it seems like the [modern recommendation is to just put a comma after the type parameter](https://stackoverflow.com/a/45576880), e.g. `const foo = <T,>(x: T)`.

schroerbrian: Awesome, thank you for the thorough explanation! That is also good to know about the syntax clash between TS and React

PR #1277, opened 2023-06-14
Migrate Edit Page to TypeScript: Final

richardxia:
Refs #1175.

This should be the last major PR to migrate the Edit Page over to TypeScript. This goes through the last remaining functions in the OrganizationEditPage and adds type annotations to them. There were a number of weird things that I hit, but I hope most of the inline comments are self-explanatory.

One thing to note is that most of the `postServices()` is basically unsalvageable, since we start with an object of type `InternalService`, and we gradually morph and delete properties off of it until we get an object that looks more or less like `Service`, the type that represents what the API sends to the frontend app upon a GET request. There's no way for TypeScript to handle this because the object is in an intermediate state during the entirety of the transition. As such, I liberally added `// @ts-ignore` comments to the function until it passed TypeScript type checking. This entire function should probably be rewritten from scratch, and instead of gradually mutating the internal representation of a Service until it looks like an API-compatible one, we should just build the API-compatible one from a brand new object.

There are actually a handful of smaller TypeScript-related things that I still need to do in a followup PR, since I had disabled the `noImplicitAny` configuration option in https://github.com/ShelterTechSF/askdarcel-web/pull/1230. Re-enabling it will require changes to stuff outside of the Edit Page because a few of the other changes we've been making to the site run afoul of that option, but I plan to do that in a separate PR.

### DIFF
--- a/app/models/Service.ts
+++ b/app/models/Service.ts
@@ -27,6 +27,7 @@ export interface Service {
   categories: Category[];
   certified_at: string | null;
   certified: boolean;
+  documents: unknown[];
   eligibilities: Eligibility[];
   email: string | null;
   featured: boolean | null;
--- a/app/pages/OrganizationEditPage.tsx
+++ b/app/pages/OrganizationEditPage.tsx
@@ -55,6 +55,32 @@ interface UriObj {
   path: string;
 }
 
+/** Safely parse a string as a decimal integer or throw an error.
+ *
+ * This is primarily intended for the use case where we have an object with
+ * numeric keys, since objects always coerce the keys to strings.
+ * Object.entries() will therefore always return string keys. Previously, we
+ * were relying on JavaScript's automatic coercion from string to number when
+ * comparing the string key to a number literal (e.g. `key < 0` would attempt to
+ * coerce `key` to a number, even if it's actually a string). TypeScript
+ * disallows these coercions (rightfully so, since this is terrifying!).
+ *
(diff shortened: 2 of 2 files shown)


app/models/Service.ts
```diff
@@ -27,6 +27,7 @@ export interface Service {
   categories: Category[];
   certified_at: string | null;
   certified: boolean;
+  documents: unknown[];
```
schroerbrian, line 30: Including `documents` is an interesting conundrum since I'm not sure that we'll use documents again, now that we dynamically generate them, but I guess we can just leave it until we make a change to the actual API itself.

schroerbrian (approved):
Looks great, thanks for bringing this past the finish line!

app/models/Service.ts
richardxia: This is actually required or else the `delete currentService.documents` statement in the `postServices()` function will fail to type check. TypeScript isn't actually great for the kind of code we have there, since TypeScript's interface/object types only guarantee the _minimum_ set of properties that will exist, but provide no guarantee that you won't find other properties.

The code we have in `postServices()` is pretty bad, IMO, and we should replace it with code that constructs the service object we'll POST to the API from scratch rather than by deleting stuff off the internal state variable until it matches the shape needed by the API. I'll be happy to finally be able to do that after getting through this TypeScript migration and after the new, non-ChangeRequest API endpoints @jjfreund's been working on are available!

richardxia:
Thanks for reviewing this!

PR #1279, opened 2023-06-15
Update partner links

schroerbrian:


schroerbrian:
Merging this in so that I can deploy today before my vacation

PR #1281, opened 2023-07-19
Fix texting component to allow texting Resource info

schroerbrian:
When looking to convert the SearchResults file to TypeScript, I noticed that our Texting component does not support texting Resource info. This is deceptive as the "text me the info" link is displayed on Resource search listings. Thus, I've created a PR to fix this before getting back to migrating the SearchResults page, which I plan to do prior to making some changes to it as part of some UCSF work. See corresponding API PR [here](https://github.com/ShelterTechSF/askdarcel-api/pull/719)

### DIFF
--- a/app/components/Texting/Texting.tsx
+++ b/app/components/Texting/Texting.tsx
@@ -8,7 +8,10 @@ import { Loader } from "../ui";
 import { SentView } from "./components/SentView";
 import { ErrorView } from "./components/ErrorView";
 
-export type TextListing = {listingName: string} & ({serviceId: number} | {resourceId: number});
+export type TextListing = { listingName: string } & (
+  | { serviceId: number }
+  | { resourceId: number }
+);
 
 /** Payload for the create Texting API endpoint. */
 export interface APITexting {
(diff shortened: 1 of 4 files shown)


app/components/Texting/Texting.tsx
```diff
@@ -8,31 +8,28 @@ import { Loader } from "../ui";
...
- * TODO: Replace this with an app-wide Service type.
- */
-export interface TextingService {
-  serviceName: string;
-  serviceId: number;
+export interface TextListing {
+  listingName: string;
+  serviceId?: number;
+  resourceId?: number;
 }
```
richardxia, line 15: I'm inspired by the DB constraint you added on the backend side, and if you wanted to do something similar here, you could write this as:

```suggestion
export type TextListing = {listingName: string} & ({serviceId: number} | {resourceId: number});
```

Or slightly less fancy:

```suggestion
export type TextListing = {listingName: string; serviceId: number} | {listingName: string; resourceId: number};
```

richardxia (approved):
LGTM! Thanks for noticing that we weren't supporting this for Organizations and fixing it!

app/components/Texting/Texting.tsx
schroerbrian: Great suggestion, thanks!

schroerbrian: Ultimately, I also had to add a "type" prop to the interface as the compiler was complaining when I would try to read a resourceId or serviceId off of a TextListing record, given either prop potentially does not exist. Adding the type and using it as a check satisfied the compiler. Anyhow, great improvement and will prevent possible bugs, thanks!

PR #1282, opened 2023-07-21
Search results typescript migration

schroerbrian:
Migrates the SearchResults page to TypeScript. I completed this as I will be adding some changes for a new UCSF tile category to this file and wanted to to clean it up beforehand.

This also expands our SearchHit interface definition to include props it didn't previously contain. 

This branch was branched off of the `support-texting-resource-info` branch, as the latter is in an [active PR](https://github.com/ShelterTechSF/askdarcel-web/pull/1281), which also contains changes to the SearchResults page. That PR should be prioritized over this one.

### DIFF
--- a/app/components/search/SearchResults/SearchResults.tsx
+++ b/app/components/search/SearchResults/SearchResults.tsx
@@ -60,8 +60,8 @@ const SearchResults = ({
           search term.
         </div>
         {hits.map((hit, index) => (
-            <SearchResult hit={hit} index={index} key={hit.id} />
-          ))}
+          <SearchResult hit={hit} index={index} key={hit.id} />
+        ))}
         <ResultsPagination noResults={!hits || !hits.length} />
       </div>
       <SearchMap
--- a/app/components/search/SearchResults/SearchResults.tsx
+++ b/app/components/search/SearchResults/SearchResults.tsx
@@ -1,8 +1,10 @@
 import React, { useState, useEffect } from "react";
 import { Link } from "react-router-dom";
 import ReactMarkdown from "react-markdown";
-import { connectStateResults } from "react-instantsearch/connectors";
-import { Props as GoogleMapProps, Coords } from "google-map-react";
+import {
+  connectStateResults,
+  SearchResults as SearchResultsProps,
+} from "react-instantsearch/connectors";
 import { whiteLabel } from "utils";
 import { SearchMap } from "components/search/SearchMap/SearchMap";
 import ResultsPagination from "components/search/Pagination/ResultsPagination";
@@ -15,12 +17,17 @@ import { SearchHit, transformHits } from "../../../models/SearchHits";
 import { icon } from "../../../assets";
 import styles from "./SearchResults.module.scss";
 
-const SearchResults = ({ searchResults, expandList }) => {
+const SearchResults = ({
+  searchResults,
+  expandList,
+}: {
+  searchResults: SearchResultsProps;
+  expandList: boolean;
+}) => {
   const [centerCoords] = useState(null);
-  interface MapObjectProps extends GoogleMapProps {
-    setCenter: (coords: Coords) => void;
-  }
-  const [googleMapObject, setMapObject] = useState<MapObjectProps | null>(null);
+  const [googleMapObject, setMapObject] = useState<google.maps.Map | null>(
+    null
+  );
 
   useEffect(() => {
     if (centerCoords && googleMapObject) {
@@ -35,7 +42,7 @@ const SearchResults = ({ searchResults, expandList }) => {
 
   if (!searchResults) return null;
 
-  const hits = transformHits(searchResults.hits as SearchHit[]);
+  const hits = transformHits(searchResults.hits as unknown as SearchHit[]);
 
   return (
     <div className={styles.searchResultsAndMapContainer}>
@@ -52,8 +59,7 @@ const SearchResults = ({ searchResults, expandList }) => {
           No results found in your area. Try a different location, category, or
           search term.
         </div>
-        {hits.length &&
-          hits.map((hit, index) => (
+        {hits.map((hit, index) => (
             <SearchResult hit={hit} index={index} key={hit.id} />
           ))}
         <ResultsPagination noResults={!hits || !hits.length} />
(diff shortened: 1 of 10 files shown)


app/components/search/SearchResults/SearchResults.tsx
```diff
@@ -2,52 +2,38 @@ import React, { useState, useEffect } from "react";
...
-  });
-
-// Todo: setExpandList will be used as part of next stage of multiple location work
-// eslint-disable-next-line no-unused-vars
-const SearchResults = ({ searchResults, expandList, setExpandList }) => {
-  const [centerCoords, setCenterCoords] = useState(null);
-  const [googleMapObject, setMapObject] = useState(null);
+const SearchResults = ({ searchResults, expandList }) => {
+  const [centerCoords] = useState(null);
+  interface MapObjectProps extends GoogleMapProps {
```
richardxia, line 18: Is this type correct? The [google-map-react docs](https://github.com/google-map-react/google-map-react#use-google-maps-api) seem to indicate that the `onGoogleApiLoaded` prop gives you a map object, not a map props object. The map objects appears to be the actual object defined by the official Google Maps JavaScript API and not the google-map-react wrapper.

It looks like the [`@types/google.maps`](https://www.npmjs.com/package/@types/google.maps) NPM package contains type definitions that are automatically generated from the official Google Maps JS API, and it defines a [`Map`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/f33d088a06faa38729454ae4278624bd84441b35/types/google.maps/index.d.ts#L3171) class that has a [`setCenter()`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/f33d088a06faa38729454ae4278624bd84441b35/types/google.maps/index.d.ts#L3352) method on it, which I think is the actual type of the object you have here.

If this sounds right, then could you add `@types/google.maps` as a dependency and use that `Map` type instead of this `MapObjectProps` interface that you defined?

app/components/search/SearchResults/SearchResults.tsx
```diff
@@ -2,52 +2,38 @@ import React, { useState, useEffect } from "react";
...
-        : null;
-    return { ...hit, recurringSchedule };
-  });
-
-// Todo: setExpandList will be used as part of next stage of multiple location work
-// eslint-disable-next-line no-unused-vars
-const SearchResults = ({ searchResults, expandList, setExpandList }) => {
-  const [centerCoords, setCenterCoords] = useState(null);
-  const [googleMapObject, setMapObject] = useState(null);
+const SearchResults = ({ searchResults, expandList }) => {
```
richardxia, line 16: Could you declare the type of the props for this component? According to `@types/react-instantsearch-core`, the [`connectStateResults()`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/f33d088a06faa38729454ae4278624bd84441b35/types/react-instantsearch-core/index.d.ts#L597) higher-order component should inject a [`StateResultsProvided`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/f33d088a06faa38729454ae4278624bd84441b35/types/react-instantsearch-core/index.d.ts#L571C29-L571C29) as the props. The `TDoc` type parameter looks like it ultimately gets used to describe the type of the hits in [`SearchResults.hits`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/f33d088a06faa38729454ae4278624bd84441b35/types/react-instantsearch-core/index.d.ts#L710C18-L710C18), so I think it might work to pass in `SearchHit` as that type parameter.

app/components/search/SearchResults/SearchResults.tsx
```diff
@@ -2,52 +2,38 @@ import React, { useState, useEffect } from "react";
...
+
     return () => {
       document.body.classList.remove("searchResultsPage");
     };
   }, [googleMapObject, centerCoords]);
 
   if (!searchResults) return null;
 
-  const hits = transformHits(searchResults.hits);
+  const hits = transformHits(searchResults.hits as SearchHit[]);
```
richardxia, line 36: This might be able to go away if you declare the types of the props of the component.

app/components/search/SearchResults/SearchResults.tsx
```diff
@@ -64,14 +50,10 @@ const SearchResults = ({ searchResults, expandList, setExpandList }) => {
...
         </div>
-        {hits.map((hit, index) => (
-          <SearchResult
-            hit={hit}
-            index={index}
-            setCenterCoords={setCenterCoords}
-            key={hit.id}
-          />
-        ))}
+        {hits.length &&
```
richardxia, line 53: What's the purpose of this check? Won't `hits.map( ... )` have the same behavior regardless of whether there are 0 or more hits? Actually, I think what you have here will differ only in that it'll render a literal `0` onto the page if there are zero hits, since `{foo && bar}` will return `0` if `foo` is `0`, and if you return a 0, React will render that 0, similar to any other number. The `&&` operator in a React conditional rendering expression is only safe to use if the left side of it is a boolean, `null`, or `undefined`. If it's a number, then although `0` is treated as false in boolean expression, it will still return `0`, the number, which will be rendered literally onto the page.

richardxia (approved):
A few questions about the types, but otherwise this LGTM!

app/components/search/SearchResults/SearchResults.tsx
schroerbrian: Ah thanks! That was puzzling me. The `Map` class you suggested is correct

TIL -- interestingly the google.maps package has no exports but its types are added to the TS index

app/components/search/SearchResults/SearchResults.tsx
schroerbrian: Ah this was also throwing me for a loop. Ultimately, I used the actual `SearchResults` class, which I aliased as `SearchResultsProps`

app/components/search/SearchResults/SearchResults.tsx
schroerbrian: The TS compiler was not happy about the discrepancy between 'Hit<BasicDoc>[]' which is the instantsearch default definition of a hit and our `SearchHit[]`. I tried to extend the `Hit<BasicDoc>` definition when declaring the `SearchHit` interface in the `app/models/SearchHits.ts` but that didn't seem to appease its demands. Anyhow, I was left with doing the following `searchResults.hits as unknown as SearchHit[]` which VSCode suggested as a quick fix...

app/components/search/SearchResults/SearchResults.tsx
schroerbrian: Oh that's a weird gotcha about the 0. I think I added this in response to some error from the compiler, which is no longer a problem after updating some of the types

PR #1283, opened 2023-07-24
Re-enable tsconfig `noImplicitAny`

richardxia:
I had temporarily disabled this setting in https://github.com/ShelterTechSF/askdarcel-web/pull/1230 in order to make the TypeScript migration of the Edit Page easier. Now that the Edit Page has been fully migrated to TypeScript, we can re-enable this setting.

There were a handful of places where I needed to add explicit type annotations in order to make this pass again, but they were relatively straightforward to fix. Note that one of the fixes was to the WeGlot component, which was code added after I had disabled `noImplicitAny`.

PR #1284, opened 2023-07-30
Add missing type annotation in SearchResults.tsx.

richardxia:
This is causing CI to fail on the `master` branch because https://github.com/ShelterTechSF/askdarcel-web/pull/1282 added this code and https://github.com/ShelterTechSF/askdarcel-web/pull/1283 increased the strictness of the type checker, but because the two merged in parallel, neither change failed in CI.


schroerbrian (approved):
Ah thanks for catching this!

richardxia:
Thanks!

PR #1285, opened 2023-08-14
feat: add v2 route for separate backend service config

lgarofalo:
Maybe not the long term plan, but for now we can consider routing to `api/v2` to hit the Go backend.  If we want to migrate any routes from the web app, we can add that `/v2` path param.


schroerbrian (approved):
LGTM!

PR #1287, opened 2023-08-15
bug: fixing nginx conf for api url

lgarofalo:


schroerbrian (commented):
This change LGTM, but there is a conflict that needs to be resolved

PR #1289, opened 2023-08-17
bug: fixing nginx conf for api url

lgarofalo:

PR #1292, opened 2023-08-24
Revert "bug: fixing nginx conf for api url"

lgarofalo:
Reverts ShelterTechSF/askdarcel-web#1289


lgarofalo:
I need to check this better, will do it manually first again.

PR #1295, opened 2023-08-31
1294 refactor notes component

schroerbrian:
This initially started as a project to remove direct state mutations in the Notes component but then I eventually decided to refactor the entire Notes component given a few bad practices I observed, not to mention some confusing code. There isn't a time crunch to get this merged in, so I think it's best to leave it open until @richardxia is back and has a chance to review it, given his deep familiarity with the OrganizationEditPage architecture and our codebase in general

### DIFF
--- a/app/components/edit/EditNotes.tsx
+++ b/app/components/edit/EditNotes.tsx
@@ -15,73 +15,66 @@ export interface InternalNoteChanges {
   isRemoved?: boolean;
 }
 
-interface Props {
-  existingNotes: Note[] | undefined;
+type Props = {
+  notes: Note[] | undefined;
   handleNotesChange: (
-    notesDictionary: Record<number, InternalNoteChanges>
+    notesObject: Record<number, InternalNoteChanges>
   ) => void;
 }
 
-export type State = {
-  notes: Record<number, InternalNoteChanges>;
-  existingNotes: InternalNote[];
-  uuid: number;
-};
+export type NotesObject = Record<number, InternalNoteChanges>;
 
-const EditNotes = ({ existingNotes, handleNotesChange }: Props) => {
-  const internalNotes = existingNotes
-    ? existingNotes.map((note) => {
-        const newNote: InternalNote = {
-          ...note,
-          key: note.id,
-        };
+const EditNotes = ({ notes = [], handleNotesChange }: Props) => {
+  const internalNotes = notes.map((note) => {
+    const newNote: InternalNote = {
+      ...note,
+      key: note.id,
+    };
 
-        return newNote;
-      })
-    : [];
+    return newNote;
+  });
 
-  const [notes, setNotes] = useState(internalNotes);
+  const [notesList, setNotesList] = useState(internalNotes);
   const [uuid, setUuid] = useState(-1);
-  const [notesDictionary, setNotesDictionary] = useState<
+  const [notesObject, setNotesObject] = useState<
     Record<number, InternalNoteChanges>
   >({});
 
   useEffect(() => {
-    handleNotesChange(notesDictionary);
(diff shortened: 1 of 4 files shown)


app/components/edit/EditNotes.tsx
```diff
@@ -12,139 +12,94 @@ export interface InternalNoteChanges {
...
-    const { existingNotes, uuid } = this.state;
-    const newUUID = uuid - 1;
-    existingNotes.push({
-      key: newUUID,
+    return newNote;
+  });
+
+  const [notesList, setNotesList] = useState(internalNotes);
+  const [uuid, setUuid] = useState(-1);
+  const [notesObject, setNotesObject] = useState<NotesObject>({});
```
richardxia, line 34: Is `notesList` equivalent to the previous `existingNotes` state variable, and `notesObject` equivalent to the previous `notes` state variable? It'd be nice if we could unify the two state variables, since it seems a bit confusing to me to have both, but if that's too large of a refactor, I'm wondering if we can revert back to the original variable names? `existingNotes` at least implied that it was originally derived from the initial response from the API, but the current naming of `notesList` and `notesObject` makes it more difficult to understand which is which. Also, if it's not too much trouble, could we also rename `notesObject` to `newNotes`, to better differentiate it from `existingNotes`?

app/components/edit/EditNotes.tsx
```diff
@@ -12,139 +12,94 @@ export interface InternalNoteChanges {
...
-      () => {
-        handleNotesChange(this.state);
-      }
-    );
-  }
-
-  renderNotes() {
-    const { existingNotes } = this.state;
-    return existingNotes.map((note, i) => (
+    newNotes.splice(index, 1);
```
richardxia, line 64: I notice that you're doing something different here than what the code used to do. In the previous code, it looked like we didn't change the length of `newNotes`/`notesList` (formerly named `existingNotes`), and instead we'd just mutate the item in place to set the `isRemoved` property to `true`.

I'm not sure what the consequences of changing the size of the array are, but if we wanted to play it safe, we should keep size of the `newNotes` array the same, and instead we should (immutably) change the object at that index. I think that would look like:

```js
newNotes[index] = {...targetNote, isRemoved: true};
```

richardxia (approved):
In general, 💯 on simplifying the state of this component and only having it sync the state of the single notes list to the OrganizationEditPage component!

The only thing I'm not sure about is whether it is safe to `splice()` the notesList array rather than replacing the item-to-be-removed with a version that has `isRemoved` set to `true`, but if you feel like your local testing of it is sufficient, then I'm 👍 on your changes here.

app/components/edit/EditNotes.tsx
schroerbrian: Yep, `notesList` is equivalent to the previous `existingNotes`. And `notesObject` is equivalent to the previous `notes`. I _believe_ that our update to the API only reference `notes` aka `notesObject`, so trying to combine the two state vars might be a heavier lift than I feel comfortable doing as part of this PR.

But in any case, I've gone ahead and changed the names to your suggestion ones - makes good sense to me!

app/components/edit/EditNotes.tsx
schroerbrian: Yep, I changed it to delete the note and thus update the size of array on purpose. Previously, deleting a note created an annoying experience in the UI where the remaining note input labels would not update with the correct note index. This is because the label uses the index from the `existingNotes` array, which persisted the deleted note in its collection. See an screenshot below from our prod site. My update didn't seem to cause any downstream effects, fortunately.

image

richardxia: Ah, makes sense. Good to know that this fixed a latent bug!

PR #1301, opened 2023-09-13
Remove unused User location reducer.

richardxia:
This doesn't appear to actually be used anywhere, and from its name, it looks like an attempt at defining a Redux reducer for setting the user location. We don't currently use Redux, so we have no use for this code.

I was originally looking into porting these files to TypeScript as part of #1175, but since we're not actually using them, it'd be better to delete them.


schroerbrian (approved):
Cool, thanks for cleaning this up!

richardxia:
Thanks for quickly reviewing!

PR #1304, opened 2023-09-25
Remove ESLint waivers for TypeScript code during transition.

richardxia:
We had disabled some ESLint rules for our TypeScript code when we were initially migrating the codebase. These should no longer be necessary, so they have been removed, and the remaining failures have either been fixed or ignored at the line level.

### DIFF
--- a/app/pages/OrganizationEditPage.tsx
+++ b/app/pages/OrganizationEditPage.tsx
@@ -901,7 +901,9 @@ class OrganizationEditPage extends React.Component<Props, State> {
       const url = `/api/resources/${resourceID}`;
       fetch(url)
         .then((r) => r.json())
-        .then((data) => this.handleAPIGetResource(data.resource));
+        .then((data) =>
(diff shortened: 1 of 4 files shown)


app/pages/OrganizationEditPage.tsx
```diff
@@ -1658,7 +1661,7 @@ class OrganizationEditPage extends React.Component<Props, State> {
         });
       } else {
         dataService
-          .APIDelete(path, { change_request: { status: "2" } } as any)
```
richardxia, line 1661: Quick note on this: The second argument was incorrect, since it's supposed to be for extra HTTP headers to add to the API request, not for extra parameters to send. I specifically tested deactivating both a Resource and a Service after making this change, and it still works correctly.

schroerbrian (approved):
Looks great! Thanks for finishing this up.

richardxia:
Thank _you_, for reviewing all the PRs in this series!

PR #1305, opened 2023-10-02
Refinements fix

schroerbrian:
This expands the refinement limit value that we pass to the Algolia search component, if we are filtering out returned refinements – in cases where we are using `transformItems` to filter the returned refinements, we want the limit to be high to ensure that Algolia returns all of the refinements that we actually want to display. It's a little bit complicated, so I tried my best to explain it in comments. Let me know if this can be improved at all.

Tl;dr: Algolia passes back a limited amount of eligibility or category refinements. We have been filtering this returned list in some cases, without realizing that the original, unfiltered list may not even contain all of the refinements that we want to display. That's why sometimes the list on the sidebar can be very short.

### DIFF
--- a/app/components/ucsf/RefinementLists/ucsfEligibilitiesMap.ts
+++ b/app/components/ucsf/RefinementLists/ucsfEligibilitiesMap.ts
@@ -22,13 +22,6 @@ interface UcsfEligibilityMap {
   [key: string]: EligibilityGroup[];
 }
 
-// N.B.: Until we have found a way to make use of Algolia's "Show More" refinements button,
-// none of the below resource groups should have a _total_ eligibility amount that exceeds the
-// refinementListLimit value defined in the UCSF whitelabel config in whitelabel.ts.
-// E.g., The ucsf-mental-health-resources should not have more than a sum total of X eligibilities
-// among all of its eligibility groups. This excludes any "See All" eligibilities, which we do not
-// display in the sidebar.
-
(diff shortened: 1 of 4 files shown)


app/components/ucsf/RefinementLists/ucsfEligibilitiesMap.ts
```diff
@@ -22,13 +22,6 @@ interface UcsfEligibilityMap {
   [key: string]: EligibilityGroup[];
 }
 
-// N.B.: Until we have found a way to make use of Algolia's "Show More" refinements button,
```
schroerbrian, line 25: Turns out this comment was moot as we are using this the eligibilities in this file to filter out those returned by Algolia anyhow

app/utils/whitelabel.ts
```diff
@@ -45,7 +45,6 @@ interface WhiteLabelSite {
   intercom: boolean;
   logoLinkDestination: string;
   navLogoStyle: string;
-  refinementListLimit: number;
```
schroerbrian, line 48: This config value is not currently need and arguably never was. Algolia's default is 10 and in cases where we are filtering by either a static list of refinements or one returned by the API (in the case of the subcategories endpoint), we need a high value to ensure that all desired categories are returned to begin with

richardxia (approved):
LGTM, thanks for digging into this and understanding the issue!

PR #1307, opened 2023-10-20
Add ucsf partner logo

schroerbrian:
Made updates to the home page partners. These include markup/CSS changes for the list of logos. Adding that extra logo causes them to wrap to the next line, which made them look awkward with our current setup. 

I also added the logo to the SFSG about page as well as the LinkSF page. I'm not exactly sure if we need to keep the LinkSF page updated or even include the UCSF logo there, but it contains our other logos, so I added it for the time being. I'm not sure we get much traffic on LinkSF as it is, anyhow.

I think the UI that we currently have is not perfect, and it can look awkward on smaller screens, but maybe we should wait until our exec discussion before perfecting the logo UI.
Screen Shot 2023-10-20 at 3 59 04 PM

Screen Shot 2023-10-20 at 3 52 58 PM


app/pages/AboutPage/AboutPage.module.scss
```diff
@@ -62,7 +62,15 @@
...
+    display: flex;
+    align-items: center;
     margin: 20px;
+    max-height: 84px;
+  }
+
+  img {
+    max-height: 84px;
+    max-width: 100%;
   }
```
richardxia, line 74: Our original CSS for this was a bit weird, but I think the following would be a cleaner way to enforce the maximum image height and the vertical centering of items without having to introduce a second layer of flexbox:

```suggestion
  align-items: center;

  li {
    flex-shrink: 0;
    margin: 20px;
  }

  img {
    max-height: 84px;
  }
```

To explain this from top to bottom:

- The new `align-items: center` on the `.partnersList` (the `<ul>`) overrides the default value of `stretch`, which would otherwise cause the flex children's heights to always stretch to the height of the largest flex child in that row. Setting it to `center` causes the flex children to have a height that matches their own content's height and to center all the items in a row.
- I reverted the changes you made to the `li` styles, since with the `align-items: center` on the `ul`, these aren't necessary anymore, and so we can get rid of the second layer of flexbox.
- For the image, we still need to set the `max-height`. I _think_ the `max-width` is unnecessary, since it won't really be able to do anything in a horizontally flowing flexbox.

I had to play around with this locally to get this all working, so what I have here may not be optimal, but I think it's got fewer moving pieces to achieve the layout we want. What makes this all tricky is that have to we keep in mind what is determining each element's height and width, and due to flexbox, sometimes it's the parent that constrains the child, but other times it's the child that influences the parent, and it's different for the horizontal and vertical directions. Making things more complicated is that images also have their own, unique resizing rules.

Anyway, it's up to you if you want to take these changes or just go with what you have. We may want to do some larger cleanup of the CSS on this page later on, anyway.

app/pages/HomePage/components/Partners/Partners.tsx
```diff
@@ -7,59 +7,64 @@ import JDCLogo from "./assets/JDCLogo.png";
...
-        <a
-          href="https://sfmohcd.org/"
-          rel="noopener"
-          className="image-container"
-        >
-          <img src={MOHCDLogo} alt="MOHCD" />
-        </a>
-      </li>
-      <li>
-        <header />
```
richardxia, line 37: Looking at the CSS and the rendered page carefully, I think I understand what these empty headers were doing before and why you removed them, but I think it's causing things to be misaligned. Interestingly enough, I have my browser set to a narrower width than you (looks like my width is 1363px), and the Bridge logo was already on its own row for me, prior to the changes in this PR.

With the new logo and other CSS changes, I see this subtle, though noticeable pattern where the logos on the first row seem to creep on upwards, almost in a straight, diagonal line:

Screen Shot 2023-10-22 at 4 43 10 PM

I feel that it's a little bit more distracting than the version you see, where the new logo is on the first row, which helps bring the "center of gravity" back down.

Ideally, we really should be putting the header stuff into their own row so that they don't affect the flex layout of the logos, but as a shorter workaround, what do you think about adding a top-padding to all items except for the first one? I think you can do that by adding the following CSS:

```css
.partners li:not(:first-child) {
  padding-top: 30px;
}
```

richardxia (approved):
LGTM, I left a few detailed CSS-related comments, but I don't think they are blockers for this to merge.

app/pages/AboutPage/AboutPageMarkup/LinkSf.tsx
```diff
@@ -182,12 +182,15 @@ export const LinkSf = () => (
             </a>
           </li>
           <li>
-            <img src={BridgeLogo} alt="Bridge" />
+            <a href="https://acutecare.ucsf.edu/" rel="noopener">
+              <img
+                src={AcuteCareLogo}
+                alt="Acute Care Lab Discharge Navigator"
```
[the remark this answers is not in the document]
schroerbrian: Oh, you know, I don't really know. It is probably more semantic to just use the text that is on the logo, so I will switch this to what you are using.

schroerbrian:
> LGTM, I left a few detailed CSS-related comments, but I don't think they are blockers for this to merge.

Thanks for reviewing! I've added your suggested changes, and everything looks better. Good catch noticing the effects of removing the `header` element from the logo `li` elements. See updated screenshots below:

Screen Shot 2023-10-23 at 12 11 34 PM

Screen Shot 2023-10-23 at 12 13 41 PM

PR #1308, opened 2023-10-23
Justify logos on Home Page

schroerbrian:
Centers logo list on home page. I hadn't checked a larger screen. Fortunately on the standard 13-14 inch laptop screen it looks decent enough; however on a larger desktop monitor the logos are left aligned, which looks a little awkward

image


richardxia (approved):
It looks like the diff is a bit off, probably because you branched off your other branch before it merged into `master`, but this LGTM!

PR #1310, opened 2023-10-31
1274 intimate partner violence

schroerbrian:
As discussed with @richardxia , I'm putting this PR up now to allow him to get a head start on reviewing. I don't expect any significant changes from here on out. I will merge and deploy this once UCSF signs off on our current content strategy. The code however will likely stay the same. The Intimate Partner PDF Page is generally similar to our ServicePDFPage; however there are some slight differences due to the different design of the PDF. I've merged this branch into QA-One and deployed the latest there, so you can check the code at https://qaone.sfserviceguide.org/


app/pages/ServiceDiscoveryResults/ServiceDiscoveryResults.tsx
```diff
@@ -167,9 +165,11 @@ const InnerServiceDiscoveryResults = ({
...
         <Configure
           filters={`categories:'${algoliaCategoryName}'`}
-          aroundLatLng={userLatLng}
-          aroundRadius={searchRadius}
-          aroundPrecision={1600}
+          {...(category.disableGeoLocation !== true && {
+            aroundLatLng: userLatLng,
+            aroundRadius: searchRadius,
+            aroundPrecision: 1600,
+          })}
```
richardxia, line 172: I'm a little surprised, to be honest, that this is legal and doesn't cause a type error. `&&` is short-circuiting, meaning that if the left-hand side of it is `false`, the whole expression returns `false`, and I feel like `{...false}` should be illegal.

I guess if the linters and type checkers aren't yelling about this, then it's OK, but I feel like it might be more readable to pull the condition outside of the `<Configure>` component and to just duplicate the `filters` prop, e.g.:

```jsx
{category.disableGeolocation
  ? <Configure filters={`categories:'${algoliaCategoryName}'`} />
  : <Configure
      filters={`categories:'${algoliaCategoryName}'`}
      aroundLatLng={userLatLng}
      aroundRadius={searchRadius}
      aroundPrecision={1600}
    />
}
```

richardxia (approved):
This LGTM, and sorry for taking so long to review it!

For full disclosure, I didn't really read the React code for the actual PDF page, but I'm going to trust that you did the right thing, especially since we already had the other one that I assume you based this one off of.

app/pages/ServiceDiscoveryResults/ServiceDiscoveryResults.tsx
schroerbrian: Ahh, interesting. I've gone ahead and made your suggested change considering the confusion.

app/components/search/SearchResults/SearchResults.tsx
```diff
@@ -74,10 +81,29 @@ const SearchResults = ({
...
+  index: number;
+  categoryId: string | undefined;
+}) => {
   const [textingIsOpen, setTextingIsOpen] = useState(false);
   const [clinicianActionsIsOpen, setClinicianActionsIsOpen] = useState(false);
   const [handoutModalIsOpen, setHandoutModalIsOpen] = useState(false);
+  type HandoutLanguage = "es" | "tl" | "zh-TW" | "vi" | "ru" | "ar";
+  const handoutUrl = (hitId: number, language: HandoutLanguage | null) => {
+    const handoutRoute =
+      categoryId === "2000006"
```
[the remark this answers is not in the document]
schroerbrian: I like the idea of creating some kind of constants file or directory. Maybe something a new volunteer could look into.

PR #1312, opened 2023-12-08
1274 Show service type/categories first in UCSF discovery steps

schroerbrian:
I think I said that UCSF wanted the refinements order switched on the sidebar during our product/dev meeting, but indeed it was actually in the discovery steps (brain fog). Fortunately that was a much easier change to make.


richardxia (approved):
LGTM!

PR #1314, opened 2023-12-21
1274 24 hour availability sort

schroerbrian:


app/components/search/SearchResults/SearchResults.tsx
```diff
@@ -44,7 +45,12 @@ const SearchResults = ({
 
   if (!searchResults) return null;
 
-  const hits = transformHits(searchResults.hits as unknown as SearchHit[]);
+  const category = CATEGORIES.find((c) => c.id === categoryId);
+  const sortBy24HourAvailability = category && category.sortBy24HourAvailability ? category.sortBy24HourAvailability  : false;
```
richardxia, line 49: I think this should be equivalent to:

```suggestion
  const sortBy24HourAvailability = category?.sortBy24HourAvailability;
```

The left-hand side will evaluate to `null` or `undefined` if `category` is null or undefined, but if not, it'll continue to access the `sortBy24HourAvailability` property. If TypeScript yells at us for passing a `null` in a place where it really wants a boolean, then I think we can wrap the whole thing with `Bool()` to coerce the `null`s and `undefined`s to `false`.

richardxia (approved):
LGTM!

app/components/search/SearchResults/SearchResults.tsx
schroerbrian: Ah good call! I have changed this to your simpler solution. I indeed had to wrap the entire expression in Bool(), even my previous code was upsetting the TS compiler. Sorry I seeem to have totally forgotten about this PR due to the holiday break

PR #1315, opened 2024-01-29
1274 ipv subcategory sort

schroerbrian:
UCSF asked us to order our sidebar subcategories by a defined order that they sent us. Our CategoryRelationship table has a "child_priority_rank" field that I set for the child categories returned by our `Ucsf-ipv` parent category. I then use this order to sort the categories returned by Algolia.

This PR also removes some of the static eligibilities from the IPV constant eligibility list defined on the f/e. These eligibilities are not used for now. Speaking of which, maybe we should create a CategoryEligibility table similar to the CategoryRelationship table that will allow us to define eligibilities to be associated with a given parent category in the DB.

### DIFF
--- a/app/components/search/Sidebar/Sidebar.tsx
+++ b/app/components/search/Sidebar/Sidebar.tsx
@@ -1,5 +1,7 @@
 import React, { useState, useMemo } from "react";
 
+import type { Category } from "models/Meta";
+
 import {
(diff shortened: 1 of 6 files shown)


richardxia (approved):
Sorry for the late review, but this LGTM!

PR #1319, opened 2024-02-08
Edit Page Enhancements

GeorgeCloud:
This ticket fixes issues with the edit page UI here is the [ticket](https://github.com/ShelterTechSF/askdarcel-api/issues/730)

We need to merge these [API changes](https://github.com/ShelterTechSF/askdarcel-api/pull/731) before this PR.

Screenshot 2024-02-27 at 3 55 55 PM

### DIFF
--- a/app/pages/EditBreakingNewsPage.scss
+++ b/app/pages/EditBreakingNewsPage.scss
@@ -55,7 +55,7 @@
   color: $color-grey5;
 }
 
-.breaking-news-round-edges {
+.breaking-news-status-badge {
   min-width: 2em;
   padding: 0.5em;
   border-radius: 5%;
--- a/app/pages/EditBreakingNewsPage.scss
+++ b/app/pages/EditBreakingNewsPage.scss
@@ -42,10 +42,7 @@
   display: flex;
   justify-content: center;
   margin: 10px 0px;
-}
-
-.button-with-margin {
-  margin-right: 10px;
+  gap: 1em;
 }
 
 .breaking-news-footer {
@@ -58,22 +55,25 @@
   color: $color-grey5;
 }
 
-.round-edges {
-  width: 80px;
-  height: 30px;
-  border-radius: 10%;
-  display: flex;
-  align-items: center;
-  justify-content: center;
+.breaking-news-round-edges {
+  min-width: 2em;
+  padding: 0.5em;
+  border-radius: 5%;
+  display: inline-block;
   font-weight: bold;
 }
 
-.active-status {
+.breaking-news-scheduled-status {
   color: white;
   background-color: $color-status-amber;
 }
 
-.inactive-status {
+.breaking-news-active-status {
   color: white;
-  background-color: red;
+  background-color: $color-status-green;
+}
+
+.breaking-news-expired-status {
+  color: white;
+  background-color: $color-status-red;
 }
--- a/app/pages/EditBreakingNewsPage.scss
+++ b/app/pages/EditBreakingNewsPage.scss
@@ -57,3 +57,23 @@
 .labelSubtext {
   color: $color-grey5;
 }
+
+.round-edges {
+  width: 80px;
+  height: 30px;
+  border-radius: 10%;
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  font-weight: bold;
+}
+
+.active-status {
+  color: white;
+  background-color: $color-status-amber;
+}
+
+.inactive-status {
+  color: white;
+  background-color: red;
+}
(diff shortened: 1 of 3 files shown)


app/pages/EditBreakingNewsPage.tsx
```diff
@@ -46,16 +46,25 @@ export const EditBreakingNewsPage = () => {
...
+  const onSave = (index: number) => {
+    const articleFromState = breakingNewsArticles[index];
+    const articleId = articleFromState.id
     const route = `/api/news_articles/${articleId ?? ""}`;
 
-    if (articleId) {
-      dataService.put(route, article);
+    if (articleId){
+      dataService.put(route, articleFromState);
+      alert(`Article "${articleFromState.headline}" has been updated!`)
```
richardxia, line 64: These alerts are great! Do you think you could also add one for the delete action? For delete, since it's a dangerous operation, we may want to use [`confirm()`](https://developer.mozilla.org/en-US/docs/Web/API/Window/confirm) instead of `alert()` so that it shows "Cancel" and "Ok" buttons, allowing the user to back out of deleting if they change their mind.

richardxia (commented):
This is really awesome! I left some suggestions, but this is shaping up to be great, and my experience from checking this out locally was already a lot better than what we had before.

Also, in the future, if it's not too much trouble, could you include a screenshot of the UI in the PR description? It's not critical, but I definitely find stuff like that helpful as a reviewer.

richardxia:
Also, you're failing lint here, at least the code formatting check. To automatically reformat your code, you can run `docker-compose run --rm web npm run prettier:fix`. We have a standard Prettier setup, so you could also install the Prettier plugin for your editor, and it should allow you to format your code based on the config file in this repo.

app/pages/EditBreakingNewsPage.scss
```diff
@@ -47,3 +54,26 @@
 .labelSubtext {
   color: $color-grey5;
 }
+
+.breaking-news-round-edges {
```
richardxia, line 58: Nit: Could we rename this to something like `.breaking-news-status` or `.breaking-news-status-badge`? "round-edges" makes it a little hard to remember exactly what this CSS class is used in, and it's not really a generic "utility" class because it's doing more than just rounding the edges.

richardxia (approved):
Thanks for making these updates! This mostly looks good to me. I think one last functional request I'd make is to sort the articles by status, at least on initial page load. I'm thinking we should do the order "Scheduled", "Active", "Expired", though I think it could also make sense to do "Active", "Scheduled", "Expired". Either way, the expired articles really should go to the bottom of the list, because otherwise, as we accumulate more and more articles, the expired ones will take up more space and push down the other two status.

Also, it looks like you're still failing the lint checks. I think some of it should be automatically fixable by running Prettier, but the remaining ones you may need to fix up manually.

app/pages/EditBreakingNewsPage.tsx
```diff
@@ -11,18 +11,25 @@ export const EditBreakingNewsPage = () => {
...
+          ? formatDate(article.effective_date)
+          : "",
+        expiration_date: article.expiration_date
+          ? formatDate(article.expiration_date)
+          : "",
+      }));
+
+      sortedArticles.sort((a, b) => {
+        const statusOrder = ["Scheduled", "Active", "Expired"];
+        const statusA = articleStatus(a).props.children;
```
richardxia, line 26: Does this actually work? Even if it does, I don't think we should be trying to access the props on a nested React component like this. Instead, I'd factor this out into two separate pieces: 1) a pure function that just returns a status string, and 2) a React component that renders the badge icon.

app/pages/EditBreakingNewsPage.tsx
```diff
@@ -46,29 +53,60 @@ export const EditBreakingNewsPage = () => {
     ]);
   };
 
-  const onSave = (articleId: string) => {
-    const article = breakingNewsArticles.find(
-      ({ id }: any) => id === articleId
+  const articleStatus = (article: NewsArticle): JSX.Element | null => {
```
richardxia, line 56: Related to my other comment, to turn this into a true React component, change the argument from just a plain article to an an object with an `article` property. Also, the convention for React components is to capitalize the first letter of the name, since React components were originally written only as classes, before they added function-based components.

```suggestion
  const ArticleStatus = ({ article}: { article: NewsArticle}) => {
```

(The return type can usually be omitted from React components as well)

richardxia:
Thanks again for working on this @GeorgeCloud! This will be greatly appreciated by the team!

PR #1327, opened 2024-03-14
Use v2 API route for GET services by ID

lgarofalo:
This would flip all GET services by ID to use Go server.

Will discuss with team before any action on this.


lgarofalo:
I will have to update this to use the Go API server as well in the test build now

schroerbrian (approved):
LGTM! 

Just to confirm, PUTs and DELETEs will still be handled by Rails?

lgarofalo:
@schroerbrian yea, the only routes of concern for that is the services/id/change_request

PR #1337, opened 2024-04-11
1321 move routing logic to its own file so that it is a child compone…

schroerbrian:
…nt of the AppProvider component and will thus have access to the auth state

This is a preparatory PR that just moves all of the routing logic into a new component that is a child of the AppContext and thus has access to the AppContext's state props. There are no other changes. My subsequent Navigator Dashboard PR (coming soon), will contain a couple changes to the routes file. I didn't want to include what are essentially copy and paste changes in that PR for ease of reviewing.


richardxia (approved):
This generally looks good to me, but super nit: I feel like the routes is super important to a React app, more so than being just a "utility component", so maybe we should even just have it directly in the `app/` directory? I think our app used to actually have a file at `app/routes.jsx` with the routes, but we had merged it into `App.jsx` at some point. There's no one standard React app layout, but by analogy in Rails and Django, you have pretty prominent files at `config/routes.rb` and `urls.py` that hold the main app routing logic, neither of which are "utils".

schroerbrian:
> This generally looks good to me, but super nit: I feel like the routes is super important to a React app, more so than being just a "utility component", so maybe we should even just have it directly in the `app/` directory? I think our app used to actually have a file at `app/routes.jsx` with the routes, but we had merged it into `App.jsx` at some point. There's no one standard React app layout, but by analogy in Rails and Django, you have pretty prominent files at `config/routes.rb` and `urls.py` that hold the main app routing logic, neither of which are "utils".

Thanks! That makes total sense. I have seen that pattern before.

schroerbrian:
Note: i'm keeping the empty components/utils directory as i'll be placing another file in that directory in an imminent PR

Issue #1321, opened 2024-02-22
Dashboard Page


richardxia:
This is just to have the basic dashboard page, but this does _not_ include either the Advance Search component or the Bookmarks. UI-wise, this can just be a blank page for now, but it should replace the Home Page when you're logged in.

Figma link (for reference, since we won't implement much on the page for now): https://www.figma.com/file/XUttHGa1csIpbSiurbYmjo/Case-Worker?type=design&node-id=7164-2799&mode=design&t=IVgzgf3eRrcoQVnd-4

More details in the [Feature Requirements Doc](https://docs.google.com/document/d/1XsgVEfRH9zn4XMQlhtqgfhaJAoL8TV8F6yNlLFa2MA4/edit#bookmark=id.or2snq786dqd)

schroerbrian:
This feature should also include logic that prevents users from accessing the page; bote: this logic could/should be used across different pages in the future

PR #1338, opened 2024-04-12
1321 dashboard page

schroerbrian:
This adds a dashboard page and protected route logic that only allows users to visit the page if they are authenticated. Refs #1321 . Note: I added bare bones UI to the dashboard page; please feel free to rip it out completely.

Question: is the route `/navigator-dashboard` okay or do we want to allow for additional navigator actions in the future with a more flexible naming pattern like  `/navigator/dashboard`?? any thoughts?


richardxia (approved):
LGTM!

> Question: is the route `/navigator-dashboard` okay or do we want to allow for additional navigator actions in the future with a more flexible naming pattern like `/navigator/dashboard`?? any thoughts?

I think what you have right now is fine. I think we still don't know how much of the interface we're building out will be navigator-specific, and how much will be generic to any kind of user login, so I'm anticipating that we may change the URL structure in the future anyway.

schroerbrian:
Thanks for the review!!

Issue #1331, opened 2024-03-28
Search by location


richardxia:
This covers the ability to filter search results by location, via dragging the Google Maps widget.

Figma: https://www.figma.com/file/XUttHGa1csIpbSiurbYmjo/Case-Worker?type=design&node-id=7167-6972&mode=design&t=hcGEvgxPwD4YO79c-4

Details in [Feature Requirements Doc](https://docs.google.com/document/d/1XsgVEfRH9zn4XMQlhtqgfhaJAoL8TV8F6yNlLFa2MA4/edit#bookmark=id.bqjzk2i0i8n2)

schroerbrian:
I believe that our Map library may support this out of the box

PR #1339, opened 2024-04-15
1331 Implement search by location

schroerbrian:
Refs #1331 . The map accepts user location as default. Upon moving and clicking the "search this area" button the user location is then overridden by map center coords.

_Note: for some reason the desktop screenshot shows the top button border looking weird. This seems to be a fluke of the screenshot rather than how the button actually looks_

Screen Shot 2024-04-15 at 12 09 38 PM
Screen Shot 2024-04-15 at 12 07 24 PM
image

### DIFF
--- a/app/pages/SearchResultsPage/SearchResultsPage.tsx
+++ b/app/pages/SearchResultsPage/SearchResultsPage.tsx
@@ -146,6 +146,10 @@ const InnerSearchResults = ({
   setSearchRadius: (radius: string) => void;
   untranslatedQuery: string | undefined | null;
 }) => {
+  const [aroundLatLng, setAroundLatLng] = useState(
+    userLocation ? `${userLocation.lat}, ${userLocation.lng}` : null
+  );
+
   if (userLocation === null) {
     return <Loader />;
   }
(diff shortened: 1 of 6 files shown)


app/pages/SearchResultsPage/SearchResultsPage.tsx
```diff
@@ -146,6 +146,10 @@ const InnerSearchResults = ({
   setSearchRadius: (radius: string) => void;
   untranslatedQuery: string | undefined | null;
 }) => {
+  const [aroundLatLng, setAroundLatLng] = useState(
+    userLocation ? `${userLocation.lat}, ${userLocation.lng}` : null
+  );
```
richardxia, line 151: Nit/minor comment: It's a little weird to use a string with a comma separating the two values as the internal state representation, since it kind of imposes a requirement on both the setter and getter sides to turn the more natural representation of having two separate numbers into a string and vice versa. From reading the code, I realize that the Algolia InstantSearch components wants the string representation, but maybe we should store it in our state as two separate numbers?

Maybe this could just be an object with the following TypeScript shape?

```ts
type Location = { lat: number; lng: number };
```

Then, the only place where we would convert it to a string is at the `<Configure aroundLatLng={...} ... />` instantiation site, and everywhere else we just pass in the two numbers?

This isn't super critical, but just a thought on how to represent this a bit more cleanly.

richardxia (approved):
Awesome, this is really neat!

app/pages/SearchResultsPage/SearchResultsPage.tsx
schroerbrian: Yes! I appreciate the feedback. That sounds logical. I've updated it.

Issue #1343, opened 2024-04-25
Add Basic Search UI to Navigator Dashboard Page


schroerbrian:
(This ticket precedes #1324, which involves implementing the advanced search features.)

Add core basic search functionality on navigator dashboard for mobile and desktop.

richardxia:
Completed in https://github.com/ShelterTechSF/askdarcel-web/pull/1344

Issue #1324, opened 2024-03-07
Advanced Search UI


richardxia:
This encompasses the advanced search form present at the top of the Dashboard. This advanced UI allows a user to search across multiple service types (categories) and multiple eligibilities at the same time.

Figma: https://www.figma.com/file/XUttHGa1csIpbSiurbYmjo/Case-Worker?type=design&node-id=7164-2799&mode=design&t=hcGEvgxPwD4YO79c-4

Much more details about how to translate the widgets into search results in the [Feature Requirements Doc](https://docs.google.com/document/d/1XsgVEfRH9zn4XMQlhtqgfhaJAoL8TV8F6yNlLFa2MA4/edit#bookmark=id.pdfuwmlly4da).

PR #1344, opened 2024-04-25
1343 advanced search UI

schroerbrian:
Despite the branch prefix, this refs #1343 . I decided to break #1324 up for ease of review and quicker iteration. This PR merely adds our basic search feature, which is a viable and functional feature in its own right. (Though, admittedly, the filter button does not do anything right now!) That said, my next major task shall be to tackle actual advanced search!

Screen Shot 2024-04-24 at 5 12 35 PM
Screen Shot 2024-04-24 at 5 12 05 PM

### DIFF
--- a/app/assets/img/ic-thick-chevron.svg
+++ b/app/assets/img/ic-thick-chevron.svg
@@ -1,0 +1,3 @@
+<svg width="25" height="15" viewBox="0 0 25 15" fill="none" xmlns="http://www.w3.org/2000/svg">
+<path d="M12.4094 14.8389L0.71808 0.588867L24.1008 0.588869L12.4094 14.8389Z" fill="#276CE5"/>
+</svg>
--- a/app/components/search/SearchMap/SearchMap.tsx
+++ b/app/components/search/SearchMap/SearchMap.tsx
@@ -23,6 +23,7 @@ export const SearchMap = ({
   mapObject,
   setMapObject,
   setAroundLatLng,
+  overlayMapWithSearchResults,
 }: {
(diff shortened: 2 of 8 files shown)


app/pages/NavigatorDashboard/NavigatorDashboard.tsx
```diff
@@ -1,19 +1,64 @@
...
+          submitSearch();
+        }}
+      >
+        <div className={styles.searchInputContainer}>
+          <input
+            type="text"
+            placeholder="Emergency Shelter"
+            className={styles.searchInput}
+            onChange={(evt) => {
+              searchValue = evt.target.value;
```
richardxia, line 43: This seems a little bit dangerous, since you're mutating a non-state, non-ref variable, which means that React doesn't know about it and it's not part of the usual way you manage state outside of React. Not a big deal to leave it here, if you were planning to completely revamp this in the next PR, but in case you weren't planning to change this, I think it'd be a lot safer to either make `searchValue` a regular React state variable, or to remove the `onChange` prop completely to leave this as an uncontrolled form component.

richardxia (approved):
LGTM, sorry for the delay!

app/pages/NavigatorDashboard/NavigatorDashboard.tsx
schroerbrian: Thank you! I will merge this i and fix this in my subsequent https://github.com/ShelterTechSF/askdarcel-web/pull/1345 PR

PR #1346, opened 2024-05-10
1322 add bookmark UI

schroerbrian:
Unfortunately, the sort of pseudo-dropdown element listing Bookmark Folders required a lot of custom UI. I didn't think it'd be possible to code this using a standard Select dropdown. Thus, there are some accessibility compromises that I had to make. In hindsight, I probably could have used React-Select's `createable` [select component](https://react-select.com/creatable), but that is also kind of weird UX, and doesn't really match the designs. Anyhow this took me a little longer than I would have liked

Screen Shot 2024-05-09 at 3 44 27 PM
Screen Shot 2024-05-09 at 3 45 03 PM


app/components/ui/BookmarkModal/BookmarkModal.tsx
```diff
@@ -0,0 +1,216 @@
...
+    if (!selectedFolder || !bookmarkName) return;
+    const isNewFolder = selectedFolder.id === -1;
+    if (isNewFolder) {
+      // Create new folder then create new bookmark
+    } else {
+      // Just create new bookmark: pass service/resource ID and folder ID, etc.
+    }
+  };
+
+  const createNewFolder = (name: string) => {
```
richardxia, line 88: This name feels a bit misleading, since I assume you don't intend to actually perform an API request to create the folder. I can't actually think of a good name for this, since the best I can come up with is `setNewFolderName()`, but that's basically what `setNewFolder()` already does.

I'm wondering if it would actually be clearer to just have these be separate functions instead of grouping them into a single function. There's really two separate operations happening here: 1) setting the name of the new folder, which is really just for synchronizing with the text input field, and 2) actually selecting a folder for the dropdown menu. It's admittedly a bit weird that the design essentially has two input fields nested within each other, one for the dropdown menu, and one for the text field for naming a new folder, but I think it might be clearer to separate the two actions.

I'll make another comment lower in the file, where this is being called, to comment more specifically on the events we're hooking into.

app/components/ui/BookmarkModal/BookmarkModal.tsx
```diff
@@ -0,0 +1,216 @@
...
+                    value={newFolder}
+                    onChange={(evt) => createNewFolder(evt.target.value)}
+                    onFocus={(evt) => {
+                      createNewFolder(evt.target.value);
+                    }}
+                    onKeyDown={(evt) => {
+                      if (evt.key === "Enter") {
+                        setExpandFolders(false);
+                      }
+                    }}
```
richardxia, line 195: I feel that the way these events are hooked up would be a bit confusing to me as a user. Like I mentioned earlier, there are two separate actions to take here: setting the name of the new folder, and selecting the new folder as the location for the bookmark. The way you have it here, both typing into the textbox and moving the focus onto the textbox (either by clicking or tabbing) will simultaneously take both actions, while hitting Enter will merely hide the drop-down menu.

I wonder if instead we did the following:

- onChange only changes the name of the new folder, without selecting it
- onFocus does nothing
- onKeyDown (evt.key === "Enter") actually sets the new folder as the selected folder and hides the folders

I think this would make it more similar to the other, non-new folder options, where the action of "selecting" a folder (via clicking) causes the selected folder to change and hides the menu.

I am also wondering if we should attach a click event handler on the `+` button, to have that be able to respond to events. For the normal, non-new folder options, you have the event handler on the entire `<li>` element.

The more I type, the more I'm second-guessing myself, and at this point, I'm not even sure if you should bother making any of the changes I'm suggesting, since it's probably best for us to clarify with a designer.

app/components/ui/BookmarkModal/BookmarkModal.module.scss
```diff
@@ -0,0 +1,190 @@
+@import "~styles/utils/_helpers.scss";
+
+.bookmarkModal {
+  width: 600px;
+  padding: 33px;
+  @media screen and (max-width: $break-tablet-s) {
+    max-width: 600px;
+    width: 100vw;
```
richardxia, line 8: Should this just be `max-width: 100vw` or `width: 100vw`? I'm not sure if setting both to different values makes sense.

app/components/search/SearchResults/SearchResults.tsx
```diff
@@ -215,6 +218,9 @@ const SearchResult = ({
   }
 
   const toggleTextingModal = () => setTextingIsOpen(!textingIsOpen);
+  // TODO: this bookmarkAdded boolean should be set in accordance with the value of the bookmark model
+  // returned by the API or just read off of the `hit` directly; it is set here for dev purposes
+  const bookmarkAdded = false;
```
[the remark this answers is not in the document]
schroerbrian: Oh right, yes! Thank you. I updated my comments to reflect that this will need to be fetched from the API. That would be quite a lot of calls to the API, 20 per page. Do you think it'd be more performant if the API had an endpoint that returned all of a users' bookmarks per service/resource ID? Each `hit` could check against this list to determine whether it is bookmarked or not

app/components/ui/BookmarkModal/BookmarkModal.tsx
```diff
@@ -0,0 +1,216 @@
...
+    const folderElement = foldersContainerRef.current as unknown as Element;
+    if (newFocus && !folderElement.contains(newFocus)) {
+      setExpandFolders(false);
+    }
+  };
+
+  const createBookmark = () => {
+    // TODO: when bookmarks/folder API is done, fix this up
+    if (!selectedFolder || !bookmarkName) return;
+    const isNewFolder = selectedFolder.id === -1;
```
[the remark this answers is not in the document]
schroerbrian: Good idea, that seems better. Thank you.

app/components/ui/BookmarkModal/BookmarkModal.tsx
```diff
@@ -0,0 +1,216 @@
...
+                <i className={`material-icons ${styles.clearBookmarkText}`}>
+                  cancel
+                </i>
+              </button>
+            </div>
+            <p className={styles.bookmarkUrl}>{`${siteUrl}/${
+              hit.type === "service" ? `services` : `organizations`
+            }/${hit.id}`}</p>
+          </div>
+          <p className={styles.label}>LOCATION</p>
```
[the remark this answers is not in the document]
schroerbrian: Good suggestion. I had actually already included the text-transform rule as a fallback, but i've removed all caps string in favor of "Location"

app/components/ui/BookmarkModal/BookmarkModal.tsx
schroerbrian: That makes more sense, thanks!

app/components/ui/BookmarkModal/BookmarkModal.tsx
```diff
@@ -0,0 +1,216 @@
...
+      addModalClass={styles.bookmarkModal}
+      closeModal={() => setIsOpen(false)}
+    >
+      <h2 className={styles.title}>{`${
+        newBookmark ? "Add Bookmark" : "Edit Bookmark"
+      }`}</h2>
+      <div className={styles.modalContent}>
+        <div>
+          <div className={styles.bookmarkDescription}>
+            <i className={`material-icons ${styles.bookmarkIcon}`}>folder</i>
```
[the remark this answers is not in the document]
schroerbrian: Star icon is a great idea. I had initially tried the bookmark icon but i didn't like the way that it looked . Updated.

app/components/ui/BookmarkModal/BookmarkModal.module.scss
richardxia: Ah, I see, I didn't realize that the breakpoint was larger than 600px. In that case, would it make sense to move both the `max-width: 600px` and `width: 100vw` out of the media query, and into the original class? The original class also currently defines `width: 600px`, but maybe it'd be more uniform to have all these styles apply unconditionally, but where the `width: 100vw` kind of replaces the need for the media query because `vw` is a viewport-width-specific unit.

No need to spend too much more time on this, though.

app/components/ui/BookmarkModal/BookmarkModal.tsx
```diff
@@ -0,0 +1,216 @@
...
+      // Create new folder then create new bookmark
+    } else {
+      // Just create new bookmark: pass service/resource ID and folder ID, etc.
+    }
+  };
+
+  const setNewFolderName = (name: string) => {
+    setNewFolder(name);
+    if (name.length) {
+      setSelectedFolder({ id: -1, name });
```
richardxia, line 91: Now that you've made the `id` field optional, you should just be able to omit the `id: -1` here.

richardxia (approved):
This LGTM now, thanks for making those changes! I guess we'll need the API to be implemented before we can fully test that this works, but I don't think there's any harm in merging this in, since the feature is guarded by being logged in, which isn't officially "released" yet.

app/components/ui/BookmarkModal/BookmarkModal.tsx
```diff
@@ -0,0 +1,216 @@
...
+  // TODO: If bookmark already exists, use bookmark name; otherwise,
+  // default to service/resource name. This should be updated with the correct
+  // bookmark name prop when the API is operational
+  const initialName = hit.name;
+  // TODO: Once the service/resource can have an associated bookmark model, we can check
+  // if the bookmark already exists and update various UI on the modal accordingly
+  const newBookmark = false;
+
+  const [folderOptions, setFolderOptions] = useState<Folder[]>([]);
+  const [selectedFolder, setSelectedFolder] = useState<Folder | null>(null);
```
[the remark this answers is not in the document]
schroerbrian: Thanks! I briefly looked into changing this , but figured that maybe it makes more sense to change once we get the actual API rigged up. I added a TODO comment to suggest changing it to an index for the time being. Makes great sense. Data normalization is a concept that I am trying to better grasp, so I appreciate the feedback here.

PR #1351, opened 2024-05-17
Prod auth bug fix

schroerbrian:
Mostly to fix an issue where the user session init code was being run twice and breaking the log-in flow. it was only happening on prod for some reason. this should patch it at least. Also added a public route component to redirect authed users navigating the the log-in/sign-up pages


schroerbrian:
Going to merge this in as i'd like to fix the prod bug

PR #1360, opened 2024-06-15
NavigatorDashboard: Fix infinite rerender bug.

richardxia:
Since we haven't filled out the `PARENT_ELIGIBILITIES` variable yet, the current logic keeps trying to fetch the child eligibilities for them, setting the `eligibilities` state variable to the empty array over and over again. If we actually set values for the `PARENT_ELIGIBILITIES`, then this wouldn't be a problem, but for now, guard the actual calling of `setEligibilities` to avoid infinite rerendering.


katerina-kossler (approved):
lgtm but didn't locally test to confirm

richardxia:
Thanks for reviewing!

schroerbrian:
Good catch!

PR #1362, opened 2024-06-16
Create state token on client in attempt to fix prod auth bug

schroerbrian:
see my slack message regarding this PR. it is an attempt to fix the prod auth bug, which affects chrome but not firefox, weirdly enough. i plan to deploy this today in the hopes that it will fix the prod env.


schroerbrian:
merging in to test on prod

PR #1365, opened 2024-06-27
[AuthService] Rename Auth0-specific ID field.

richardxia:
We have two different notions of user IDs: one from Auth0, which is a string and is contained in the JWT, and one from our own API, which is an integer. This renames the `id` field within the auth state to `externalId`, to make it clearer that these are from the identity provider.

I wanted to split this off from a much larger PR I'm making, since this is kind of important on its own, and I didn't want it to get lost in the noise.

### DIFF
--- a/app/utils/AuthService.ts
+++ b/app/utils/AuthService.ts
@@ -43,7 +43,7 @@ export const initializeUserSession = (
         const authObject = {
           user: {
             email: idTokenPayload.email,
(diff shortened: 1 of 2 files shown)


richardxia:
Also, I've done some pretty extensive testing with this, since I did the entire development of the hooking up of the Bookmarks API with this change, so I'm pretty confident that this is all that needs to change.

PR #1369, opened 2024-06-27
Our415 87 header nav base

rosschapman:


richardxia:
Hi @rosschapman, I was wondering if you mistakenly submitted this PR against ShelterTech's repo instead of Exygy's fork? It looks like there are many commits in this branch that are PR merge commits that don't exist in ShelterTech's repo.

We are happy in general to accept changes that would make your integration smoother, so long as they don't have any negative impact on our main site or our whitelabeled sites, but I'm guessing that was not your intention here.

PR #1392, opened 2024-08-05
Boost service discovery (tile) results based on boosted category.

richardxia:
In order to surface some specific results on the tile results page, we make use of an optional filter on a newly added Algolia index property.

On the backend, we've created a new column, `boosted_category`, which we can optionally set on each Service. This allows us perform an optional filter only on the tile results pages, and not the general search page, where services that match the optional filter are boosted in ranking.

For now, there can only be one `boosted_category` per service, but this is sufficient for the current use case, where only two services need to be boosted, and they each only apply to one category.

### DIFF
--- a/app/pages/ServiceDiscoveryResults/ServiceDiscoveryResults.tsx
+++ b/app/pages/ServiceDiscoveryResults/ServiceDiscoveryResults.tsx
@@ -168,7 +168,6 @@ const InnerServiceDiscoveryResults = ({
       // Our Algolia plan only allows for up to one optional filter, so just
       // drop all but the first.
       .slice(0, 1);
-    console.log(configureProps);
   }
 
   return (
--- a/app/pages/ServiceDiscoveryResults/ServiceDiscoveryResults.tsx
+++ b/app/pages/ServiceDiscoveryResults/ServiceDiscoveryResults.tsx
@@ -15,6 +15,7 @@ import Sidebar from "components/search/Sidebar/Sidebar";
 import { Header } from "components/search/Header/Header";
 import { Category } from "models/Meta";
 
+import whitelabel from "utils/whitelabel";
 import {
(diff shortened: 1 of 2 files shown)


app/pages/ServiceDiscoveryResults/ServiceDiscoveryResults.tsx
```diff
@@ -148,6 +150,27 @@ const InnerServiceDiscoveryResults = ({
...
+  }
+  if (whitelabel.useBoostedCategories) {
+    const selectedSubcategories = searchState.refinementList?.categories ?? [];
+    configureProps.optionalFilters = subcategoryNames
+      .filter((x) => selectedSubcategories.includes(x))
+      .map((x) => `boosted_category:${x}`)
+      // Our Algolia plan only allows for up to one optional filter, so just
+      // drop all but the first.
+      .slice(0, 1);
+    console.log(configureProps);
```
schroerbrian, line 171: Did you intend to leave in this console.log?

schroerbrian (approved):
only one comment about a console.log. 
nice work! it's great to finally have this implemented

app/pages/ServiceDiscoveryResults/ServiceDiscoveryResults.tsx
richardxia: Whoops, nice catch! Just removed it in https://github.com/ShelterTechSF/askdarcel-web/pull/1392/commits/46d2cebec68a39edec210e52da2c61f0e8199e06.

richardxia:
It looks like I'm failing CI because GitHub Actions removed support for the old version of docker-compose, and the new one is just a subcommand of the `docker` command: https://github.com/orgs/community/discussions/116610

Let me fix that in a separate PR, since I don't want to mix random infra changes with this PR.

PR #1393, opened 2024-08-15
Switch from `docker-compose` to `docker compose`.

richardxia:
Docker apparently moved the standalone `docker-compose` command into a subcommand of the `docker` command, and they deprecated the standalone command. GitHub Actions recently updated to a new enough Docker version that dropped the standalone `docker-compose` command, which is causing CI to fail.

This updates both the CI configuration as well as our README to only use the subcommand version.

See https://github.com/orgs/community/discussions/116610


richardxia:
Gonna merge this without review to unblock CI on my other PR.

Issue #1391, opened 2024-08-04
Refreshing access token can fail if user session has expired


richardxia:
When constructing the app context, we have an explicit check as to whether the authentication access token has expired, and if so, we attempt to refresh it using the Auth0 library in https://github.com/ShelterTechSF/askdarcel-web/blob/c8f11101dc0297fe073a42eab294116cd61368e7/app/utils/useAppContext.tsx#L114. However, refreshing the access token can fail if the user's session, as stored in a cookie, has itself expired, which would require the user to authenticate with Auth0 again.

Currently, we do not provide any user-facing indication that authentication has failed, and even worse, the application state may be inconsistent, with other parts of the code assuming that the presence of the `authState` object implies that the user is authenticated.

A better user experience would be to fully reset the application authentication state, and if the user were on a page that required authentication, redirect them to the login page.

PR #1398, opened 2024-10-22
Reset app auth state if Auth0 session has expired

schroerbrian:
refs https://github.com/ShelterTechSF/askdarcel-web/issues/1391

### DIFF
--- a/app/utils/useAppContext.tsx
+++ b/app/utils/useAppContext.tsx
@@ -6,6 +6,7 @@ import React, {
   useContext,
 } from "react";
 import auth0, { Auth0Result, WebAuth } from "auth0-js";
+import { useHistory } from "react-router-dom";
 import * as Sentry from "@sentry/browser";
 import * as SessionCacher from "utils/SessionCacher";
 import * as AuthService from "utils/AuthService";
@@ -69,6 +70,7 @@ export const AppProvider = ({
   userLocation: GeoCoordinates | null;
   setBookmarksMenuOpen: (open: boolean) => void;
 }) => {
+  const history = useHistory();
   const authObject = SessionCacher.getAuthObject();
   const [authState, setAuthState] = useState<AuthState>(authObject);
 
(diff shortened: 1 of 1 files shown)


katerina-kossler (commented):
the changes here look right from what I can tell but I see that some tests in CI are failing and I don't see any test coverage for this (is that normal for the web project?)

richardxia (approved):
LGTM! Though should we also actually clear that state when we detect this issue? I guess I'm thinking that if someone navigated away from the log-in page, then they'd still be in this inconsistent state where the app thinks they're logged in but their credentials have expired.

richardxia:
> the changes here look right from what I can tell but I see that some tests in CI are failing and I don't see any test coverage for this (is that normal for the web project?)

Regrettably, it's normal for the web project to have flaky failures with the end-to-end browser tests. Normally I just restart it until it finally succeeds.

schroerbrian:
> > the changes here look right from what I can tell but I see that some tests in CI are failing and I don't see any test coverage for this (is that normal for the web project?)
> 
> Regrettably, it's normal for the web project to have flaky failures with the end-to-end browser tests. Normally I just restart it until it finally succeeds.

Unfortunately, this failure seems to do with a deprecated artifact in our build system. I just haven’t had time to look further into this.  I’m away this weekend so i can’t work on it. Anyhow, i suppose this could be safely merged in if we want to, but i don’t mind waiting until we get to the bottom of this , either 

`Error: This request has been automatically failed because it uses a deprecated version of `actions/upload-artifact: v2`. Learn more: https://github.blog/changelog/2024-02-13-deprecation-notice-v1-and-v2-of-the-artifact-actions/`

schroerbrian:
> LGTM! Though should we also actually clear that state when we detect this issue? I guess I'm thinking that if someone navigated away from the log-in page, then they'd still be in this inconsistent state where the app thinks they're logged in but their credentials have expired.

Is there another state I’m missing? The code sets the authState to null prior to the redirect

schroerbrian:
Thanks for the review @richardxia and @katerina-kossler ! Sorry for my late replies here

richardxia:
> Is there another state I’m missing? The code sets the authState to null prior to the redirect

Wow, sorry, I don't know how I missed that. My brain managed to completely not see that line in between the `if` and the redirect. Yeah, sorry, this LGTM!

schroerbrian:
I am going to go ahead and merge this in as the test failure is unrelated. Thanks all!

PR #1399, opened 2024-11-16
Fix github CI action

schroerbrian:
Github deprecated v2 of their upload-artifact action, which we were using. This updates it to v4.

CI was failing. See: https://github.com/ShelterTechSF/askdarcel-web/actions/runs/11751183980/job/32741450392
And: https://github.blog/changelog/2024-02-13-deprecation-notice-v1-and-v2-of-the-artifact-actions/


richardxia (approved):
Thanks! LGTM

PR #1413, opened 2025-03-27
DC-Nav: Add African/Black eligibility to IPV category.

richardxia:
UCSF would like to add this to the second page of DC-Nav's Intimate Partner Violence form, which lists service eligibilities in the form of patient demographics. Our actual internal eligibility name is "African/Black", but they would like it to be displayed as "Black/African American", so we make use of the `alias` field to rename it in the UI.

Screenshot 2025-03-26 at 7 29 51 PM


schroerbrian (approved):
Lgtm!

richardxia:
Thanks for the quick review!

PR #1418, opened 2025-06-20
Scroll window to top on search state change

dfloo:


dfloo:
This mostly solves the issue, however it doesn't scroll all the way to the top. The version of react-instantsearch that we are using doesn't support adding a scroll offset to account for the header bar. I'm looking into a more custom approach using `window.scrollTo()` but I haven't had much luck yet.

richardxia:
Hey @dfloo, sorry for the late review. It looks like the ScrollTo component has been deprecated, and the latest version of the Algolia (react-instantsearch) library no longer has it. Fortunately, their helpdesk page on this topic has some example code for implementing this manually, so you might find that useful for a manual implementation that works around the offset issue you were hitting: https://support.algolia.com/hc/en-us/articles/22898869484433-How-can-I-fix-pagination-scrolling-up-issue-on-React-InstantSearch

richardxia (approved):
Hey @dfloo, sorry for the late review. This LGTM, though one thing I noticed when testing this out locally is that the scrolling didn't seem to be "smooth", like the function call implied. Nothing immediately jumps out to me when I look at the docs for the `scrollTo()` method, but I was curious if you actually god smooth scrolling on your end?

Also, this can either go into this PR or a followup PR, but we actually have two slightly different variations of the search page. The code you modified corresponds to using the search bar at the top of the page, and it's the generic search. We also have the "tile" search pages, called the `ServiceDiscoveryResults` in the code, which you get to if you click on one of the tiles on the home page and fill out the forms. There are slight subtle differences in the search pages, and so the code is separate, so if you could update the logic on that page as well to scroll to the top, that'd be really useful.

PR #1420, opened 2025-07-09
Add textRendering attribute to map marker text element

dfloo:
Setting the textRendering attribute of the map marker text element to "geometricPrecision" fixes the issue with the text overflowing in chrome. The appearance is still not identical between the two browsers but I think its close enough.
chrome:
map-markers-chrome
firefox:
map-markers-firefox


richardxia (approved):
Awesome, thank you for figuring that out! I agree that the difference between the two browsers is negligible, so this LGTM!

PR #1421, opened 2025-07-23
Switch over to v2 endpoint for categories

dfloo:
use v2 endpoint, remove extra nested "category" property


schroerbrian (approved):
LGTM!

PR #1433, opened 2025-10-30
feat: update version.json

lgarofalo:


.github/workflows/publish-release.yml
```diff
@@ -32,7 +32,7 @@ jobs:
       - run: npm ci
       - run: cp config.example.yml config.yml
       - run: CONFIG_YAML=config.docker.yml npm run build
-      - run: touch version.json
+      - run: echo ${{ steps.meta.outputs.tags }} > version.json
```
richardxia: Not super important, but this is probably not technically legal JSON, since it's a string that isn't quoted. I have no idea if this actually makes a difference.

PR #1463, opened 2026-05-07
Revert "Update phone change request paths to v2"

### LINKED ISSUE #260: Phone change request endpoints incorrectly require authentication
The [phone and change request endpoints are currently nested](https://github.com/ShelterTechSF/sheltertech-go/blob/1997b71b36f8e988ce594c0c225b6b85a380d438/cmd/sheltertech-go/main.go#L136-L139) under the router group which requires authentication. This is incorrect, as the Edit page has historically not required authentication, and the frontend app is not prepared to pass in the authentication token. Although we may want to enforce authentication in the future, for now we want to match the status quo.

The scope of this ticket is to ensure that the change request and other resource editing routes do not require authentication.

richardxia:
Reverts ShelterTechSF/askdarcel-web#1447

https://github.com/ShelterTechSF/sheltertech-go/issues/260 describes a bug we have where the v2 phone change request endpoints currently require auth, which is breaking the site because these routes didn't previously require authentication, and we don't actually want them to enforce authentication. To fix prod quickly, we're just reverting this PR so that the edit page switches back to the v1 endpoint for editing phones, routing it back to the Rails app.

Once we fix https://github.com/ShelterTechSF/sheltertech-go/issues/260, we can restore these changes again.

### DIFF
--- a/app/pages/OrganizationEditPage.tsx
+++ b/app/pages/OrganizationEditPage.tsx
@@ -174,19 +174,11 @@ function updateCollectionObject(
   path: string,
   promises: Promise<unknown>[]
 ) {
-  if (path === "phones") {
-    promises.push(
-      dataService.post(`/api/v2/${path}/${id}/change_requests`, {
-        change_request: { action: ACTION_EDIT, field_changes: object },
-      })
-    );
-  } else {
-    promises.push(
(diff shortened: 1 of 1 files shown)


Anchellon (approved):
lgtm

PR #1466, opened 2026-05-29
#1361 Remove resource categories from OrganizationListingPage

### LINKED ISSUE #1361: Remove resource categories from OrganizationListingPage
Related to https://github.com/ShelterTechSF/askdarcel-api/issues/745, but this can be done first.

The Resource (Organization) page currently lists the categories associated with the resource, but it's confusing because 1) there is no heading or label that indicates what this information even is, and 2) the Content team has no way of setting the resource categories, as there is no field for this on the Edit page. We should just remove this from being displayed at all.

ivanleeswe:
Removed resource categories from the Organization Listing page to reduce confusion and display only relevant organization information.


richardxia (approved):
Sorry for the delayed review, but this LGTM! Thanks for working on it!

